PR #31, opened 2015-01-08
Handle no-arg steps better

jglick:


tfennelly:
@jglick Looks fine to me as far as I get what the change is about.

PR #33, opened 2015-01-12
Make a test involving Jenkins restart pass

jglick:
Cf. [JENKINS-26137](https://issues.jenkins-ci.org/browse/JENKINS-26137).

PR #35, opened 2015-01-12
Do not call EnvVars.expand from ArtifactArchiverStepExecution

jglick:
Original code was @kohsuke’s; probably just an unexamined copy-and-paste from `hudson.tasks.ArtifactArchiver`, which _does_ need to make this call.


jglick:
@reviewbybees

The test failure seems unrelated and might in fact be solved by #33.

stephenc:
+1 from TBMOTGH

PR #36, opened 2015-01-13
[WIP] Added a simple copyartifact test

tfennelly:
See [copyartifact-plugin PR #55](https://github.com/jenkinsci/copyartifact-plugin/pull/55)


aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/copyartifact/CopyArtifactBasicTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import java.util.List;
+
+/**
+ * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
+ */
+public class CopyArtifactBasicTest {
```
jglick: Can just be dumped in `CoreStepTest` with the rest of them, I think.

tfennelly: Sure, I'll do that. I noticed `CoreStepTest` after I had created `CopyArtifactBasicTest`.

`CoreStepTest` will get very cluttered if all of these tests are dumped in there.

tfennelly:
Just to save looking... that CI test failure is because these test changes are depending on a SNAPSHOT of the coptartifact-plugin.

pom.xml
```diff
@@ -89,6 +89,11 @@
                 <artifactId>subversion</artifactId>
                 <version>2.5</version>
             </dependency>
+            <dependency>
+                <groupId>org.jenkins-ci.plugins</groupId>
+                <artifactId>copyartifact</artifactId>
+                <version>1.33-SNAPSHOT</version>
```
jglick: The `version` can be placed directly in `aggregator/pom.xml`. The only reason for having `dependencyManagement` is when multiple modules need to use the same version of a dependency, which is not the case here.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/copyartifact/CopyArtifactBasicTest.java
```diff
@@ -0,0 +1,76 @@
...
+                true));
+        WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
+
+        // check the archive is in project_2
+        assertArtifactInArchive(b);
+    }
+
+    public void createArtifactInProject_1() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "project_1");
+        p.setDefinition(new CpsFlowDefinition("node {sh 'echo hello > hello.txt'; step([$class: 'ArtifactArchiver', artifacts: 'hello.txt', fingerprint: true])}", true));
```
jglick: Do not need to enable fingerprinting for purposes of this test.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/copyartifact/CopyArtifactBasicTest.java
```diff
@@ -0,0 +1,76 @@
...
+                        "    step([$class: 'ArtifactArchiver', artifacts: 'hello.txt', fingerprint: true])\n" +
+                        "}",
+                true));
+        WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
+
+        // check the archive is in project_2
+        assertArtifactInArchive(b);
+    }
+
+    public void createArtifactInProject_1() throws Exception {
```
jglick: Could be inlined.

jglick:
> Could not find artifact org.jenkins-ci.plugins:copyartifact:jar:1.33-SNAPSHOT in repo.jenkins-ci.org (http://repo.jenkins-ci.org/public/)

You can `deploy` a `SNAPSHOT` version from https://github.com/jenkinsci/copyartifact-plugin/pull/55 to make this CI build pass (I think).

Alternately, delete this PR, and just add `workflow-aggregator` as a `test`-scoped dependency to `copyartifact-plugin`. See https://github.com/jenkinsci/credentials-binding-plugin/pull/3 for an example of this. (You need a 1.580.1 baseline in that case.)

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/copyartifact/CopyArtifactBasicTest.java
jglick: I think it makes most sense to just move this into jenkinsci/copyartifact-plugin#55 to begin with. Adding a test-only dependency should be unobjectionable.

jglick:
I think this makes more sense inside `copyartifact-plugin`. `CoreStepTest` tested build steps present in Jenkins core or bundled plugins, as a proof of concept, but clearly that does not scale to lots of integrations. (And I might just move the tests for the bundled plugins into those plugins, though one of them also tests `CatchErrorStep` which is hard to test with the steps in core.)

tfennelly:
Yes I think it made sense to have these in workflow aggregator before workflow was released, but seems like these plugins can now depend on workflow and maintain these tests themselves, right? Less messy.

PR #37, opened 2015-01-14
[JENKINS-24673] CoreWrapperStep

jglick:
Matches https://github.com/jenkinsci/jenkins/pull/1530. Generalization of https://github.com/jenkinsci/credentials-binding-plugin/pull/3. https://github.com/jenkinsci/config-file-provider-plugin/pull/9 and https://github.com/jenkinsci/xvnc-plugin/pull/5 are downstream.

- [X] fix handling of `PATH+X` environments (requires #41)
- [x] expand `${OTHERENV}` (corresponding to https://github.com/jenkinsci/jenkins/pull/1553)
- [x] wait for core release
- [x] fix test failures
- [x] update documentation
- [ ] add `CHANGES.md` entry
- [ ] add on https://github.com/jglick/workflow-plugin/issues/1

@reviewbybees (although the non-LTS baseline means I do not plan to merge it to `master` immediately)


jglick:
`WorkflowTest.env` test failure is not reproducible locally; need to investigate (probably as part of #41).

`CoreWrapperStepTest.useWrapper` failure was not reproducible locally, and passed the next time. Suggests that the test has some robustness flaw, TBD.

`PushdStepTest.restarting` failure looks like an unrelated race condition affecting restarts, perhaps.

jglick:
`PushdStepTest.restarting` still failing frequently in this branch. I cannot see any relationship between the failures and changes I am making, nor can I reproduce locally. Still investigating.

basic-steps/CORE-STEPS.md
```diff
@@ -72,8 +72,28 @@ node {
...
+
+The `wrap` step may be used to run a build wrapper defined originally for freestyle projects.
+In a workflow, any block of code (inside `node`) may be wrapped in this way, not necessarily the whole build.
+
+For example, the Xvnc plugin allows a headless build server to run GUI tests by allocating an in-memory-only X11 display.
+To use this plugin from a workflow, assuming a version with the appropriate update:
+
+```groovy
+node('linux') {
+  wrap([$class: 'Xvnc', useXauthority: true]) {
```
kohsuke: `$class` must die, granted not in this PR, but soon enough.

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStep.java
```diff
@@ -0,0 +1,156 @@
...
+            return false;
+        }
+
+        @Override public void stop(Throwable cause) throws Exception {
+            // should be no need to do anything special (but verify in JENKINS-26148)
+        }
+
+    }
+
+    private static final class Callback extends BodyExecutionCallback {
```
kohsuke: This pattern of a clean-up followed by delegation to `context` happens often enough that there it should be defined as another class. See `WorkspaceStepExecution.Callback` `TimerStepExecution.Callback`, and `ExecutorStepExecution.Callback`. `TmpDirStepExecution.Callback` is probably also that category.

kohsuke:
:+1: with just a few minor comments.

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStep.java
jglick: `ExecutorStepExecution.Callback` does not fit the pattern. `TimeoutStepExecution.Callback` might, but maybe not, since it is not `static` (uses `TimeoutStepExecution.killer`). `WorkspaceStepExecution.Execution` and `TmpDirStepExecution.Execution` could share a common convenience superclass with `CoreWrapperStep.Callback`. There is not much boilerplate here but could make sense as a separate refactoring PR.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
```diff
@@ -0,0 +1,156 @@
...
+    @ClassRule public static BuildWatcher buildWatcher = new BuildWatcher();
+    @Rule public RestartableJenkinsRule story = new RestartableJenkinsRule();
+
+    @Test public void useWrapper() throws Exception {
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                Map<String,String> slaveEnv = new HashMap<String,String>();
+                slaveEnv.put("PATH", "/usr/bin:/bin");
+                slaveEnv.put("HOME", "/home/jenkins");
+                JenkinsRuleExt.createSpecialEnvSlave(story.j, "slave", "", slaveEnv);
```
jtnord: assume isUnix?

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
```diff
@@ -0,0 +1,156 @@
...
+            }
+        });
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                SemaphoreStep.success("restarting/1", null);
+                WorkflowJob p = story.j.jenkins.getItemByFullName("p", WorkflowJob.class);
+                WorkflowRun b = p.getLastBuild();
+                story.j.assertBuildStatusSuccess(JenkinsRuleExt.waitForCompletion(b));
+                String expected = "/home/jenkins/extra/bin:/usr/bin:/bin";
+                story.j.assertLogContains("groovy PATH=" + expected, b);
```
jtnord: don't you want to check it contains exactly that path (ie /home/jenkins/extra/bin:/usr/bin:/bin:/opt/bin should fail right?)

basic-steps/CORE-STEPS.md
jtnord: does this still work with checkpointing?

jtnord:
one nit, one question and looks like some Unix dependencies creeping back in :(

assuming the answer to the checkpointing is yes then nothing that would prevent a :+1:

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
jglick: Yes, making this pass or be skipped on Windows will be necessary…just did not want to fire up a VM to check this now.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/CoreWrapperStepTest.java
jglick: Yeah, probably should do that. IIRC the easiest way is to append `:` to the message, which Bourne shell will never treat as part of a variable name.

basic-steps/CORE-STEPS.md
jglick: Good question. It will not. Making it do so would require `CoreWrapperStep` to define a `PickleFactory` for a `SimpleBuildWrapper` (which would also need to be `Serializable`), and then there would need to be some API allowing a `Pickle` to differentiate between being deserialized as part of a Jenkins restart vs. being deserialized in another `Run` as part of a checkpoint. Even that is probably not enough, because simply calling `setUp` again will not work in all cases; say the original Xvnc server set `DISPLAY=:11` and the new one sets `DISPLAY=:12` yet some code in the block is somehow caching the environment.

And adding subtle logic like this is out of the question without good test coverage, which is hard since the checkpoints are not available in OSS, so the tests would need to be in _that_ plugin, and branched to match this one.

Anyway our [documented recommendation](http://jenkins-enterprise.cloudbees.com/docs/user-guide-docs/workflow-sect-checkpoint.html#workflow-sect-checkpoint-files) for checkpoints is that they not be used inside a `node` block to begin with. There are too many things that can go wrong; you would need to get the exact same slave back, with an untouched workspace. And if checkpoints live at “top level” (i.e., in a scope not holding locks on any external resources), then they cannot be inside `wrap` either.

jglick:
I should be clear that although I consider this PR to be mergeable as it is, I am still hoping to incorporate [JENKINS-27392](https://issues.jenkins-ci.org/browse/JENKINS-27392) so that a `SimpleBuildWrapper` can also decorate log output. That would mean pushing the core dependency up to at least 1.607, which is 11 weeks after the last LTS—risky. So those changes might need to be in a cascading PR atop this one (also merging in parallel `workflow-plugin` changes to handle console decoration generally).

basic-steps/CORE-STEPS.md
jtnord: Thanks for the clarification.
*BTW - in the linked documents the example checkpoint is inside a node :-o 
 Most people will have stopped reading by the get down to the important bit of info that is burried further down.*

jtnord:
2 NITs outstanding - nothing preventing a merge :+1: as it's "better than before™"

jglick:
Merged via #120 into `lts-609`.

PR #38, opened 2015-01-15
Make a class compilable against core 1.592+.

jglick:
Needed for #37. @reviewbybees (especially @stephenc who added this new interface method)


jglick:
The PR builder seems to be confused, by building the changes here together with those in #37. Bug in `github-pull-request-build` I presume (CC @kohsuke).

recampbell:
+1

PR #39, opened 2015-01-16
[FIXED JENKINS-26104] "mail" step

tfennelly:
See [JENKINS-26104](https://issues.jenkins-ci.org/browse/JENKINS-26104).

Drags in a dep on the mailer-plugin to get MimeMessageBuilder (currrently 1.13-SNAPSHOT but will obviously be 1.13 once we release the mailer-plugin changes for [JENKINS-23074](https://issues.jenkins-ci.org/browse/JENKINS-23074)).


jglick:
Depends on https://github.com/jenkinsci/mailer-plugin/pull/16.

jglick:
Probably the `basic-steps` plugin would be the better place for this.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/mail/MailStepTest.java
```diff
@@ -0,0 +1,96 @@
...
+        job.setDefinition(new CpsFlowDefinition("node {" +
+                "   mail(to: 'tom.abcd@jenkins.org');" +
+                "}", true));
+
+        QueueTaskFuture<WorkflowRun> workflowRunQueueTaskFuture = job.scheduleBuild2(0);
+        workflowRunQueueTaskFuture.waitForStart();
+        WorkflowRun run = jenkinsRule.assertBuildStatus(Result.FAILURE, workflowRunQueueTaskFuture.get());
+
+        FlowExecution execution = run.getExecution();
+        Assert.assertEquals("Email not sent. All mandatory properties must be supplied ('to', 'subject', 'body').", execution.getCauseOfFailure().getMessage());
```
jglick: Easier to just say:

    jenkinsRule.assertBuildLogContains("…", run);

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/mail/MailStepTest.java
```diff
@@ -0,0 +1,96 @@
...
+    @Test
+    public void test_missing_param_has_continue() throws Exception {
+        WorkflowJob job = jenkinsRule.jenkins.createProject(WorkflowJob.class, "workflow");
+        // leave out he subject and body
+        job.setDefinition(new CpsFlowDefinition("node {" +
+                "   mail(to: 'tom.abcd@jenkins.org', continueOnError: true);" +
+                "}", true));
+
+        WorkflowRun run = jenkinsRule.assertBuildStatusSuccess(job.scheduleBuild2(0));
+        Assert.assertTrue(run.getLog(1000).toString().contains("Running: Mail, ERROR: Email not sent. All mandatory properties must be supplied ('to', 'subject', 'body')."));
```
jglick: Again use `assertBuildLogContains`.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/mail/MailStepTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Test
+    public void test_missing_param_has_continue() throws Exception {
+        WorkflowJob job = jenkinsRule.jenkins.createProject(WorkflowJob.class, "workflow");
+        // leave out he subject and body
+        job.setDefinition(new CpsFlowDefinition("node {" +
+                "   mail(to: 'tom.abcd@jenkins.org', continueOnError: true);" +
+                "}", true));
+
+        WorkflowRun run = jenkinsRule.assertBuildStatusSuccess(job.scheduleBuild2(0));
```
jglick: This makes no sense to me. What is `continueOnError` for? If a field is mandatory, then the flow must set it, and there is no reason to allow recovery from the mistake if it does not.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStepExecution.java
```diff
@@ -0,0 +1,104 @@
...
+import javax.mail.Message;
+import javax.mail.MessagingException;
+import javax.mail.Transport;
+import javax.mail.internet.MimeMessage;
+import java.io.UnsupportedEncodingException;
+
+/**
+ * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
+ */
+public class MailStepExecution extends AbstractStepExecutionImpl {
```
jglick: Prefer to put this as a nested class of `MailStep` for readability.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStep.java
```diff
@@ -0,0 +1,93 @@
...
+    public String replyTo;
+
+    @DataBoundSetter
+    public String mimeType;
+
+    @DataBoundSetter
+    public String defaultSuffix;
+
+    @DataBoundSetter
+    public boolean continueOnError;
```
jglick: Get rid of this flag. Either the mistake was in the flow configuration, in which case no recovery is appropriate; or it was a runtime error, in which case the flow can use a `try`-`catch` block if it wants to handle the error and continue.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/mail/MailStepTest.java
```diff
@@ -0,0 +1,96 @@
...
+        Assert.assertTrue(run.getLog(1000).toString().contains("Running: Mail, ERROR: Email not sent. All mandatory properties must be supplied ('to', 'subject', 'body')."));
+    }
+
+    @Test
+    public void test_send() throws Exception {
+        Mailbox.clearAll();
+
+        WorkflowJob job = jenkinsRule.jenkins.createProject(WorkflowJob.class, "workflow");
+        job.setDefinition(new CpsFlowDefinition("node {" +
```
jglick: Should be no need to wrap it in `node`; it does not require a `FilePath` or other node-related context parameters.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/mail/MailStepTest.java
```diff
@@ -0,0 +1,96 @@
...
+    public void test_missing_param_no_continue() throws Exception {
+        WorkflowJob job = jenkinsRule.jenkins.createProject(WorkflowJob.class, "workflow");
+        // leave out he subject and body
+        job.setDefinition(new CpsFlowDefinition("node {" +
+                "   mail(to: 'tom.abcd@jenkins.org');" +
+                "}", true));
+
+        QueueTaskFuture<WorkflowRun> workflowRunQueueTaskFuture = job.scheduleBuild2(0);
+        workflowRunQueueTaskFuture.waitForStart();
+        WorkflowRun run = jenkinsRule.assertBuildStatus(Result.FAILURE, workflowRunQueueTaskFuture.get());
```
jglick: Can be simplified:

    WorkflowRun run = jenkinsRule.assertBuildStatus(Result.FAILURE, job.scheduleBuild2(0));

support/pom.xml
```diff
@@ -54,6 +54,15 @@
             <artifactId>jboss-marshalling-river</artifactId>
             <version>1.4.9.Final</version>
         </dependency>
+        <!--
+        Required for the MailStep.
+        TODO: Is this OK here, or want to create a dedicated module and keep the deps clean here?
```
jglick: Recommend `basic-steps`.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStep.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
+ */
+public class MailStep extends AbstractStepImpl {
+
+    @DataBoundSetter
+    public String charset;
+
+    @DataBoundSetter
+    public @CheckForNull String subject;
```
jglick: This is mandatory, so it should not be `@CheckForNull`, or even a `@DataBoundSetter`; it should be a parameter of the `@DataBoundConstructor`. Same for `body`. (`to` could I guess be omitted, if you want to only use `bcc` for example.)

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStep.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @DataBoundSetter
+    public @CheckForNull String body;
+
+    @DataBoundSetter
+    public String from;
+
+    @DataBoundSetter
+    public @CheckForNull String to;
```
jglick: `List<String>` might be more intuitive. I know it all gets flattened to one comma-separated string in the actual mail message, but logically from the program you have a list of recipients.

The downside of that signature is that I am not sure how to make data binding in `config.jelly` bind a `List<String>` without some hacks. `f:repeatable` might work, but I think this only takes `Describable` elements.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStepExecution.java
jglick: Use `AbstractSynchronousStepExecution` if you do not intend to use a callback.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStepExecution.java
```diff
@@ -0,0 +1,104 @@
...
+
+    public static final long serialVersionUID = 1L;
```
jglick: Make it `private`.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStepExecution.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Inject
+    private transient MailStep step;
+
+    @StepContextParameter
+    private transient TaskListener listener;
+
+    @Override
+    public boolean start() throws Exception {
+        if (StringUtils.isBlank(step.to) || StringUtils.isBlank(step.subject) || StringUtils.isBlank(step.body)) {
```
jglick: Legitimate to have blank `to` if some other kind of recipient is specified.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStepExecution.java
```diff
@@ -0,0 +1,104 @@
...
+            String errorMessage = "Email not sent. All mandatory properties must be supplied ('to', 'subject', 'body').";
+            if (step.continueOnError) {
+                listener.error(errorMessage);
+                getContext().onSuccess(null);
+                return true;
+            } else {
+                throw new IllegalArgumentException(errorMessage);
```
jglick: With `continueOnError` gone, can all be collapsed to one line.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStepExecution.java
```diff
@@ -0,0 +1,104 @@
...
+            Transport.send(mimeMessage);
+            getContext().onSuccess(null);
+        } catch (Exception e) {
+            if (step.continueOnError) {
+                throw e;
+            } else {
+                listener.error("Failed to send email.", e);
+                getContext().onSuccess(null);
+            }
+        }
```
jglick: With `AbstractSynchronousStepExecution`, the whole `try` block can be deleted; any error is assumed to be thrown up. Just

    Transport.send(buildMimeMessage());
    return null; // for Void return value

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStepExecution.java
```diff
@@ -0,0 +1,104 @@
...
+
+        // Sync exec.
+        // TODO: See about doing it async ?
+        return true;
+    }
+
+    @Override
+    public void stop(@Nullable Throwable cause) throws Exception {
+    }
```
jglick: This goes away with `AbstractSynchronousStepExecution`.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStepExecution.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private MimeMessage buildMimeMessage() throws UnsupportedEncodingException, MessagingException {
+        return new MimeMessageBuilder()
+                .setListener(listener)
+                .setSubject(step.subject)
+                .setBody(step.body)
+                .setCharset(step.charset)
```
jglick: What does this due when the `charset` is null? Is that allowed?

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStep.java
```diff
@@ -0,0 +1,93 @@
...
+import org.kohsuke.stapler.DataBoundSetter;
+
+import javax.annotation.CheckForNull;
+
+/**
+ * Simple email sender step.
+ *
+ * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
+ */
+public class MailStep extends AbstractStepImpl {
```
jglick: You are missing a `config.jelly`, needed for the snippet generator to work.

To test the form, use `StepConfigTester`.

aggregator/pom.xml
```diff
@@ -131,7 +131,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>mailer</artifactId>
-            <version>1.12</version>
+            <version>1.13-SNAPSHOT</version>
```
jglick: Depends on https://github.com/jenkinsci/mailer-plugin/pull/16. You could try

    mvn -DskipTests clean deploy

from a checkout of that branch, then set this version to the timestamped snapshot, something like `1.13-20150116.193108-1` (you will see the output from `deploy:deploy`). This allows us to pin this PR to a particular revision of the upstream PR, and should allow the PR builder to actually run on this one.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/mail/MailStepTest.java
tfennelly: Makes no sense to me either :( Has all the signs and smells of a brain fart.

I'd say I thought it was needed in the case where there's an error sending, but then stupidly applied to to the config checks too.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStep.java
tfennelly: A thanks for spotting `@CheckForNull`. I copied it from elsewhere thinking it would result in a runtime check, but then realised it didn't and was just a build time findbugs check.

When you said it "should be a `@DataBoundConstructor`" I was thinking the args would be auto-checked for null but that does not seem to be the case. If that's the case, then I don't really see the benefit of `@DataBoundConstructor` other than if it's a convention.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStepExecution.java
tfennelly: Fair point.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStepExecution.java
tfennelly: `charset` can't be set to `null` or an empty string.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStepExecution.java
tfennelly: I agree that `MailStepExecution` would be easier to maintain inside `MailStep` (my natural inclination would be to do that), but this is not how it's been done for the other steps i.e. they have all been implemented in separate java files. It's a small thing though either way. I'll move them together into one.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStep.java
jglick: > I don't really see the "benefit" of `@DataBoundConstructor`

Every `Describable` class must have exactly one, even if empty.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/mail/MailStepExecution.java
jglick: > they have all been implemented in separate java files (at least the ones I checked)

Only the ones Kohsuke wrote; I have already inlined most of them, but apparently you looked at some of the remaining ones. (Once you release the plugin, then you cannot change it, since the serial form of running flow builds encodes the class name of the `StepExecution` implementation.)

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/MailStepTest.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+public class MailStepTest {
+
+    @Rule
+    public JenkinsRule jenkinsRule = new JenkinsRule();
+
+    @Test
+    public void test_missing_subject() throws Exception {
+        WorkflowJob job = jenkinsRule.jenkins.createProject(WorkflowJob.class, "workflow");
+        // leave out he subject
```
jglick: Typo.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/MailStepTest.java
```diff
@@ -0,0 +1,96 @@
...
+        job.setDefinition(new CpsFlowDefinition("mail(to: 'tom.abcd@jenkins.org', body: 'body');", true));
+
+        WorkflowRun run = jenkinsRule.assertBuildStatus(Result.FAILURE, job.scheduleBuild2(0).get());
+        jenkinsRule.assertLogContains("Email not sent. All mandatory properties must be supplied ('subject', 'body').", run);
+    }
+
+    @Test
+    public void test_missing_body() throws Exception {
+        WorkflowJob job = jenkinsRule.jenkins.createProject(WorkflowJob.class, "workflow");
+        // leave out he subject
```
jglick: Copy-paste error.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/MailStepTest.java
```diff
@@ -0,0 +1,96 @@
...
+        job.setDefinition(new CpsFlowDefinition("mail(to: 'tom.abcd@jenkins.org', subject: 'subject');", true));
+
+        WorkflowRun run = jenkinsRule.assertBuildStatus(Result.FAILURE, job.scheduleBuild2(0).get());
+        jenkinsRule.assertLogContains("Email not sent. All mandatory properties must be supplied ('subject', 'body').", run);
+    }
+
+    @Test
+    public void test_missing_recipient() throws Exception {
+        WorkflowJob job = jenkinsRule.jenkins.createProject(WorkflowJob.class, "workflow");
+        // leave out he subject and body
```
jglick: Typo.

basic-steps/pom.xml
```diff
@@ -41,6 +41,11 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>mailer</artifactId>
+            <version>1.13-SNAPSHOT</version>
```
jglick: Need to move this version up into `dependencyManagement` of the root POM.

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/MailStep.java
```diff
@@ -0,0 +1,145 @@
...
+ * Simple email sender step.
+ *
+ * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
+ */
+public class MailStep extends AbstractStepImpl {
+
+    @DataBoundSetter
+    public String charset;
+
+    public String subject;
```
jglick: Can be `final`. (Ditto `body`.)

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/MailStep.java
```diff
@@ -0,0 +1,145 @@
...
+
+        @Override public String getDisplayName() {
+            return "Mail";
+        }
+    }
+
+    /**
+     * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
+     */
+    public static class MailStepExecution extends AbstractSynchronousStepExecution {
```
jglick: `<Void>`

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/MailStep.java
```diff
@@ -0,0 +1,145 @@
...
+        @Override
+        protected Object run() throws Exception {
+            MimeMessage mimeMessage = buildMimeMessage();
+            Transport.send(mimeMessage);
+            return null;
+        }
+
+        private MimeMessage buildMimeMessage() throws UnsupportedEncodingException, MessagingException {
+            if (StringUtils.isBlank(step.subject) || StringUtils.isBlank(step.body)) {
+                throw new IllegalArgumentException("Email not sent. All mandatory properties must be supplied ('subject', 'body').");
```
jglick: Use `AbortException` for user errors like this where no stack trace is needed.

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/MailStep.java
```diff
@@ -0,0 +1,145 @@
...
+                    .addRecipients(step.to, Message.RecipientType.TO)
+                    .addRecipients(step.cc, Message.RecipientType.CC)
+                    .addRecipients(step.bcc, Message.RecipientType.BCC)
+                    .setReplyTo(step.replyTo)
+                    .setDefaultSuffix(step.defaultSuffix)
+                    .buildMimeMessage();
+
+            Address[] allRecipients = message.getAllRecipients();
+            if (allRecipients == null || allRecipients.length == 0) {
+                throw new IllegalArgumentException("Email not sent. No recipients of any kind specified ('to', 'cc', 'bcc').");
```
jglick: Ditto here.

basic-steps/src/main/resources/org/jenkinsci/plugins/workflow/steps/MailStep/config.jelly
```diff
@@ -0,0 +1,58 @@
...
+    <f:entry field="bcc" title="BCC">
+        <f:textbox />
+    </f:entry>
+    <f:entry field="subject" title="Subject">
+        <f:textbox />
+    </f:entry>
+    <f:entry field="body" title="Body">
+        <f:textbox />
+    </f:entry>
+    <f:entry field="replyTo" title="Reply-To">
```
jglick: Some of these could probably be inside `f:advanced` since there is not commonly a need to customize them.

Should also create `help-replyTo.html` etc. where it is not obvious what the field is for, especially when it is not obvious what happens when it is left blank.

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/MailStep.java
```diff
@@ -0,0 +1,145 @@
...
+    public String bcc;
+
+    @DataBoundSetter
+    public String replyTo;
+
+    @DataBoundSetter
+    public String mimeType;
+
+    @DataBoundSetter
+    public String defaultSuffix;
```
jglick: What is this field for? There is a default suffix configured in Jenkins global configuration which can be used for recipients missing `@` (a somewhat dubious feature). But if you do not want to use that suffix, you do not need this field. Just write out the full address to begin with.

basic-steps/src/test/java/org/jenkinsci/plugins/workflow/steps/MailStepTest.java
```diff
@@ -0,0 +1,65 @@
...
+        Assert.assertEquals("subject", step2.subject);
+        Assert.assertEquals("body", step2.body);
+        Assert.assertEquals("tom.fennelly@gmail.com", step2.from);
+        Assert.assertEquals("tom.fennelly@gmail.com", step2.to);
+        Assert.assertEquals("tom.fennelly@gmail.com", step2.cc);
+        Assert.assertEquals("tom.fennelly@gmail.com", step2.bcc);
+        Assert.assertEquals("UTF-8", step2.charset);
+        Assert.assertEquals("tom.fennelly@gmail.com", step2.replyTo);
+        Assert.assertEquals("text/html", step2.mimeType);
+        Assert.assertEquals("@blah.net", step2.defaultSuffix);
```
jglick: `r.assertEqualDataBoundBeans(step1, step2);`

tfennelly:
I think/hope this can be merged as soon as PR https://github.com/jenkinsci/mailer-plugin/pull/16 is merged upstream.

jglick:
:+1:

tfennelly:
@reviewbybees any one else that can review this?

kohsuke:
:+1:

jglick:
@tfennelly it looks like you “merged” this PR by rebasing all of its commits into `master` for some reason. For the future you can just click the merge button in GH (or use `git merge`, which automatically closes the PR).

tfennelly:
@jglick oops I didn't realise that was a sin, sorry.

It's kinda funny. Last place I worked, you'd get your b**ls handed to you in a jar if you didn't `git pull --rebase ...` before pushing changes to a project. It was more or less universally considered bad to just merge due of the ugly history that was left after. Doctors differ and patients die :)

PR #40, opened 2015-01-21
[FIXED JENKINS-26101] Load script from SCM

jglick:
[JENKINS-26101](https://issues.jenkins-ci.org/browse/JENKINS-26101)

- [X] proof of concept
- [x] tests
- [x] properly handle `TaskListener`
- [x] make snippet generator work
- [X] script SCM participate in changelog/polling
- [x] use from `demo/JENKINS_HOME/jobs/cd/config.xml`
- [x] inline help
- [x] update `TUTORIAL.md` and `CHANGES.md`

@reviewbybees


jtnord:
The relative path checker would have been usefull for catching misconfigurations (rather than security).
e.g. if you load a non relative file then when the job moves to a different slave at an abitrary moment the step command  - rather than a shell script failing when run from a command.  I think the steps failing are at the moment generally harder to diagnose - so am questioning if this should have been removed (even if it was disabled by default).

jglick:
> if you load a non relative file

This was actually a valid customer use case reported to me, which is why I disabled the validation. IIRC they wanted to use `dir` in a special location known to exist on the slave machine. Obviously if you change the slave configuration that will not work, but neither will plain old shell scripts using `pushd`, and I think the error in such a case would be self-explanators.

jtnord:
fair enough.
+1

jglick:
@reviewbybees ready to merge as far as I am concerned.

kohsuke:
Firstly, I just want to say this feature is incredible!!

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
```diff
@@ -128,32 +123,8 @@ public FormValidation doCheckScript(@QueryParameter String value, @QueryParamete
...
-                return HttpResponses.plainText(Functions.printThrowable(x));
-            }
-            try {
-                return HttpResponses.plainText(Snippetizer.object2Groovy(o));
-            } catch (UnsupportedOperationException x) {
-                Logger.getLogger(CpsFlowExecution.class.getName()).log(Level.WARNING, "failed to render " + json, x);
-                return HttpResponses.plainText(x.getMessage());
-            }
+        public Snippetizer getSnippetizer() {
+            return Snippetizer.get();
```
kohsuke: You can just inject Snippetizer as a field like `@Inject private Snippetizer snippetizer;` and avoid static `get` method.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/FlowHead.java
```diff
@@ -92,7 +92,13 @@ public CpsFlowExecution getExecution() {
...
     }
 
-    void newStartNode(BlockStartNode n) throws IOException {
+    void newStartNode(FlowStartNode n) throws IOException {
+        for (Action a : execution.flowStartNodeActions) {
+            if (a instanceof FlowNodeAction) {
+                ((FlowNodeAction) a).onLoad(n);
+            }
+            n.addAction(a);
+        }
```
kohsuke: Perhaps `execution.flowStartNodeAction.clear()` here?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -0,0 +1,139 @@
...
+                return ((CpsFlowFactoryAction2) a).create(this, owner, actions);
+            }
+        }
+        FilePath dir;
+        Queue.Executable _build = owner.getExecutable();
+        if (!(_build instanceof Run)) {
+            throw new IOException("can only check out SCM into a Run");
+        }
+        Run<?,?> build = (Run<?,?>) _build;
+        Jenkins jenkins = Jenkins.getInstance();
```
kohsuke: I guess the future TODO is to support a specific slave. Sometimes credentials/network visibility demands that the checkout happen in a specific node.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        public Collection<? extends SCMDescriptor<?>> getApplicableDescriptors() {
+            StaplerRequest req = Stapler.getCurrentRequest();
+            Job job = req != null ? req.findAncestorObject(Job.class) : null;
+            return job != null ? SCM._for(job) : /* TODO 1.599+ does this for job == null */ SCM.all();
+        }
+
+        public Snippetizer getSnippetizer() {
+            return Snippetizer.get();
```
kohsuke: Injection can be used here, too.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
jglick: Right, there is no particular reason the same procedure could not be done on a slave; you would just need to allow a label expression to be configured. I think it can be skipped unless someone requests it.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/FlowHead.java
jglick: Makes sense, just to be on the safe side.

jglick:
@reviewbybees re-review

kohsuke:
:+1:

recampbell:
+1

tfennelly:
Awesome feature Jesse !!

apemberton:
Agree. Very good idea. :)

PR #42, opened 2015-01-22
simple build fails without correct plugin repositories listed 

gclayburg:
(mvn -DskipTests clean install)
without this fix the build fails with this error:
[ERROR]     Unresolveable build extension: Plugin org.jenkins-ci.tools:maven-hpi-plugin:1.110 or one of its dependencies could not be resolved: Could not find artifact org.jenkins-ci.tools:maven-hpi-plugin:jar:1.110 in central (http://repo.maven.apache.org/maven2) -> [Help 2]
[ERROR]

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -106,12 +106,6 @@
             <scope>provided</scope>
         </dependency>
     </dependencies>
-    <repositories>
-        <repository>
-            <id>repo.jenkins-ci.org</id>
-            <url>http://repo.jenkins-ci.org/public/</url>
-        </repository>
-    </repositories>
     <build>
         <plugins>
             <!-- TODO replace with java.level=6 once we have a better plugin parent POM: -->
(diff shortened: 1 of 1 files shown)


jglick:
Interesting… of course that plugin is only in the Jenkins repo, not in Central, yet this `pluginRepository` is declared in the parent POM and I thought Maven would first resolve the project model before looking for it. Perhaps because it is an extension it is loaded during an earlier part of the model building process.

gclayburg:
Hey Jesse - great work on the plugin.  This opens up much more advanced ways to manage builds!

Anyway, this was my first time trying to build a Jenkins plugin.  I don't have any special settings.xml or other Jenkins projects built in my local repo.  That probably has something to do with it.

jglick:
Yes, that is why you are seeing it—but of course the intention is that you should not _need_ any special settings to build. (I think our CI builder is configured to predefine the Jenkins repo for all plugin jobs, which is probably a poor decision.)

PR #44, opened 2015-01-23
Jenkins workflow plugin groovy code completion support for Intellij IDEA

gclayburg:
I created the beginnings of code completion support for IntelliJ IDEA.  With the gdsl file in the workflow-basic-steps module, it can be used by simply including this maven module as a dependency:

        <dependency>
            <groupId>org.jenkins-ci.plugins.workflow</groupId>
            <artifactId>workflow-basic-steps</artifactId>
            <version>1.2-SNAPSHOT</version>
        </dependency>

Hopefully this is useful to someone else.  Maybe there is a better place to put this?
https://groups.google.com/forum/#!topic/jenkinsci-users/FCSUZXI8K7w

### DIFF
--- a/basic-steps/src/main/resources/org/jenkinsci/plugins/workflow/steps/workflow.gdsl
+++ b/basic-steps/src/main/resources/org/jenkinsci/plugins/workflow/steps/workflow.gdsl
@@ -1,0 +1,137 @@
+package org.jenkinsci.plugins.workflow.steps
+/*
+Author: Gary Clayburg
+This file allows IntelliJ IDEA to perform basic syntax checking and code completion for
+Jenkins workflow groovy scripts. https://github.com/jenkinsci/workflow-plugin
+
+These workflow steps are supported
+sh
+readFile
+node
+echo
+
+Usage:
+1. Place this file somewhere in your classpath for the your IntelliJ project that contains your flow.groovy Jenkins workflow
+2. You may see a note in IntelliJ IDEA that says "DSL descriptor file has been changed and isn't currently executed."  If you see this, click the "Activate back" button.
+OR
+1. Specify this maven dependency in your project:
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-basic-steps</artifactId>
+            <version>1.2-SNAPSHOT</version>
+        </dependency>
+
+
+Your jenkins workflow plugin groovy script should now offer quick Javadoc for supported build steps (sh,readFile,node,echo) and basic autocompletion hints
+
+More background on IntelliJ and groovy DSL support here:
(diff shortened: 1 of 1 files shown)


FlxPeters:
+1 Thats pretty cool

jglick:
As currently conceived, this would be better placed in the `workflow-aggregator` plugin. (Is there actually a specific reason to bundle it inside a plugin, or can it just be retrieved from static web content? Your first installation suggestion implies that it could; your second implies that IDEA somehow detects this file automatically when it is somewhere in the classpath.)

More broadly, as a solution to [JENKINS-26126](https://issues.jenkins-ci.org/browse/JENKINS-26126) this is unsatisfactory, because you are only including syntax for a few steps (which people could easily memorize anyway), whereas the important use case is less-common steps you are not already familiar with, perhaps added by plugins outside the standard Workflow suite; or steps with obscure options that are likely to evolve from release to release. What is needed is for the `workflow-cps` plugin to be able to generate complete DSLD and/or GDSL on demand, according to the set of currently installed plugins providing steps (or extensions like `SCM` and `SimpleBuildStep` that can be used by metasteps). This presupposes some refactoring of `DescribableHelper` so that it can report on class metadata independently of a particular instance.

gclayburg:
Hey Jesse - you are correct, IDEA just needs this static file to exist somewhere in the classpath.  I just stuck it in one of your modules so that maybe it could be maintained there alongside the workflow step implementation.  I'm not that familiar with the workflow codebase, but you probably could create something to generate this file.  In some cases, there is duplication between Javadoc in the step implementations and the existing help system that you see on the Jenkins config page.  In any event, it is also possible to create multiple gdsl files, perhaps one for each plugin?

So, the motivation for doing this was to allow IntelliJ to help flag syntax errors in the build script before a complete Jenkins build ran and reported errors.  In my case, I was learning the basic workflow suite and the options available.  I just prefer having the documentation available in the editor when possible.  I suppose once you are intimately familiar with all the steps you wouldn't need this, but it helped me.

jglick:
> maybe it could be maintained there alongside the workflow step implementation

As I wrote, that is not sustainable. Remember that not only `Step`s provided in plugins can change their direct options, but metasteps can pick up options from Jenkins core or other plugins with no Workflow dependencies at all, which are in some cases quite complex and frequently evolving. For example,

```groovy
checkout([$class: 'GitSCM', …lots of stuff here…])
```

> you probably could create something to generate this file

Well that is the hard part. :-)

> it is also possible to create multiple gdsl files, perhaps one for each plugin?

Not practical. My idea is rather that the `workflow-cps` plugin would offer a button that would generate GDSL and/or DSLD on the fly according to all currently installed plugins. You would then save the result somewhere and ask your IDE to use it.

jglick:
This is definitely a useful beginning but I do not think it is far enough along to be worth including in the official release yet. People who want to experiment with this can use the sources here as a starting point.

PR #45, opened 2015-01-23
Skip unit tests when required tooling (hg, svn) is not installed.

jtnord:
If the SCM tools are not installed (subversion/mercurial) then don't fail the unit tests.

@reviewbybees


recampbell:
+1

tfennelly:
:+1: 

I assume there's a warning of some sort allowing you to follow up and fix?

jglick:
> I assume there's a warning of some sort allowing you to follow up and fix?

That would be a _Skipped_ entry in the test report.

:+1: and merging, thanks.

PR #46, opened 2015-01-26
Noting copyartifact and mail steps in COMPATIBILITY.md

tfennelly:
@reviewbybees please

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -23,7 +23,7 @@ See [this guide](basic-steps/CORE-STEPS.md#adding-support-from-plugins) for maki
 - [X] `JUnitResultArchiver` (`junit`)
 - [X] `JavadocArchiver` (`javadoc`)
 - [X] `Mailer` (`mailer`)
-- [ ] `CopyArtifact` (`copyartifact`): [JENKINS-24887](https://issues.jenkins-ci.org/browse/JENKINS-24887)
+- [X] `CopyArtifact` (`copyartifact`): supported as of 1.34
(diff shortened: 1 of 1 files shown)


rsandell:
+1

COMPATIBILITY.md
```diff
@@ -69,4 +69,4 @@ For cases when a first-class Workflow step (rather than an adaptation of functio
 
 - [X] `parallel-test-executor`: supported with `splitTests` step since 1.6
 - [ ] `gerrit-trigger`: [JENKINS-26102](https://issues.jenkins-ci.org/browse/JENKINS-26102), [JENKINS-26103](https://issues.jenkins-ci.org/browse/JENKINS-26103)
-- [ ] `mailer`: [JENKINS-26104](https://issues.jenkins-ci.org/browse/JENKINS-26104)
+- [X] `mailer` (`workflow`): supported with `mail` step since 1.2
```
jtnord: this is confusing as mailer plugin 1.2 is very old.  THis should say supported since 1.2 of *workflow* (since all other sinces refer to the plugin version not workflow version.

jtnord:
see file comment 
:-1:

COMPATIBILITY.md
tfennelly: The new `mail` step is in the `workflow` plugin (not the `mailer` plugin).

My interpretation of the format in this doc is that the first token is the name of the step, then the second is the name of the plugin (i.e. `workflow` here). So I should probably change the first token from `mailer` to `mail`.

I do agree though that the current format is not too clear and can be misinterpreted.

tfennelly: Nah, on second review, the first token is probably the plugin name.

jglick:
I already did this myself.

PR #47, opened 2015-01-30
[JENKINS-26121] Approver Logged

varmenise:
This solves https://issues.jenkins-ci.org/browse/JENKINS-26121.

The input step now records and shows who approved the request.

However, if you are NOT logged in and try to approve, you are prompted with the error message which says "you need to be X to submit".
if you log in, then you are asked to "try POST" and if you click POST an empty request(request has not parameters) is SENT causing the doSubmit to abort  the input step (even if you have approved instead).

   @RequirePOST
    public HttpResponse doSubmit(StaplerRequest request) throws IOException, ServletException, InterruptedException {
        if (request.getParameter("proceed")!=null) {
            User user= User.current();
            doProceed(request, user);
        } else {
            doAbort();
        }

        // go back to the Run console page
        return HttpResponses.redirectTo("../../console");
    }

This behaviour is detected in the master as well and should be fixed.

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepTest.java
@@ -1,3 +1,27 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2013-2014, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
 package org.jenkinsci.plugins.workflow.steps.input;
 
 import com.gargoylesoftware.htmlunit.html.HtmlPage;
@@ -5,6 +29,7 @@ import hudson.model.BooleanParameterDefinition;
 import hudson.model.Job;
 import hudson.model.User;
 import hudson.model.queue.QueueTaskFuture;
+
 
 import java.util.Arrays;
 import java.util.List;
@@ -17,6 +42,7 @@ import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.jenkinsci.plugins.workflow.support.steps.input.ApproverAction;
 import org.jenkinsci.plugins.workflow.support.steps.input.InputAction;
 import org.jenkinsci.plugins.workflow.support.steps.input.InputStepExecution;
 import org.junit.Assert;
@@ -24,6 +50,8 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
+
+import java.util.Arrays;
 
 /**
  * @author Kohsuke Kawaguchi
@@ -37,11 +65,15 @@ public class InputStepTest extends Assert {
      */
     @Test
     public void parameter() throws Exception {
+
+
+        //set up dummy security real
+        j.jenkins.setSecurityRealm(j.createDummySecurityRealm());
         // job setup
         WorkflowJob foo = j.jenkins.createProject(WorkflowJob.class, "foo");
         foo.setDefinition(new CpsFlowDefinition(StringUtils.join(Arrays.asList(
                 "echo('before');",
-                "def x = input message:'Do you want chocolate?', id:'Icecream', ok: 'Purchase icecream', parameters: [[$class: 'BooleanParameterDefinition', name: 'chocolate', defaultValue: false, description: 'Favorite icecream flavor']];",
+                "def x = input message:'Do you want chocolate?', id:'Icecream', ok: 'Purchase icecream', parameters: [[$class: 'BooleanParameterDefinition', name: 'chocolate', defaultValue: false, description: 'Favorite icecream flavor']], submitter:'alice';",
                 "echo(\"after: ${x}\");"),"\n"),true));
 
 
@@ -61,11 +93,14 @@ public class InputStepTest extends Assert {
         InputStepExecution is = a.getExecution("Icecream");
         assertEquals("Do you want chocolate?", is.getInput().getMessage());
         assertEquals(1, is.getInput().getParameters().size());
+        assertEquals("alice", is.getInput().getSubmitter());
 
         j.assertEqualDataBoundBeans(is.getInput().getParameters().get(0), new BooleanParameterDefinition("chocolate", false, "Favorite icecream flavor"));
 
         // submit the input, and run workflow to the completion
-        HtmlPage p = j.createWebClient().getPage(b, a.getUrlName());
+        JenkinsRule.WebClient wc = j.createWebClient();
+        wc.login("alice");
+        HtmlPage p = wc.getPage(b, a.getUrlName());
         j.submit(p.getFormByName(is.getId()),"proceed");
         assertEquals(0, a.getExecutions().size());
         q.get();
@@ -73,6 +108,11 @@ public class InputStepTest extends Assert {
         // make sure 'x' gets assigned to false
         System.out.println(b.getLog());
         assertTrue(b.getLog().contains("after: false"));
+
+        //make sure the approver name corresponds to the submitter
+        ApproverAction action = b.getAction(ApproverAction.class);
+        assertNotNull(action);
+        assertEquals(is.getInput().getSubmitter(), action.getUserId());
     }
 
     @Test
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepTest.java
@@ -24,6 +24,8 @@
 
 package org.jenkinsci.plugins.workflow.steps.input;
 
+import com.gargoylesoftware.htmlunit.ElementNotFoundException;
+import com.gargoylesoftware.htmlunit.html.HtmlAnchor;
 import com.gargoylesoftware.htmlunit.html.HtmlPage;
 import hudson.model.BooleanParameterDefinition;
 import hudson.model.Job;
@@ -101,9 +103,21 @@ public class InputStepTest extends Assert {
         JenkinsRule.WebClient wc = j.createWebClient();
         wc.login("alice");
         HtmlPage p = wc.getPage(b, a.getUrlName());
-        j.submit(p.getFormByName(is.getId()),"proceed");
+        j.submit(p.getFormByName(is.getId()), "proceed");
         assertEquals(0, a.getExecutions().size());
         q.get();
+
+        // make sure the valid hyperlink of the approver is created in the build index page
+        HtmlAnchor pu =null;
+
+        try {
+            pu = p.getAnchorByText("alice");
+        }
+        catch(ElementNotFoundException ex){
+            System.out.println("valid hyperlink of the approved does not appears on the build index page");
+        }
+
+        assertNotNull(pu);
 
         // make sure 'x' gets assigned to false
         System.out.println(b.getLog());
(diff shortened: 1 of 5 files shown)


aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepTest.java
```diff
@@ -52,11 +54,14 @@ public void parameter() throws Exception {
         InputStepExecution is = a.getExecution("Icecream");
         assertEquals("Do you want chocolate?", is.getInput().getMessage());
         assertEquals(1, is.getInput().getParameters().size());
+        assertEquals("alice", is.getInput().getSubmitter());
```
jglick: How can this be `alice` when she has not yet logged in?

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepTest.java
```diff
@@ -52,11 +54,14 @@ public void parameter() throws Exception {
...
 
         j.assertEqualDataBoundBeans(is.getInput().getParameters().get(0), new BooleanParameterDefinition("chocolate", false, "Favorite icecream flavor"));
 
         // submit the input, and run workflow to the completion
-        HtmlPage p = j.createWebClient().getPage(b, a.getUrlName());
+        JenkinsRule.WebClient wc = j.createWebClient();
+        wc.login("alice");
```
jglick: This will not work unless you have first called

    j.jenkins.setSecurityRealm(j.createDummySecurityRealm());

which you would probably want to do at the top of the test method.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepTest.java
varmenise: I think I am just comparing "Alice" with the submitter parameter defined in the input set definition above:
 assertEquals("alice", is.getInput().getSubmitter()); does not look like failing

jglick: Ah, right, never mind.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/ApproverAction.java
```diff
@@ -0,0 +1,25 @@
+package org.jenkinsci.plugins.workflow.support.steps.input;
+
+import hudson.model.InvisibleAction;
+import org.kohsuke.stapler.export.Exported;
+import org.kohsuke.stapler.export.ExportedBean;
+
+/**
+ * Created by valentinaarmenise on 21/01/2015.
+ */
```
jglick: Can kill the dummy comment (use `@author` and/or add real Javadoc if you like), but please add a copyright header as in other source files.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/InputAction.java
```diff
@@ -3,10 +3,10 @@
 import hudson.model.Run;
 import jenkins.model.RunAction2;
 
+import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
-import javax.annotation.Nonnull;
```
jglick: No-op change could be reverted to minimize diff.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/InputStepExecution.java
```diff
@@ -41,16 +33,19 @@
...
+    private transient FlowNode node;
 
     /**
      * Result of the input.
      */
     private Outcome outcome;
 
-    @Inject(optional=true) InputStep input;
+    @Inject(optional=true)
+    InputStep input;
```
jglick: In general, PRs should avoid making semantically null changes to lines not otherwise related to the diff. Whitespace-only changes make `git blame` less useful, increase the chance of conflicts during `git merge` or `git cherry-pick`, etc. And of course it is harder to review a longer diff.

In particular, you should configure your IDE to _not_ automatically reformat files just because you are editing something in them.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/InputStepExecution.java
```diff
@@ -135,9 +131,12 @@ public HttpResponse doSubmit(StaplerRequest request) throws IOException, Servlet
      * REST endpoint to submit the input.
      */
     @RequirePOST
-    public HttpResponse doProceed(StaplerRequest request) throws IOException, ServletException, InterruptedException {
+    public HttpResponse doProceed(StaplerRequest request, User user) throws IOException, ServletException, InterruptedException {
```
jglick: `doProceed` is an implicit `@WebMethod` so I am not sure what the impact is of adding a new parameter (even if it is not marked `@QueryParameter`).

Anyway this seems unnecessary. The check for `User.current()` could just be in `doProceed` to begin with, right? Just as it is in `doAbort`.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/ApproverAction.java
```diff
@@ -0,0 +1,25 @@
...
+
+@ExportedBean
+public class ApproverAction extends InvisibleAction {
+
+    public ApproverAction(String approverName) {
+        this.approverName = approverName;
+    }
+
+    final private String approverName;
```
jglick: Conventional name (matching similar classes in core) would be `userId`, to make it clear that it is `User.id`.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/InputStepExecution.java
```diff
@@ -135,9 +131,12 @@ public HttpResponse doSubmit(StaplerRequest request) throws IOException, Servlet
...
         preSubmissionCheck();
-
+        if (user!=null){
+            run.addAction(new ApproverAction(user.getDisplayName()));
```
jglick: The `User.displayName` could change at any time, and cannot be used to uniquely identify this person. You should instead store the `User.id`. (That also allows the `summary.jelly` to hyperlink to the `/user/*` page, assuming the user still exists when it is rendered. [Here](https://github.com/jenkinsci/jenkins/blob/419d670cd0f672b0b3eba6ad4bcc97e8c6948424/core/src/main/resources/hudson/model/Cause/UserIdCause/description.jelly) is an example.)

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/InputStepExecution.java
```diff
@@ -200,7 +200,7 @@ private void postSettlement() throws IOException {
     }
 
     /**
-     * Parse the submitted {@link ParameterValue}s
+     * Parse the submitted {@link hudson.model.ParameterValue}s
```
jglick: Gratuitous; `ParameterValue` was already imported.

support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/input/ApproverAction/summary.jelly
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+The MIT License
+
+Copyright 2014 Jesse Glick.
```
jglick: !

jglick:
Did not get a chance to review this today, will try tomorrow.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepTest.java
```diff
@@ -64,5 +97,10 @@ public void parameter() throws Exception {
         // make sure 'x' gets assigned to false
         System.out.println(b.getLog());
         assertTrue(b.getLog().contains("after: false"));
+
+        //make sure the approver name corresponds to the submitter
+        ApproverAction action = b.getAction(ApproverAction.class);
+        assertNotNull(action);
+        assertEquals(is.getInput().getSubmitter(), action.getUserId());
```
varmenise: this is wrong: should be 
assertEquals(is.getInput().getSubmitter(), action.getUserName())
but when I run the mvn test, it does not fail.
Why is that?

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepTest.java
jglick: (resolved)

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepTest.java
jglick: `getUserId` is fine here.

You could simplify slightly to

```java
assertEquals("alice", action.getUserId());
```

but it does not really matter.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/ApproverAction.java
```diff
@@ -0,0 +1,58 @@
...
+@ExportedBean
+public class ApproverAction extends InvisibleAction {
+
+    public ApproverAction(String userId) {
+        this.userId = userId;
+        this.username=User.get(userId).getDisplayName();
+    }
+
+    final private String userId;
+    final private String username;
```
jglick: I would not store this as a field. Just do the lookup of `User.get(userId).getDisplayName()` on demand.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/ApproverAction.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * @author Valentina Armenise
+ */
+
+@ExportedBean
+public class ApproverAction extends InvisibleAction {
+
+    public ApproverAction(String userId) {
+        this.userId = userId;
+        this.username=User.get(userId).getDisplayName();
```
jglick: And what if `userId` is null? You check that in `getUserName` but that is too late to save this line!

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/ApproverAction.java
```diff
@@ -0,0 +1,58 @@
...
+
+    final private String userId;
+    final private String username;
+
+    @Exported
+    public String getUserId() {
+        return userId;
+    }
+
+    @Exported
```
jglick: Does not really to be `@Exported`. This is only for the benefit of `summary.jelly`.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/InputStepExecution.java
```diff
@@ -137,7 +137,11 @@ public HttpResponse doSubmit(StaplerRequest request) throws IOException, Servlet
     @RequirePOST
     public HttpResponse doProceed(StaplerRequest request) throws IOException, ServletException, InterruptedException {
         preSubmissionCheck();
-
+        User user = User.current();
+        if (user!=null){
```
jglick: If you are not adding an `ApproverAction` when the approver is anonymous (which I think is fine), then `ApproverAction` can declare `userId` as `@Nonnull` and skip the null checks, simplifying several things.

support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/input/ApproverAction/summary.jelly
```diff
@@ -0,0 +1,31 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-->
+
+<?jelly escape-by-default='true'?>
+ <j:jelly xmlns:j="jelly:core" xmlns:t="/lib/hudson">
+   <t:summary icon="orange-square.png">
+        ${%approved_by_user(it.userId, it.userName, rootURL)}.
```
jglick: The hyperlink is not going to work when you have `escape-by-default='true'`; `<` will become `&lt;`. You can solve that using `<j:out value="${…}"/>` but then you *must be careful* to call `h.escape(…)` on the user’s display name, so that they cannot launch a (stored) XSS attack.

It would be wise to use HtmlUnit (`JenkinsRule.WebClient`) to verify that the expected approval text, including valid hyperlink, in fact appears on the build index page.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/ApproverAction.java
jglick: (I would even suggest using `@Restricted(DoNotUse.class)` to make it clear that this is expected to be called only from the Jelly view, not from Java code.)

jglick:
This was initially filed some time ago but I am taking the liberty of assuming you meant it to be @reviewbybees.

support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/input/ApproverAction/summary.jelly
varmenise: I understand your point - but then, I am not sure why the code above works, [as well as] (https://github.com/jenkinsci/jenkins/blob/419d670cd0f672b0b3eba6ad4bcc97e8c6948424/core/src/main/resources/hudson/model/Cause/UserIdCause/description.jelly). 
screen shot 2015-03-04 at 12 24 34

jglick: Not sure why it works, but I guess if it does then this can be ignored for now.

jglick: Update: `InternationalizedStringExpression` specifically escapes parameters, but not the actual bundle key.

jglick:
Will need to merge with `master` since there is a merge conflict with #70.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/ApproverAction.java
```diff
@@ -0,0 +1,61 @@
...
+public class ApproverAction extends InvisibleAction {
+
+    public ApproverAction(String userId) {
+        this.userId = userId;
+    }
+
+    @Nonnull
+    final private String userId;
+
+    @Restricted(DoNotUse.class)
```
jglick: Still need to be `@Exported`, or there is no point in having the class as a whole be an `@ExportedBean`.

jglick:
:+1: at last!

varmenise:
@reviewbybees (need another +1)

PR #48, opened 2015-02-02
@RandomlyFails

jglick:
Noticed when doing a full build. Cause of failure unknown; passed the next time. @reviewbybees


stephenc:
:+1:

tfennelly:
:+1:

jtnord:
:+1: +1:

- need to discuss getting the flaky test plugin on jenkins and then moving to surefire with 2 retries...

jglick:
> moving to surefire with 2 retries

Still awaiting a PR to do this, or I can try it if you do not wish to.

jtnord:
@jglick I do not want to make the change until we have the flaky test plugin[1] on Jenkins - otherwise we would loose the ability to see these (right now they are skipped in the XML report).

[1] https://wiki.jenkins-ci.org/display/JENKINS/Flaky+Test+Handler+Plugin

PR #49, opened 2015-02-06
Display Name configuration

jeanblanchard:
Copy the Display Name configuration from the AbstractProject.
It is in AbstractItem, so it works without changing anything. The only thing missing was the configuration.

(Also I have to use Eclipse at work, so I added the .gitignore config for Eclipse)

### DIFF
--- a/job/src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowJob/configure-entries.jelly
+++ b/job/src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowJob/configure-entries.jelly
@@ -27,6 +27,15 @@
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form" xmlns:p="/lib/hudson/project">
     <p:config-quietPeriod/>
     <p:config-trigger/>
+
+    <f:section title="${%Advanced Project Options}">
+        <f:advanced>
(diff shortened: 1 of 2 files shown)


job/src/main/resources/org/jenkinsci/plugins/workflow/job/WorkflowJob/configure-entries.jelly
```diff
@@ -27,6 +27,15 @@
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form" xmlns:p="/lib/hudson/project">
     <p:config-quietPeriod/>
     <p:config-trigger/>
+
+    <f:section title="${%Advanced Project Options}">
+        <f:advanced>
+            <f:entry title="${%Display Name}" field="displayNameOrNull">
+                <f:textbox checkUrl="'${rootURL}/checkDisplayName?displayName='+encodeURIComponent(this.value)+'&amp;jobName='+encodeURIComponent('${h.jsStringEscape(it.name)}')"/>
+            </f:entry>
```
jglick: This `entry` block should ideally be defined as a reusable tag in `/lib/hudson/project` so we could call `<p:config-displayName/>`; bonus points for a core PR with that refactoring.

PR #50, opened 2015-02-17
Handle nonserializable causes of interruption

jglick:
Workaround until baseline is 1.591+. @reviewbybees (ref. ZD-24677)


tfennelly:
:+1:

PR #51, opened 2015-02-17
Avoid storing an empty array in Tag.actions.

jglick:
Minor memory optimization.

@reviewbybees (ref. ZD-24759)

### DIFF
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
@@ -46,10 +46,13 @@ import java.lang.ref.SoftReference;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 
 /**
  * {@link FlowNodeStorage} that stores one node per one file.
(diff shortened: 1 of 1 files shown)


support/src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
```diff
@@ -160,12 +164,16 @@ public boolean canConvert(Class type) {
...
 
-        private Tag(FlowNode node, List<Action> actions) {
+        private Tag(@Nonnull FlowNode node, @Nonnull List<Action> actions) {
             this.node = node;
-            this.actions = actions.toArray(new Action[actions.size()]);
+            this.actions = actions.isEmpty() ? null : actions.toArray(new Action[actions.size()]);
+        }
+
+        public @Nonnull List<Action> actions() {
+            return actions != null ? Arrays.asList(actions) : Collections.<Action>emptyList();
```
tfennelly: Why convert the `actions` `List` provided in the constructor to an array and then back to a `List` in the `actions` method?

jglick: Just because the array should take a bit less heap than a wrapping `List`.

stephenc:
:+1: from the blind man on the galloping horse

support/src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
jglick: To be clear, the _original_ code was already converting `List<Action>` to `Action[]`. The only real change I am making here is replacing `new Action[0]` with `null`. The other changes merely encapsulate the null check when serving the list back.

tfennelly:
:+1:

PR #53, opened 2015-02-20
Random test failure

jglick:
Observed once to fail saying that the build status was `null` rather than `FAILURE`, perhaps because the flow had suspended and was just about to complete but had not done so yet.

@reviewbybees


stephenc:
:+1:

tfennelly:
:+1:

PR #54, opened 2015-02-20
Handle FlowInterruptedException from CatchErrorStep

jglick:
Previously was marking a build as failed when it was merely supposed to be canceled due to a superseded stage, etc.

@reviewbybees (ref. ZD-24677)


tfennelly:
:+1:

PR #57, opened 2015-02-21
[JENKINS-26761] Verifying that polling still works after a restart

jglick:
Failed attempt at reproducing [JENKINS-26761](https://issues.jenkins-ci.org/browse/JENKINS-26761); useful to have test anyway to guard against other problems.

@reviewbybees


tfennelly:
:+1:

PR #58, opened 2015-02-22
[JENKINS-25779] DescribableHelper.uninstantiate and default values

jglick:
[JENKINS-25779](https://issues.jenkins-ci.org/browse/JENKINS-25779)

@reviewbybees

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ Only noting significant user-visible or major API changes, not internal code cle
 ## 1.3 (upcoming)
 
 ### User changes
+* JENKINS-25779: snippet generator now omits default values of complex steps.
 * Ability to configure project display name.
 * Fixing `java.io.NotSerializableException: org.jenkinsci.plugins.workflow.support.steps.StageStepExecution$CanceledCause` thrown from certain scripts using `stage`.
 * JENKINS-26605: Missing link to _Full Log_ under _Running Steps_ when a single step produced >150Kb of output.
--- a/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
+++ b/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
@@ -64,7 +64,6 @@ import org.kohsuke.stapler.StaplerResponse;
                 StringBuilder b = new StringBuilder(d.getFunctionName());
                 Step step = (Step) o;
                 Map<String,Object> args = new TreeMap<String,Object>(d.defineArguments(step));
-                args.values().removeAll(Collections.singleton(null)); // do not write null values
                 boolean first = true;
                 boolean singleMap = args.size() == 1 && args.values().iterator().next() instanceof Map;
                 for (Map.Entry<String,Object> entry : args.entrySet()) {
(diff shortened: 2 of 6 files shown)


stephenc:
It's all greek to me!

jglick:
So I probably need to explain better. In fact `DescribableHelper` is a standalone class (no non-core dependencies) so it is not too hard to cover. Basically this takes a data-bound object with some configuration and converts it into a JSONish representation of the configuration: so `Map<String,Object>` for bean properties, `List` for arrays/lists, `String` or primitives for basic values including `enum`s, etc. It also handles the `$class` syntax to select an implementation of an abstract extension point.

Workflow uses it in two ways: when you write a call to a step, `instantiate` is used to take your Groovy data expression, e.g.

```groovy
somestep mode: 'NORMAL', handlers: [[$class: 'WhateverHandler', useStuff: true]]
```

and translate it into an actual `Step` object, as if you had written out the API calls in Java syntax:

```java
SomeStep s = new Somestep(Somestep.Mode.NORMAL);
s.setHandlers(Collections.<Handler>singletonList(new WhateverHandler(true)));
```

Conversely, the Snippet Generator lets you configure `SomeStep/config.jelly` and then `uninstantiate` is used to translate that back into some maps/lists/primitives (which `Snippetizer` just renders in Groovy syntax).

The problem was with steps (or data-bound structures used somewhere inside a step definition) that had object-valued properties with non-null defaults. Notoriously, `InputStep` only requires one property—`String message`—but has two other properties (`ok` and `parameters`) with non-null default values, and another (`id`) with a default value _computed_ from `message`. So when you typed in a simple message in the form and asked for a DSL snippet, you got a long messy statement, most of which was actually redundant. This is confusing to new users in particular: they assume they need to copy and paste the whole thing into their script and there is no indication that some parts are optional. With this patch, you get a simple

```groovy
input 'Hello?'
```

unless you actually configured something in the _Advanced_ block.

tfennelly:
I think I get it for the most part so :+1: 

I do think some of the data bound getter checking/inspect code looks a bit incomplete (e.g. assumption about method names always starting with "get"), but that wasn't part of this changeset and I suppose once Jenkins people know about this they will always use "get*" and not "is*" etc.

stephenc:
:+1: from the blind man on the galloping horse (I know I need to dig deeper into this, but I have other stuff to work on)

jglick:
> I do think some of the data bound getter checking/inspect code looks a bit incomplete

There are some known bugs and surely some unknown bugs.

jglick:
BTW @tfennelly please comment on the aggregate diff (`files` link from the PR), not on an individual commit inside the PR. This ensures that your comments appear inline in the appropriate place and are automatically hidden if obsoleted by a follow-up commit.

tfennelly:
@jglick Ah I was wondering why it appeared like that - I thought it was some new Github thing :smile:. I probably did exactly as you suggest coz I sometimes go through the changes one commit at a time as it can be easier to see what's happening. I think you might have done this too coz I seem to remember a similar review comment from yourself in a different PR.

Anyway ... :+1:

stephenc:
:+1: from TBMOAGH

jglick:
> I think you might have done this too

By accident, yes I think so.

PR #59, opened 2015-02-23
[WiP] Make tests pass on Windows

jglick:
- [x] `CpsFlowDefinitionTest`
- [X] `LogActionTest`
- [X] `SnippetizerTest`
- [x] `WorkflowLibRepositoryTest`
- [x] `LoadStepTest`
- [x] `RestartingLoadStepTest`
- [x] `ArtifactArchiverStepTest`
- [x] `CoreStepTest`
- [x] `ShellStepTest`
- [x] `ParallelStepTest` (subsumes #60)
- [x] `PushdStepTest`
- [x] `PwdStepTest`
- [x] `ReadWriteFileStepTest`
- [x] `GitStepRestartTest`
- [x] `GitStepTest`
- [ ] `StageTest`
- [ ] `ToolStepTest`
- [ ] `WaitForConditionStepTest`
- [ ] `WorkflowJobNonRestartingTest`
- [ ] `WorkflowRunTest`
- [ ] `WorkflowTest`
- [ ] `FlowExecutionList.xml` lock problem from `ShellStepTest` (does not cause test failure)
- [x] `TimeoutStepTest` (probably passes only by accident)
- [ ] `BuildTriggerStepTest` (ditto; check for usages of `Shell`)
- [x] `SubversionStepTest` (perhaps @jtnord did not have `svn` in `%Path%`?)
- [X] `MercurialStepTest` (ditto) (still fails under some drive-letter conditions that I do not want to bother sorting out now)

Some overlap with [JENKINS-25975](https://issues.jenkins-ci.org/browse/JENKINS-25975).

@reviewbybees esp. @jtnord

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -35,6 +35,7 @@ See [this guide](basic-steps/CORE-STEPS.md#adding-support-from-plugins) for maki
 - [ ] `HtmlPublisher` (`htmlpublisher`): [JENKINS-26343](https://issues.jenkins-ci.org/browse/JENKINS-26343)
 - [ ] `JaCoCoPublisher` (`jacoco`): [JENKINS-27120](https://issues.jenkins-ci.org/browse/JENKINS-27120)
 - [ ] `Publisher` (`testng`): [JENKINS-27121](https://issues.jenkins-ci.org/browse/JENKINS-27121)
+- [ ] `GroovyPostbuildRecorder` (`groovy-postbuild`): [JENKINS-26918](https://issues.jenkins-ci.org/browse/JENKINS-26918)
 
 ## Build wrappers
 
@@ -42,12 +43,15 @@ See [this guide](basic-steps/CORE-STEPS.md#adding-support-from-plugins) for maki
 - [ ] `ConfigFileBuildWrapper` (`config-file-provider`): [JENKINS-26339](https://issues.jenkins-ci.org/browse/JENKINS-26339)
 - [ ] `Xvnc` (`xvnc`): [JENKINS-26477](https://issues.jenkins-ci.org/browse/JENKINS-26477)
 - [ ] `BuildUser` (`build-user-vars`): [JENKINS-26953](https://issues.jenkins-ci.org/browse/JENKINS-26953)
+- [ ] `DashboardBuilder` (`environment-dashboard`): [PR 20](https://github.com/vipinsthename/environment-dashboard/issues/20)
 
 ## Triggers
 
 Implement `Trigger<ParameterizedJobMixIn.ParameterizedJob>` and implement `TriggerDescriptor.isApplicable` accordingly.
 
 - [ ] `gerrit-trigger`: [JENKINS-26010](https://issues.jenkins-ci.org/browse/JENKINS-26010)
+- [ ] `ghprb`: [JENKINS-26591](https://issues.jenkins-ci.org/browse/JENKINS-26591)
+- [ ] `github`: [JENKINS-27136](https://issues.jenkins-ci.org/browse/JENKINS-27136)
 
 ## Clouds
 
@@ -68,6 +72,8 @@ Do not necessarily need any special integration, but are encouraged to use `Once
 - [X] `credentials-binding`: `withCredentials` step as of 1.3
 - [ ] `build-failure-analyzer`: [JENKINS-27123](https://issues.jenkins-ci.org/browse/JENKINS-27123)
 - [ ] `shelve-project`: [JENKINS-26432](https://issues.jenkins-ci.org/browse/JENKINS-26432)
+- [X] `job-dsl`: implemented in 1.29
+- [ ] `zentimestamp`: [JENKINS-26958](https://issues.jenkins-ci.org/browse/JENKINS-26958)
 
 ## Custom steps
 


cps/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
```diff
@@ -0,0 +1,59 @@
+/*
```
jtnord: :100: for replacing groovy with java

cps/src/test/groovy/org/jenkinsci/plugins/workflow/cps/LogActionTest.groovy
```diff
@@ -51,6 +51,6 @@ echo("Hello I'm Gilbert")
         assert exec.isComplete()
         FlowNode atom = exec.currentHeads[0].parents[0]
         LogActionImpl la = atom.getAction(LogAction)
-        assert la.logFile.text=="Hello I'm Gilbert\n"
+        assert la.logFile.text.trim() == "Hello I'm Gilbert"
```
jtnord: this brings an intresting question - what will happen if you have both an echo in a windows node and an echo in a Linux node?
WIll the console render correctly - will it contain spruious \r if the main task in on a linux host  - or will lines be munged if this is on a Windows host?

cps/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
```diff
@@ -0,0 +1,59 @@
...
+        createExecution(flow);
+        exec.start();
+
+        SemaphoreStep.waitForStart("watch/1", null);
+
+        assertFalse("Expected the execution to be suspended but it has completed", exec.isComplete());
+
+        exec = roundtripXStream(exec);    // poor man's simulation of Jenkins restart
+        exec.onLoad();
+
```
jtnord: it would be good to see that the roundtrip leaves this still in the waiting state.

jtnord:
1 question inline - otherwise lgtm.
Let me know if you want me to re-run the tests.

cps/src/test/groovy/org/jenkinsci/plugins/workflow/cps/LogActionTest.groovy
jglick: Good question, try it please!

cps/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
jglick: Not sure offhand how to do that, and anyway the original test did not check it so out of scope.

jtnord: agreed out of scope.

wouldn't a check on exec.isComplete() do this though?

jglick: Might; feel free to try that after this is merged.

jglick:
@reviewbybees interim: I would like to merge the last-known-good commit to `master`, since for this PR there is no harm in only some tests having been corrected but not all, and it may be helpful.

stephenc:
@jglick :+1:  on interim merge

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
```diff
@@ -27,17 +25,15 @@
...
      */
-    @RandomlyFails("TODO sometimes the whole build just hangs (for longer than the 10s)")
     @Test
     public void basic() throws Exception {
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
         p.setDefinition(new CpsFlowDefinition(
-                "node { timeout(time:5, unit:'SECONDS') { sh 'sleep 10'; echo 'NotHere' } }"));
-        WorkflowRun b = assertBuildStatus(Result.FAILURE, p.scheduleBuild2(0));
+                "node { timeout(time:5, unit:'SECONDS') { sleep 10; echo 'NotHere' } }"));
+        WorkflowRun b = assertBuildStatus(Result.ABORTED, p.scheduleBuild2(0));
```
jtnord: why has this changed - I don't see a corresponding TimeoutStep change?

as a user I would expect a timeout to be reported as failure (which matches IIRC standard Jenkins behaviour).
something has failed.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
```diff
@@ -45,27 +41,27 @@ public void killingParallel() throws Exception {
...
-                            " b: { echo 'ShouldBeHere2'; sh 'sleep 10'; echo 'NotHere' }, ",
+                            " a: { echo 'ShouldBeHere1'; sleep 10; echo 'NotHere' }, ",
+                            " b: { echo 'ShouldBeHere2'; sleep 10; echo 'NotHere' }, ",
                         ");",
                         "echo 'NotHere'",
                     "}",
                     "echo 'NotHere'",
                 "}")));
-        WorkflowRun b = assertBuildStatus(Result.FAILURE, p.scheduleBuild2(0));
+        WorkflowRun b = assertBuildStatus(/* TODO JENKINS-25894 should really be ABORTED */Result.FAILURE, p.scheduleBuild2(0));
```
jtnord: I disagree here - the symtop should not change the cause.

jtnord:
see inline comments, however I am ok on an interim merge and prepared for a ~~clazz~~ result status war

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
jglick: > why has this changed

In fact the build status was _already_ `ABORTED`, but due to a mistake in @kohsuke’s original code this was not being checked. I fixed that (later in this test), and corrected the expected status.

> as a user I would expect a timeout to be reported as failure

The `timeout` step intentionally sets the build result to `ABORTED`. Proposed changes to that behavior are off topic here.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
jglick: Ditto above; the intention of code is for it to be `ABORTED`, but the wrapping in `ParallelStepException` breaks the transmission of this status.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
jglick: > standard Jenkins behaviour

FWIW the `build-timeout` plugin has a complex set of `BuildTimeOutOperation`s; the default seems to be to set the result to `ABORTED`. Unlike in a freestyle project, a workflow can customize timeout handling by just catching `InterruptedException` and doing something else, such as `error`.

jglick:
@reviewbybees for another interim merge, as I may not have time to finish this in the near future, and do not want merge conflicts in tests with unrelated work.

jtnord:
:-1: on interim merge as GitHub is as usefull as a chocolate fireguard for tracking differences between various commits and being able to comment on them.

:+1: to just merge it and do the follow up changes in a different PR.

jglick:
Was just a pain having this open. Can refer to the open checklist items from another PR as needed.

PR #60, opened 2015-02-24
Sleep step

jglick:
[JENKINS-26120](https://issues.jenkins-ci.org/browse/JENKINS-26120)

@reviewbybees

### DIFF
--- a/basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/SleepStep.java
+++ b/basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/SleepStep.java
@@ -1,0 +1,122 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 Jesse Glick.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.plugins.workflow.steps;
+
+import com.google.inject.Inject;
+import hudson.Extension;
+import hudson.util.ListBoxModel;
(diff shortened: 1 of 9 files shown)


basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/SleepStep.java
```diff
@@ -0,0 +1,122 @@
...
+
+        private static final long serialVersionUID = 1L;
+
+        @Inject(optional=true) private transient SleepStep step;
+        private long end;
+        private transient volatile ScheduledFuture<?> task;
+
+        @Override public boolean start() throws Exception {
+            long now = System.currentTimeMillis();
+            end = now + step.unit.toMillis(step.time);
```
tfennelly: I assume the fact that `step` is marked `optional=true` means it can, in theory, end up being null, right? If so, does it's use need to be checked?

jglick: See the sordid explanation in `AbstractStepExecutionImpl` (and I curse @kohsuke for plaguing this plugin with Guice). Since I only need `step` inside `start` (to compute `end`), `optional=true` instructs the system to leave `step` as null after a Jenkins restart if the `Execution` is resumed. If I needed the `step` later (as some `StepExecution`s do), I would make it non-`optional`, not `transient`, and would have to make the `Step` implement `Serializable`.

stephenc: > I curse @kohsuke for plaguing this plugin with Guice

What do you expect from the Guice fanboy. Excessive use of Guice considered harmful.

(Just like goto has some legitimate and useful use cases, the same for Guice... but you need to clearly define the layer at which Guice is being applied and be consistent in that application otherwise you end up with a similar mess to the cross layer EJB issue that plagued JavaEE apps)

tfennelly:
:+1:

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/SleepStep.java
tfennelly: As a general rule I am not a fan of "magic" code e.g. CDI, Aspects (anyone use them at all now?) and the like. Sure they can make the code easier to read if you're just interested in getting a general feel for how something works, which is great as long as everything works. If there's a problem though, debugging etc become a pitarse.

jglick: > I am not a fan of "magic" code

Yes, my feeling too. @kohsuke wanted a “convention-over-configuration” system. I did push back on one point: the original no-arg constructor for `AbstractStepDescriptorImpl`, which inferred the `StepExecution` type by appending `"Execution"` to the `Step` implementation class name, meaning a rename refactoring could cause runtime failures.

By way of background, there is one key advantage of the use of Guice in step definitions, which is that `AbstractStepDescriptorImpl.getRequiredContext` automatically stays synchronized with the `@StepContextParameter`s you are actually requesting in your `AbstractStepExecutionImpl`. (Even this is not quite as watertight as it may seem, because you can still call `StepContext.get(Class)` directly; and some steps actually need to do so because the call does not come from the `StepExecution` itself, but rather from some other object, such as a `BodyExecutionCallback` (which is passed a `StepContext` in its methods). For these steps, there is sometimes a `@StepContextParameter` field which is not actually used, just to force `getRequiredContext` to list that type.)

The other uses of Guice more or less boil down to your not needing to explicitly call constructors (or write code in `onResume`), but I find the LoC reduction to be modest—not really justifying the high risks of injection, such as the “throw-up-your-hands-in-despair" exception in [JENKINS-26513](https://issues.jenkins-ci.org/browse/JENKINS-26513).

At any rate, the use of Guice is optional here, so if there is consensus on this point we can “deguice” some steps as a proof of concept, and if the result seems preferable, just deprecate the API types using it: `AbstractStepDescriptorImpl`, `AbstractStepExecutionImpl`, `AbstractStepImpl`, and `StepContextParameter`. `AbstractSynchronousStepExecution` would also need a replacement, though [JENKINS-25879](https://issues.jenkins-ci.org/browse/JENKINS-25879) suggests it needs to be redesigned anyway.

PR #61, opened 2015-02-27
[JENKINS-26619] DescribableHelper does not handle GitSCMExtension

ndeloof:
[JENKINS-26619](https://issues.jenkins-ci.org/browse/JENKINS-26619)

@reviewByBees this also cover #685

### DIFF
--- a/step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
+++ b/step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
@@ -200,7 +200,9 @@ public class DescribableHelper {
 
     @SuppressWarnings("unchecked")
     private static boolean isList(Type type) {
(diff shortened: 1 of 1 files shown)


jglick:
@ndeloof I suggest starting PRs from a forked repo to avoid bugs in (a) the JIRA notification daemon and (b) PR builder; except for complex changes where you expect multiple committers.

jglick:
This causes multiple test regressions. Anyway any change to `DescribableHelper` must be accompanied by a new unit test in `DescribableHelperTest`. They are not difficult to write.

ndeloof:
I indeed have seen test regression, don't understand how this could work using isAssignableFrom the wrong way. Will run them in a debugger.

jglick:
Well I would start by reverting your current change, then writing a test reproducing the condition described in the issue.

jglick:
It is not necessary to close a PR, you can just push follow-up commits as needed.

jglick:
`isAssignableFrom` was in fact in the right direction (the method name was misleading); the problem was related but different.

PR #63, opened 2015-02-27
[JENKINS-26123] Non-blocking build trigger step

jglick:
[JENKINS-26123](https://issues.jenkins-ci.org/browse/JENKINS-26123)

Also threw in a fix for a related issue I saw in this code.

@reviewbybees

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/build/BuildTriggerStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/build/BuildTriggerStepTest.java
@@ -1,5 +1,6 @@
 package org.jenkinsci.plugins.workflow.steps.build;
 
+import hudson.model.Action;
 import hudson.model.BooleanParameterDefinition;
 import hudson.model.Cause;
 import hudson.model.FreeStyleBuild;
@@ -23,7 +24,9 @@ import org.jvnet.hudson.test.JenkinsRule;
 import org.jvnet.hudson.test.MockFolder;
 
 import java.util.Arrays;
+import java.util.List;
 import org.jvnet.hudson.test.Issue;
+import org.jvnet.hudson.test.TestExtension;
 
 /**
  * @author Vivek Pandey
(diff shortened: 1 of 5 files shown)


aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/build/BuildTriggerStepTest.java
```diff
@@ -165,4 +170,25 @@ public void cancelBuildQueue() throws Exception {
         j.assertLogContains("branch=release extra=true", ds.getBuildByNumber(3));
     }
 
+    @Issue("JENKINS-26123")
+    @Test public void noWait() throws Exception {
+        j.createFreeStyleProject("ds").setAssignedLabel(Label.get("nonexistent"));
+        WorkflowJob us = j.jenkins.createProject(WorkflowJob.class, "us");
+        us.setDefinition(new CpsFlowDefinition("build job: 'ds', wait: false"));
+        j.assertBuildStatusSuccess(us.scheduleBuild2(0));
```
jtnord: AS the schedule will be rejected by the TestExtension (rejectedStart) isn;t this just a test that has a race condition and may randomly fail?

jglick: Scheduling the `WorkflowJob` will succeed. Its attempt to schedule the `FreeStyleProject` will fail. This causes `scheduleBuild2` to return `null`, which needs to make the step fail, which is the bug I fixed.

jtnord: if that's the case then why are we asserting that our build was a success - shouldn't it fail?

tfennelly: @jtnord I think it's the downstream job that fails, which should be irrelevant to the Workflow job since it's not waiting on (`wait: false`), right?

tfennelly:
:+1: assuming that test is ok

jtnord:
:+1:

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/build/BuildTriggerStepTest.java
jtnord: was my confusion - I incorrectly assumed the value of the TestExtension was to aid debuggin in Jenkins - not to restrict it to only be injected into the named test.

jglick: Cleared up some confusion on the phone: `QDH` is active during the `rejectedStart` test, _not_ during the `noWait` test. For the first test, `ds` is successfully added to the queue (but never builds), so `us` succeeds; for the second test, `ds` is not even added to the queue, so `us` fails.

PR #62, opened 2015-02-27
[JENKINS-26513] Deserialization error in ExecutorStepExecution

jglick:
[JENKINS-26513](https://issues.jenkins-ci.org/browse/JENKINS-26513)

Guice-related error, sigh.

@reviewbybees

PR #65, opened 2015-03-02
[JENKINS-25890] Deadlock

jglick:
[JENKINS-25890](https://issues.jenkins-ci.org/browse/JENKINS-25890)

Reproduced by a revision of the test added in #62.

Removing the exceptions from the declaration of `isReady` to emphasize that it is not intended to block.

@reviewbybees esp. @kohsuke


rsandell:
:+1:

tfennelly:
:+1:

PR #70, opened 2015-03-03
[FIXED JENKINS-26363] Input step should permit cancellation from anyone with Job.CANCEL permission

tfennelly:
See [JENKINS-26363](https://issues.jenkins-ci.org/browse/JENKINS-26363)

Also moved `InputStep.canSubmit` and `InputStep.canSettle` to `InputStepExecution` (and made them private there). I deprecated those methods in `InputStep` i.e. didn't remove them in case they are somehow in use from outside. Can remove them if we think they're not.

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepTest.java
@@ -2,8 +2,16 @@ package org.jenkinsci.plugins.workflow.steps.input;
 
 import com.gargoylesoftware.htmlunit.html.HtmlPage;
 import hudson.model.BooleanParameterDefinition;
+import hudson.model.Job;
+import hudson.model.User;
 import hudson.model.queue.QueueTaskFuture;
+
 import java.util.Arrays;
+import java.util.List;
+
+import hudson.security.ACL;
+import hudson.security.GlobalMatrixAuthorizationStrategy;
+import jenkins.model.Jenkins;
 import org.apache.commons.lang.StringUtils;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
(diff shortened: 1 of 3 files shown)


jglick:
Was this intended to be @reviewbybees?

tfennelly:
Oh yeah thanks ... @reviewbybees

:smile:

tfennelly:
@reviewbybees or the 12hr rule will be invoked

(just kidding about the 12hr rule :smile: )

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepTest.java
```diff
@@ -65,4 +74,77 @@ public void parameter() throws Exception {
...
+        addUserWithPrivs("charlie", authorizationStrategy);
+        authorizationStrategy.add(Job.CANCEL, "charlie");
+
+        j.jenkins.setAuthorizationStrategy(authorizationStrategy);
+
+        final WorkflowJob foo = j.jenkins.createProject(WorkflowJob.class, "foo");
+        ACL.impersonate(User.get("alice").impersonate(), new Runnable() {
+            @Override
+            public void run() {
+                foo.setDefinition(new CpsFlowDefinition("input id: 'InputX', message: 'OK?', ok: 'Yes', submitter: 'alice'"));
```
jglick: BTW this `impersonate` and granting `RUN_SCRIPTS` would probably be unnecessary if you just passed `true` for `sandbox`.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepTest.java
```diff
@@ -65,4 +74,77 @@ public void parameter() throws Exception {
...
+        try {
+            j.submit(p.getFormByName(is.getId()), "abort");
+            assertEquals(0, inputAction.getExecutions().size());
+            queueTaskFuture.get();
+
+            List<String> log = run.getLog(1000);
+            System.out.println(log);
+            assertTrue(expectAbortOk);
+            assertEquals("Finished: ABORTED", log.get(log.size() - 1)); // Should be aborted
+        } catch (Exception e) {
```
jglick: `FailingHttpStatusCodeException` I guess?

jglick:
:+1: other than those minor comments.

PR #71, opened 2015-03-04
[FIXED JENKINS-25958] Check for already known Computers/Channels in FilePathPickle.Listener

tfennelly:
Hoping this is enough to fix [JENKINS-25958](https://issues.jenkins-ci.org/browse/JENKINS-25958). Is there a way we can easily test this?

@reviewbybees

### DIFF
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/actions/WorkspaceActionImpl.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/actions/WorkspaceActionImpl.java
@@ -52,9 +52,9 @@ public final class WorkspaceActionImpl extends WorkspaceAction implements FlowNo
 
     public WorkspaceActionImpl(FilePath workspace, FlowNode parent) {
         // TODO see FilePathPickle:
-        node = FilePathPickle.Listener.channelNames.get(workspace.getChannel());
+        node = FilePathPickle.Listener.getChannelName(workspace.getChannel());
         if (node == null) {
-            throw new IllegalStateException("no known slave for " + workspace + " among " + FilePathPickle.Listener.channelNames.values());
+            throw new IllegalStateException("no known slave for " + workspace + " among " + FilePathPickle.Listener.getChannelNames());
         }
         Jenkins j = Jenkins.getInstance();
         Node n = j == null ? null : node.isEmpty() ? j : j.getNode(node);
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
@@ -86,6 +86,18 @@ public class FilePathPickle extends Pickle {
         // TODO better to use a synchronized accessor
         @Restricted(NoExternalUse.class)
         public static final Map<VirtualChannel,String> channelNames = new WeakHashMap<VirtualChannel,String>();
+
+        static {
+            // Add channels that may already be known to Jenkins. This can be the case
+            // just after the Workflow plugins are installed (see JENKINS-25958).
+            Jenkins instance = Jenkins.getInstance();
+            if (instance != null) {
+                for (Computer c : instance.getComputers()) {
+                    channelNames.put(c.getChannel(), c.getName());
+                }
+            }
+        }
+
         @Override public void onOnline(Computer c, TaskListener l) { // TODO currently preOnline is not called for MasterComputer
             if (c instanceof Jenkins.MasterComputer) {
                 channelNames.put(c.getChannel(), c.getName());
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
@@ -34,22 +34,31 @@ import hudson.remoting.Channel;
 import hudson.remoting.VirtualChannel;
 import hudson.slaves.ComputerListener;
 import java.io.IOException;
+import java.util.Collection;
 import java.util.Map;
 import java.util.WeakHashMap;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
 import jenkins.model.Jenkins;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+import javax.annotation.Nonnull;
 
 /**
  * @author Kohsuke Kawaguchi
  */
 public class FilePathPickle extends Pickle {
+
+    private static final Logger LOGGER = Logger.getLogger(FilePathPickle.class.getName());
+
     private final String slave;
     private final String path;
 
     private FilePathPickle(FilePath v) {
         // TODO JENKINS-26096 switch to FilePath.toComputer once it uses something like Listener here and so is actually reliable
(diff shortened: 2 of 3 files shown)


support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
```diff
@@ -86,6 +86,18 @@ protected FilePath tryResolve() {
...
         @Restricted(NoExternalUse.class)
         public static final Map<VirtualChannel,String> channelNames = new WeakHashMap<VirtualChannel,String>();
+
+        static {
+            // Add channels that may already be known to Jenkins. This can be the case
+            // just after the Workflow plugins are installed (see JENKINS-25958).
+            Jenkins instance = Jenkins.getInstance();
+            if (instance != null) {
+                for (Computer c : instance.getComputers()) {
+                    channelNames.put(c.getChannel(), c.getName());
```
jglick: `getChannel` might return null IIRC.

support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
```diff
@@ -86,6 +86,18 @@ protected FilePath tryResolve() {
         // TODO better to use a synchronized accessor
         @Restricted(NoExternalUse.class)
         public static final Map<VirtualChannel,String> channelNames = new WeakHashMap<VirtualChannel,String>();
+
+        static {
+            // Add channels that may already be known to Jenkins. This can be the case
+            // just after the Workflow plugins are installed (see JENKINS-25958).
+            Jenkins instance = Jenkins.getInstance();
+            if (instance != null) {
+                for (Computer c : instance.getComputers()) {
```
jglick: Just make sure this includes `MasterComputer` (I think it does).

jglick:
To test: rebuild all plugins as snapshots, then install in a virgin 1.580.1 using _Plugin Manager » Advanced » Upload_ in dependency order. Or there is a CLI command which might be more convenient.

jglick:
As @jtnord notes, the simpler patch is to add `(dynamicLoadable=YesNoMaybe.NO)` to the `Listener` extension point. But admins may really prefer to use the plugin immediately.

support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
```diff
@@ -86,6 +86,18 @@ protected FilePath tryResolve() {
         // TODO better to use a synchronized accessor
         @Restricted(NoExternalUse.class)
         public static final Map<VirtualChannel,String> channelNames = new WeakHashMap<VirtualChannel,String>();
+
+        static {
+            // Add channels that may already be known to Jenkins. This can be the case
```
jglick: Alternate idea (might be more robust): encapsulate the lookups of `channelNames` (in `WorkspaceActionImpl`, `WorkspaceListLeasePickle`, and `FilePathPickle` itself) into a method which before returning `null` (or throwing `IllegalStateException`, as all currently impls do) would fall back to doing this inspection of the currently online `Computer`s.

jglick:
Extra credit: file a core PR to copy your impl to `FilePath.toComputer` to fix [JENKINS-26096](https://issues.jenkins-ci.org/browse/JENKINS-26096).

tfennelly:
@jglick tested this fix so far and it fixes the issue (without requiring a Jenkins restart). Will now look at the other things you were proposing.

support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
tfennelly: It does.

support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
tfennelly: I'll check for it but the existing code was not checking for it.

support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
tfennelly: @jglick We can do that but not sure I see how it's more robust.

As a matter of interest - what is this listener giving anyway i.e. why doesn't `WorkspaceActionImpl` etc just call something that iterates the list of `Computer`s directly off the `Jenkins` instance? It's all in memory already so I don't see the benefit of creating a secondary list.

jglick: > how it's more robust

Handles the case that `onOnline` was not delivered for some reason.

> iterates the list of `Computer`s directly

See JENKINS-26096 for why that does not work.

support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
jglick: > the existing code was not checking for it

Because the existing code was in a listener callback which is called as soon as the channel is constructed and set. This code is not.

support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
```diff
@@ -83,16 +92,53 @@ protected FilePath tryResolve() {
...
+            String channelName = channelNames.get(channel);
+
+            if (channelName == null) {
+                // We don't have a cache entry for the name of the Computer associated with the supplied Channel
+                // instance. Lets fallback and search the list of Computers on Jenkins.
+                Jenkins jenkins = Jenkins.getInstance();
+                if (jenkins != null) {
+                    for (Computer computer : jenkins.getComputers()) {
+                        VirtualChannel computerChannel = computer.getChannel();
+                        if (computerChannel != null && computerChannel.equals(channel)) {
```
jglick: `channel = computer.getChannel()` should suffice.

support/src/main/java/org/jenkinsci/plugins/workflow/support/actions/WorkspaceActionImpl.java
```diff
@@ -52,9 +52,9 @@
 
     public WorkspaceActionImpl(FilePath workspace, FlowNode parent) {
         // TODO see FilePathPickle:
-        node = FilePathPickle.Listener.channelNames.get(workspace.getChannel());
+        node = FilePathPickle.Listener.getChannelName(workspace.getChannel());
         if (node == null) {
-            throw new IllegalStateException("no known slave for " + workspace + " among " + FilePathPickle.Listener.channelNames.values());
+            throw new IllegalStateException("no known slave for " + workspace + " among " + FilePathPickle.Listener.getChannelNames());
```
jglick: You can just make `getChannelName` be `@Nonnull` and throw `IllegalStateException`; then the callers are simpler, and you do not need `getChannelNames` since that is only used for the exception message.

jglick:
Minor suggestions but :+1: good enough if you want to stop here.

tfennelly:
@jglick Might leave as is for now. Thanks.

If the `IllegalStateException` is thrown inside `getChannelName` then it's harder to provide a decent exception message. In the case of `WorkspaceActionImpl`, you'd need to pass in the `workspace` name to get the message to the same level of detail in the error.

PR #75, opened 2015-03-05
[JENKINS-27145] Linking to env-vars.html from UI

tfennelly:
See [JENKINS-27145](https://issues.jenkins-ci.org/browse/JENKINS-27145).

Please check the variables I listed as not being available outside a `node`. And if there are other exceptions that should be noted in the text help.

### DIFF
--- a/cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/help-script.html
+++ b/cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/help-script.html
@@ -34,15 +34,12 @@ below the list for Workflow specific details.
     <iframe src="${rootURL}/env-vars.html" width="100%"></iframe>
 </p>
 
-The following variables are currently not available inside a workflow script:
+The following variables are currently unavailable inside a workflow script:
 <ul>
     <li>EXECUTOR_NUMBER</li>
     <li>NODE_NAME</li>
     <li>NODE_LABELS</li>
     <li>WORKSPACE</li>
-    <li>JENKINS_URL</li>
-    <li>BUILD_URL</li>
-    <li>JOB_URL</li>
 </ul>
 
 <h4>Using Environment Variables</h4>
@@ -55,7 +52,7 @@ constructing email content when using the <code>mail</code> step:
 <pre>
     mail (to: 'devops@acme.com',
         subject: "Job '${env.JOB_NAME}' (${env.BUILD_NUMBER}) is waiting for input",
-        body: "Please go to http://builds.acme.ccom:8080/jenkins/job/foo/${env.BUILD_NUMBER}/.");
+        body: "Please go to ${env.BUILD_URL} and verify the build");
 </pre>
 
 <p/>
--- a/cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/help-script.html
+++ b/cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/help-script.html
@@ -34,28 +34,31 @@ below the list for Workflow specific details.
     <iframe src="${rootURL}/env-vars.html" width="100%"></iframe>
 </p>
 
-The following variables are only available inside <code>node</code>s (i.e. not available outside):
+The following variables are currently not available inside a workflow script:
 <ul>
     <li>EXECUTOR_NUMBER</li>
     <li>NODE_NAME</li>
     <li>NODE_LABELS</li>
     <li>WORKSPACE</li>
+    <li>JENKINS_URL</li>
+    <li>BUILD_URL</li>
+    <li>JOB_URL</li>
 </ul>
 
 <h4>Using Environment Variables</h4>
 
-Environemnt variables are injected into scripts through a variable named "<strong>env</strong>". This variable,
+Environment variables are injected into scripts through a variable named "<strong>env</strong>". This variable,
 like any other variable, can be used in the general flow of the script, or in variable substitutions e.g. when
 constructing email content when using the <code>mail</code> step:
 
-<p>
+<p/>
 <pre>
     mail (to: 'devops@acme.com',
         subject: "Job '${env.JOB_NAME}' (${env.BUILD_NUMBER}) is waiting for input",
-        body: "Please go to ${env.BUILD_URL}.");
+        body: "Please go to http://builds.acme.ccom:8080/jenkins/job/foo/${env.BUILD_NUMBER}/.");
 </pre>
-</p>
 
+<p/>
 For more on environment variables,
-<a href="https://github.com/jenkinsci/workflow-plugin/blob/master/TUTORIAL.md#managing-the-environment">see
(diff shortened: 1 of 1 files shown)


tfennelly:
@reviewbybees please

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/help-script.html
```diff
@@ -23,3 +23,39 @@
...
+
+A set of environment variables are made available to all Jenkins Job types, including Workflow Jobs.
+The following is a general list of variables (by name) that are available to all Job types. See the notes
+below the list for Workflow specific details.
+
+<p>
+    <iframe src="${rootURL}/env-vars.html" width="100%"></iframe>
+</p>
+
+The following variables are only available inside <code>node</code>s (i.e. not available outside):
```
jglick: In fact they are not currently available inside `node` either. As noted in JENKINS-27145:

> Setting these variables from `ExecutorStepExecution` would be welcome.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/help-script.html
```diff
@@ -23,3 +23,39 @@
...
 <pre>stage name: 'Build', concurrency: 1</pre>
+
+<h3>Environment Variables</h3>
+
+A set of environment variables are made available to all Jenkins Job types, including Workflow Jobs.
+The following is a general list of variables (by name) that are available to all Job types. See the notes
+below the list for Workflow specific details.
+
+<p>
+    <iframe src="${rootURL}/env-vars.html" width="100%"></iframe>
```
jglick: Did you test this—does `${rootURL}` actually work here? AFAIK this does not work for a help page rendered as static HTML, which is why the standard link comes from `Shell/config.groovy` in `entry.description`, rather than from a help page. The one example I know of where dynamic content is rendered in a help page is `ArtifactArchiver/help-defaultExcludes.groovy`.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/help-script.html
```diff
@@ -23,3 +23,39 @@
...
+<ul>
+    <li>EXECUTOR_NUMBER</li>
+    <li>NODE_NAME</li>
+    <li>NODE_LABELS</li>
+    <li>WORKSPACE</li>
+</ul>
+
+<h4>Using Environment Variables</h4>
+
+Environemnt variables are injected into scripts through a variable named "<strong>env</strong>". This variable,
```
jglick: typo

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/help-script.html
```diff
@@ -23,3 +23,39 @@
...
+like any other variable, can be used in the general flow of the script, or in variable substitutions e.g. when
+constructing email content when using the <code>mail</code> step:
+
+<p>
+<pre>
```
jglick: IIRC `p` and `pre` are both block-level elements, so I would suggest `</p><pre><code>`. (BTW if you switch to Jelly/Groovy you have to produce well-formed X(HT)ML.)

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/help-script.html
```diff
@@ -23,3 +23,39 @@
...
+    mail (to: 'devops@acme.com',
+        subject: "Job '${env.JOB_NAME}' (${env.BUILD_NUMBER}) is waiting for input",
+        body: "Please go to ${env.BUILD_URL}.");
+</pre>
+</p>
+
+For more on environment variables,
+<a href="https://github.com/jenkinsci/workflow-plugin/blob/master/TUTORIAL.md#managing-the-environment">see
```
jglick: Suggest `target="_blank"` (IIRC) to avoid wiping out the config screen.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/help-script.html
tfennelly: Yes I tested it and, to my surprise, it does work.

jglick:
The problem with putting content in `CpsFlowDefinition/help-script.html` is that this is not accessible to users of `CpsScmFlowDefinition`, but that problem already exists so I would not worry about it here.

jglick:
> check the variables I listed

That is your job. :-)

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition/help-script.html
jglick: Ah, [found it](https://github.com/jenkinsci/jenkins/blob/2983905e459a0311a1ca549847663999f4e5cd76/core/src/main/java/hudson/model/Descriptor.java#L831)!

tfennelly:
@jglick I updated the list of unavailable variables and it's kinda lengthy :disappointed:

Maybe we need to do something specific to this coz this looks bad imo?

jglick:
`JENKINS_URL`, `BUILD_URL`, and `JOB_URL` are definitely _supposed_ to be there. `DefaultStepContext.getEnvironment` might be at fault here. File a bug, and if you feel up to it, fix it.

jglick:
JENKINS-27145 has an analysis of what normally sets every variable I know of.

tfennelly:
> `JENKINS_URL`, `BUILD_URL`, and `JOB_URL` are definitely _supposed_ to be there. `DefaultStepContext.getEnvironment` might be at fault here. File a bug, and if you feel up to it, fix it.

Right, I thought they were too because I had an example script that I know used to work and was using `BUILD_URL`. I got the following list of vars from outputting `env.getEnvironment().keySet()` in a script and it produced:

```
[BUILD_DISPLAY_NAME, BUILD_ID, BUILD_NUMBER, BUILD_TAG, CLASSPATH, HUDSON_HOME, HUDSON_SERVER_COOKIE, JENKINS_HOME, JENKINS_SERVER_COOKIE, JOB_NAME]
```

> JENKINS-27145 has an analysis of what normally sets every variable I know of.
I did read that but would need to read it again ... kinda didn't add up for me at the time. Sorry.

jglick:
> outputting `env.getEnvironment().keySet()` in a script

Well I would cross-check against

```groovy
sh 'env'
```

which may have different results.

tfennelly:
@jglick ok, not sure how I tested `BUILD_URL` and came to the conclusion it was not there (I was sure I did an `echo` test on it and it printed `null`). Anyway, did what you said above and you are right; some of the vars I said were not present are present.

So are we good with this then?

jglick:
:+1:

jglick:
BTW technically this needed another review, but I forgive you. :-)

PR #76, opened 2015-03-05
[FIXED JENKINS-26072] Specify alternate path to lock in WorkspaceStep

jglick:
[JENKINS-26072](https://issues.jenkins-ci.org/browse/JENKINS-26072)

While this sounds similar to `AbstractProject.customWorkspace`, that only selects an alternate directory to run the build in, and does no locking whatsoever, which you could already do in a Workflow with `dir`.

@reviewbybees

### DIFF
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepExecution.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepExecution.java
@@ -23,42 +23,42 @@ import org.jenkinsci.plugins.workflow.support.actions.WorkspaceActionImpl;
 public class WorkspaceStepExecution extends AbstractStepExecutionImpl {
 
     @Inject(optional=true) private transient WorkspaceStep step;
-    @StepContextParameter private transient Computer c;
-    @StepContextParameter private transient Run<?,?> r;
+    @StepContextParameter private transient Computer computer;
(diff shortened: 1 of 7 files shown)


support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepExecution.java
```diff
@@ -37,11 +39,25 @@ public boolean start() throws Exception {
         if (n == null) {
             throw new Exception("computer does not correspond to a live node");
         }
-        FilePath p = n.getWorkspaceFor((TopLevelItem) j);
-        if (p == null) {
-            throw new IllegalStateException(n + " is offline");
+        WorkspaceList.Lease lease;
+        String dir = step.getDir();
+        if (dir == null) {
+            FilePath p = n.getWorkspaceFor((TopLevelItem) j);
```
tfennelly: Nit: variable names like `p` and `n` are not very friendly when reviewing code where the change does not show the variable decl e.g. in this case you can't see what `n` is, meaning you have to view the full source to see what it is.

tfennelly:
:+1:

jglick:
All test failures here are `ClosedByInterrupt`.

jglick:
@reviewbybees still awaiting (re-)review

stephenc:
:+1:

PR #79, opened 2015-03-06
[FIXED JENKINS-26085] Credentials support for git step

ndeloof:
[JENKINS-26085](https://issues.jenkins-ci.org/browse/JENKINS-26085)

### DIFF
--- a/scm-step/README.md
+++ b/scm-step/README.md
@@ -64,7 +64,19 @@ You may specify `poll: false` to disable polling for an SCM checkout.
 # Built-in integrations
 
 Currently there are special integrations with the Git (`git` step) and Subversion (`svn` step) plugins.
-At the moment these are very simple and take just a `url` parameter.
+
+## git
+
+the `git` step lets you define a repository to clone. For authentication, a reference to a configured credential can
+be passed.
+
+```groovy
+git url: 'git@git.local.domain:ps/project.git', credentialsId: '67c3072d-b9a7-44fa-a5aa-560ba9c1662f'
+```
+
+## svn
+
+At the moment this step is very simple and takes just a `url` parameter.
 Richer configuration may come in the future.
 
 # Generic SCM step
--- a/scm-step/src/main/java/org/jenkinsci/plugins/workflow/steps/scm/GitStep.java
+++ b/scm-step/src/main/java/org/jenkinsci/plugins/workflow/steps/scm/GitStep.java
@@ -24,17 +24,41 @@
 
 package org.jenkinsci.plugins.workflow.steps.scm;
 
+import com.cloudbees.plugins.credentials.CredentialsMatchers;
+import com.cloudbees.plugins.credentials.CredentialsProvider;
+import com.cloudbees.plugins.credentials.common.StandardCredentials;
+import com.cloudbees.plugins.credentials.common.StandardListBoxModel;
+import hudson.EnvVars;
 import hudson.Extension;
+import hudson.Util;
+import hudson.model.Item;
+import hudson.model.TaskListener;
 import hudson.plugins.git.BranchSpec;
+import hudson.plugins.git.GitException;
 import hudson.plugins.git.GitSCM;
+import hudson.plugins.git.GitTool;
+import hudson.plugins.git.Messages;
 import hudson.plugins.git.SubmoduleConfig;
 import hudson.plugins.git.UserRemoteConfig;
 import hudson.scm.SCM;
+
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+
+import hudson.security.ACL;
+import hudson.util.FormValidation;
+import hudson.util.ListBoxModel;
+import org.jenkinsci.plugins.gitclient.Git;
+import org.jenkinsci.plugins.gitclient.GitClient;
+import org.jenkinsci.plugins.gitclient.GitURIRequirementsBuilder;
+import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
+import org.kohsuke.stapler.QueryParameter;
+
+import javax.annotation.Nonnull;
 
 /**
  * Runs Git using {@link GitSCM}.
@@ -43,6 +67,7 @@ public final class GitStep extends SCMStep {
 
     private final String url;
     private String branch = "master";
+    private String credentialsId;
 
     @DataBoundConstructor public GitStep(String url) {
         this.url = url;
@@ -56,19 +81,35 @@ public final class GitStep extends SCMStep {
         return branch;
     }
 
+    public String getCredentialsId() {
+        return credentialsId;
+    }
+
     @DataBoundSetter public void setBranch(String branch) {
         this.branch = branch;
     }
 
+    @DataBoundSetter public void setCredentialsId(String credentialsId) {
+        this.credentialsId = credentialsId;
+    }
+
     @Override public SCM createSCM() {
-        return new GitSCM(createRepoList(url), Collections.singletonList(new BranchSpec("*/" + branch)), false, Collections.<SubmoduleConfig>emptyList(), null, null, null);
+        return new GitSCM(createRepoList(url, credentialsId), Collections.singletonList(new BranchSpec("*/" + branch)), false, Collections.<SubmoduleConfig>emptyList(), null, null, null);
     }
 
     // copied from GitSCM
-    static private List<UserRemoteConfig> createRepoList(String url) {
+    static private List<UserRemoteConfig> createRepoList(String url, String credentialsId) {
         List<UserRemoteConfig> repoList = new ArrayList<UserRemoteConfig>();
-        repoList.add(new UserRemoteConfig(url, null, null, null));
+        repoList.add(new UserRemoteConfig(url, null, null, credentialsId));
         return repoList;
+    }
+
+
+    private static StandardCredentials lookupCredentials(Item project, @Nonnull String credentialId, String uri) {
+        return CredentialsMatchers.firstOrNull(
+                CredentialsProvider.lookupCredentials(StandardCredentials.class, project, ACL.SYSTEM,
+                        GitURIRequirementsBuilder.fromUri(uri).build()),
+                        CredentialsMatchers.withId(credentialId));
     }
 
     @Extension(optional=true) public static final class DescriptorImpl extends SCMStepDescriptor {
@@ -76,6 +117,57 @@ public final class GitStep extends SCMStep {
         public DescriptorImpl() {
             // Fail now if dependency plugin not loaded. Descriptor.<init> will actually fail anyway, but this is just to be sure.
             GitSCM.class.hashCode();
+        }
+
+        // copy/paste from GitSCM.DescriptorImpl
+
+        public ListBoxModel doFillCredentialsIdItems(@AncestorInPath Item project,
+                                                     @QueryParameter String url) {
+            if (project == null || !project.hasPermission(Item.CONFIGURE)) {
+                return new StandardListBoxModel();
+            }
+            return new StandardListBoxModel()
+                    .withEmptySelection()
+                    .withMatching(
+                            GitClient.CREDENTIALS_MATCHER,
+                            CredentialsProvider.lookupCredentials(StandardCredentials.class,
+                                    project,
+                                    ACL.SYSTEM,
+                                    GitURIRequirementsBuilder.fromUri(url).build())
+                    );
+        }
+
+        public FormValidation doCheckUrl(@AncestorInPath Item project,
+                                         @QueryParameter String credentialsId,
+                                         @QueryParameter String value) throws IOException, InterruptedException {
+
+            if (project == null || !project.hasPermission(Item.CONFIGURE)) {
+                return FormValidation.ok();
+            }
+
+            String url = Util.fixEmptyAndTrim(value);
+            if (url == null)
+                return FormValidation.error("Please enter repository url.");
+
+            // get git executable on master
+            final EnvVars environment = new EnvVars(EnvVars.masterEnvVars);
+
+            GitClient git = Git.with(TaskListener.NULL, environment)
+                    .using(GitTool.getDefaultInstallation().getGitExe())
+                    .getClient();
+            if (credentialsId != null) {
+                StandardCredentials credentials = lookupCredentials(project, credentialsId, url);
+                if (credentials != null) git.addDefaultCredentials(credentials);
+            }
+
+            // attempt to connect the provided URL
+            try {
+                git.getHeadRev(url, "HEAD");
+            } catch (GitException e) {
+                return FormValidation.error(Messages.UserRemoteConfig_FailedToConnect(e.getMessage()));
+            }
+
+            return FormValidation.ok();
         }
 
         @Override public String getFunctionName() {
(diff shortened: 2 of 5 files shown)


ndeloof:
@reviewbybees

scm-step/README.md
```diff
@@ -64,7 +64,19 @@ You may specify `poll: false` to disable polling for an SCM checkout.
...
 Currently there are special integrations with the Git (`git` step) and Subversion (`svn` step) plugins.
-At the moment these are very simple and take just a `url` parameter.
+
+## git
+
+Git step let you define a repository to clone. For authentication, a reference to a configured credential can
+be passed.
+
+```
+git url: 'git@git.local.domain:ps/project.git', credentialsId:67c3072d-b9a7-44fa-a5aa-560ba9c1662f
```
jglick: I think you meant

```groovy
git url: 'git@git.local.domain:ps/project.git', credentialsId: '67c3072d-b9a7-44fa-a5aa-560ba9c1662f'
```

BTW use

    ```groovy

to open code blocks intended to be parts of flow scripts, so we get syntax coloring.

scm-step/README.md
```diff
@@ -64,7 +64,19 @@ You may specify `poll: false` to disable polling for an SCM checkout.
...
+```
+
+## svn
+
+At the moment these is very simple and take just a `url` parameter.
```
jglick: `this`, `takes`

scm-step/src/test/java/org/jenkinsci/plugins/workflow/steps/scm/GitStepTest.java
```diff
@@ -0,0 +1,31 @@
...
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
+ */
+public class GitStepTest {
+
+    @Rule public JenkinsRule r = new JenkinsRule();
+
+    @Test
+    public void roundtrip() throws Exception {
+        GitStep step = new GitStep("git@github.com:jenkinsci/workflow-plugin.git");
+        Step roudtrip = new StepConfigTester(r).configRoundTrip(step);
```
jglick: typo

scm-step/src/test/java/org/jenkinsci/plugins/workflow/steps/scm/GitStepTest.java
```diff
@@ -0,0 +1,31 @@
...
+        r.assertEqualDataBoundBeans(step, roudtrip);
+    }
+
+    @Test
+    public void roundtrip_withcredentials() throws Exception {
+        GitStep step = new GitStep("git@github.com:jenkinsci/workflow-plugin.git");
+        step.setCredentialsId("1234567890abcdef");
+        Step roudtrip = new StepConfigTester(r).configRoundTrip(step);
+        r.assertEqualDataBoundBeans(step, roudtrip);
```
jglick: Test results are not in yet, but I cannot imagine how this could pass! You have not edited `GitStep/config.jelly` to add an entry with `<c:select/>`, nor created `GitStep.DescriptorImpl.doFillCredentialsIdItems` to enumerate the available credentials.

And of course the test would need to actually create a credentials item of a suitable type: a dummy `credentialsId` would not appear in the pulldown and so could not be round-tripped. Note that the actual Snippet Generator does not ever round-trip anything, but `StepConfigTester` is nonetheless helpful in ensuring that all required components of form databinding are present and functioning.

scm-step/README.md
```diff
@@ -64,7 +64,19 @@ You may specify `poll: false` to disable polling for an SCM checkout.
 # Built-in integrations
 
 Currently there are special integrations with the Git (`git` step) and Subversion (`svn` step) plugins.
-At the moment these are very simple and take just a `url` parameter.
+
+## git
+
+Git step let you define a repository to clone. For authentication, a reference to a configured credential can
```
jglick: perhaps

    The `git` step…

scm-step/README.md
jglick: Since you can now pick the ID for a newly created credentials item, and the `git` step is specifically optimized for writing without snippet generation, it would make more sense to show as an example a human-readable ID.

scm-step/src/test/java/org/jenkinsci/plugins/workflow/steps/scm/GitStepTest.java
```diff
@@ -0,0 +1,31 @@
+package org.jenkinsci.plugins.workflow.steps.scm;
```
jglick: License header missing.

jglick:
Please use add-on commits. Every time you rebase, all of my comments from the earlier commit get hidden, even the ones that apparently are still valid.

ndeloof:
ok, then need to rebase before getting this merged.

jglick:
IMO there is no need to rebase ever.

jglick:
[`doFillCredentialsIdItems` example](https://github.com/jenkinsci/git-plugin/blob/9acee7ab9f7afacd3cf19a067c7c04aae4eeffe5/src/main/java/hudson/plugins/git/UserRemoteConfig.java#L77-91)

jglick:
The `c:select` example from `UserRemoteConfig/config.groovy` should probably _not_ be copied, since it is mostly a workaround for JENKINS-19124, which AFAICT was fixed in a core version predating the current baseline for `git-plugin`. (CC @stephenc, @MarkEWaite)

ndeloof:
afaik the user-friendly name of a credential is not guaranteed to be unique (I've seen duplicated on my own instance, maybe not using "latest"). So if I use it here I'm not sure about users intention.

ndeloof:
or maybe you're talking about GitStep/config.jelly, not DSL syntax ?

jglick:
> the user-friendly name of a credential is not guaranteed to be unique

I am not talking about a display name, but about `IdCredentials.id`, which in sufficiently recent versions of `credentials-plugin` (and downstream) you are allowed to pick explicitly. And yes it must be unique.

scm-step/README.md
```diff
@@ -64,7 +64,19 @@ You may specify `poll: false` to disable polling for an SCM checkout.
...
 Currently there are special integrations with the Git (`git` step) and Subversion (`svn` step) plugins.
-At the moment these are very simple and take just a `url` parameter.
+
+## git
+
+the `git` step lets you define a repository to clone. For authentication, a reference to a configured credential can
+be passed.
+
+```groovy
+git url: 'git@git.local.domain:ps/project.git', credentialsId: '67c3072d-b9a7-44fa-a5aa-560ba9c1662f'
```
jglick: (unresolved comment from rebased commit)

scm-step/src/main/java/org/jenkinsci/plugins/workflow/steps/scm/GitStep.java
```diff
@@ -78,6 +109,66 @@ public DescriptorImpl() {
...
+            String url = Util.fixEmptyAndTrim(value);
+            if (url == null)
+                return FormValidation.error("Please enter Git repository.");
+
+            if (url.indexOf('$') >= 0)
+                // set by variable, can't validate
+                return FormValidation.ok();
+
+            // get git executable on master
+            final EnvVars environment = new EnvVars(System.getenv()); // GitUtils.getPollEnvironment(project, null, launcher, TaskListener.NULL, false);
```
jglick: Use `EnvVars.masterEnvVars`.

scm-step/src/test/java/org/jenkinsci/plugins/workflow/steps/scm/GitStepTest.java
```diff
@@ -0,0 +1,55 @@
...
+        Step roundtrip = new StepConfigTester(r).configRoundTrip(step);
+        r.assertEqualDataBoundBeans(step, roundtrip);
+    }
+
+    @Test
+    public void roundtrip_withcredentials() throws Exception {
+        GitStep step = new GitStep("git@github.com:jenkinsci/workflow-plugin.git");
+        step.setCredentialsId("1234567890abcdef");
+        Step roundtrip = new StepConfigTester(r).configRoundTrip(step);
+        r.assertEqualDataBoundBeans(step, roundtrip);
```
jglick: This _should_ not pass since no such credentials exists.

scm-step/src/main/java/org/jenkinsci/plugins/workflow/steps/scm/GitStep.java
```diff
@@ -43,6 +65,7 @@
 
     private final String url;
     private String branch = "master";
+    private StandardCredentials credentials;
```
jglick: No, you cannot store a `Credentials` like this, you should only keep the `credentialsId` as you had it before (not sure why you changed it).

scm-step/src/test/java/org/jenkinsci/plugins/workflow/steps/scm/GitStepTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
+ */
+public class GitStepTest {
+
+    @Rule public JenkinsRule r = new JenkinsRule();
+
+    @Extension
+    public final static CredentialsStore store = new SystemCredentialsProvider.StoreImpl();
```
jglick: Tend to prefer the `CredentialsProvider.lookupStores(r.jenkins).iterator().next()` technique since it avoids referring to this implementation class.

BTW there is no reason for this to be an `@Extension` here, since it is normally; you could just inline it. (It does not seem to have any important state, at least for now, though it could in the future—another reason to prefer looking up the existing providers.)

ndeloof:
https://jenkins.ci.cloudbees.com/job/plugins/job/workflow-plugin/986/ fails but pass on my workstation :-\

scm-step/src/main/java/org/jenkinsci/plugins/workflow/steps/scm/GitStep.java
ndeloof: Why ? For standard component which get serialized to xml I know we can't do that, be a workflow step is serialized according to it's @DataBound attributes.

jglick:
We have been having problems with the CI builds generally, but these test failures look real.

scm-step/src/main/java/org/jenkinsci/plugins/workflow/steps/scm/GitStep.java
jglick: Treat a `Step` like any other `Describable` object.

ndeloof:
annoying if I can't reproduce :-\

jglick:
Well go back to storing `credentialsId` before you pay attention to test failures.

scm-step/src/main/java/org/jenkinsci/plugins/workflow/steps/scm/GitStep.java
```diff
@@ -56,21 +79,39 @@ public String getBranch() {
...
+        return credentialsId;
+    }
+
     @DataBoundSetter public void setBranch(String branch) {
         this.branch = branch;
     }
 
+    @DataBoundSetter public void setCredentialsId(String credentialsId) {
+        StandardCredentials credentials = lookupCredentials(null, credentialsId, url);
+        if (credentials == null) throw new IllegalArgumentException("invalid Credentials Id "+credentialsId);
```
jglick: You cannot do this check here because you do not have the `Item context` which you need to look up folder credentials. Just skip it; if the credentials are missing, this will get reported by `GitSCM` anyway.

scm-step/src/main/java/org/jenkinsci/plugins/workflow/steps/scm/GitStep.java
```diff
@@ -78,6 +119,58 @@ public DescriptorImpl() {
...
+                // set by variable, can't validate
+                return FormValidation.ok();
+
+            // get git executable on master
+            final EnvVars environment = new EnvVars(EnvVars.masterEnvVars);
+
+            GitClient git = Git.with(TaskListener.NULL, environment)
+                    .using(GitTool.getDefaultInstallation().getGitExe())
+                    .getClient();
+            git.addDefaultCredentials(lookupCredentials(project, credentialsId, url));
```
jglick: Do not pass in a null `credentialsId` and then pass null to `addDefaultCredentials`. If `credentialsId != null`, then verify that `lookupCredentials` returns non-null; otherwise skip the call to `addDefaultCredentials`.

scm-step/src/main/java/org/jenkinsci/plugins/workflow/steps/scm/GitStep.java
```diff
@@ -56,21 +79,39 @@ public String getBranch() {
...
+    static private List<UserRemoteConfig> createRepoList(String url, String credentialsId) {
         List<UserRemoteConfig> repoList = new ArrayList<UserRemoteConfig>();
-        repoList.add(new UserRemoteConfig(url, null, null, null));
+        repoList.add(new UserRemoteConfig(url, null, null, credentialsId));
         return repoList;
     }
 
+
+    private static StandardCredentials lookupCredentials(Item project, String credentialId, String uri) {
+        return (credentialId == null) ? null : CredentialsMatchers.firstOrNull(
```
jglick: As noted below, you can just make `credentialsId` here be `@Nonnull`.

scm-step/src/main/java/org/jenkinsci/plugins/workflow/steps/scm/GitStep.java
```diff
@@ -78,6 +119,58 @@ public DescriptorImpl() {
...
+                return FormValidation.ok();
+            }
+
+            String url = Util.fixEmptyAndTrim(value);
+            if (url == null)
+                return FormValidation.error("Please enter Git repository.");
+
+            if (url.indexOf('$') >= 0)
+                // set by variable, can't validate
+                return FormValidation.ok();
```
jglick: This block can be deleted for workflows, as variable substitution only makes sense for freestyle projects. (Even if the URL is computed in the flow script, the `@DataBoundConstructor` will get the evaluated result, and anyway this form validation is only for the snippet generator, where you should only be entering a _sample_ URL.)

scm-step/src/main/java/org/jenkinsci/plugins/workflow/steps/scm/GitStep.java
ndeloof: will need to backport this to git-plugin then

jglick:
:+1:

scm-step/README.md
jtnord: BTW: for documentation example.com should be used in future.

scm-step/src/main/java/org/jenkinsci/plugins/workflow/steps/scm/GitStep.java
```diff
@@ -78,6 +119,57 @@ public DescriptorImpl() {
...
+                                         @QueryParameter String credentialsId,
+                                         @QueryParameter String value) throws IOException, InterruptedException {
+
+            if (project == null || !project.hasPermission(Item.CONFIGURE)) {
+                return FormValidation.ok();
+            }
+
+            String url = Util.fixEmptyAndTrim(value);
+            if (url == null)
+                return FormValidation.error("Please enter Git repository.");
```
jtnord: nit: should be a sentence.  "Please enter the Git repository url." (or as this is already git in the context  - "Please enter repository url".

jtnord:
:+1:

PR #80, opened 2015-03-06
[JENKINS-26122] Prepend parallel step execution logs with the branch label

tfennelly:
See [JENKINS-26122](https://issues.jenkins-ci.org/browse/JENKINS-26122)

Prefixes log entries from the `parallel` step using the branch name. Added `BodyExecutionLabelAction` (currently located in the `workflow-api` module, which seems wrong ... see comment in `BodyExecutionLabelAction`) and modified `ParallelLabelAction` to extend it.

Example output - see the `[a]` and `[b]` prefixes on some of the log lines in the following sample.

```
Started by user anonymous
Running: Allocate node : Start
Running on master in /Users/tfennelly/projects/jenkins-plugins/workflow-plugin/aggregator/work/jobs/Workflwo/workspace
Running: Allocate node : Body : Start
Running: Execute sub-workflows in parallel : Start
[a] Running: Parallel branch: a
[b] Running: Parallel branch: b
[a] Running: Print Message
[a] echo a
[a] Running: Print Message
[a] echo a
[a] Running: Print Message
[a] echo a
Running: Execute sub-workflows in parallel : Body : End
[b] Running: Print Message
[b] echo b
[b] Running: Print Message
[b] echo b
[b] Running: Print Message
[b] echo b
Running: Execute sub-workflows in parallel : Body : End
Running: Execute sub-workflows in parallel : End
Running: Allocate node : Body : End
Running: Allocate node : End
Running: End of Workflow
Finished: SUCCESS
```

@reviewbybees please

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
@@ -357,6 +357,7 @@ public class ParallelStepTest extends SingleJobTestBase {
                 assertBuildCompletedSuccessfully();
 
                 // Check that the individual labeled lines are as expected
+                //System.out.println(b.getLog());
                 List<String> logLines = b.getLog(50);
                 assertGoodLabeledLogs(logLines);
 
@@ -381,7 +382,8 @@ public class ParallelStepTest extends SingleJobTestBase {
                         String.format("[%s] Running: Print Message", label),
                         String.format("[%s] echo %s", label, label)
                 );
-                assertTrue(possibleLogLines.contains(logLine));
+                boolean contains = possibleLogLines.contains(logLine);
+                assertTrue(contains);
             }
             private void assertGoodSequence(String label, List<String> logLines) {
                 String running = String.format("[%s] Running: Print Message", label);
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
@@ -279,7 +279,7 @@ public class ParallelStepTest extends SingleJobTestBase {
         for (Row row : t.getRows()) {
             ParallelLabelAction a = row.getNode().getAction(ParallelLabelAction.class);
             if (a!=null)
-                actual.add(a.getBranchName());
+                actual.add(a.getThreadName());
         }
 
         assertEquals(Arrays.asList(expected),actual);
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
@@ -9,9 +9,9 @@ import static java.util.Arrays.*;
 import java.util.List;
 import org.apache.commons.io.IOUtils;
 import org.jenkinsci.plugins.workflow.SingleJobTestBase;
+import org.jenkinsci.plugins.workflow.actions.ThreadNameAction;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.cps.steps.ParallelStep;
-import org.jenkinsci.plugins.workflow.cps.steps.ParallelStep.ParallelLabelAction;
 import org.jenkinsci.plugins.workflow.cps.steps.ParallelStepException;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepAtomNode;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
@@ -277,7 +277,7 @@ public class ParallelStepTest extends SingleJobTestBase {
         List<String> actual = new ArrayList<String>();
 
         for (Row row : t.getRows()) {
-            ParallelLabelAction a = row.getNode().getAction(ParallelLabelAction.class);
+            ThreadNameAction a = row.getNode().getAction(ThreadNameAction.class);
             if (a!=null)
                 actual.add(a.getThreadName());
         }
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
@@ -331,4 +331,68 @@ public class ParallelStepTest extends SingleJobTestBase {
             }
         });
     }
+
+    @Test
+    @Issue("JENKINS-26122")
+    public void parallelBranchLabels() {
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                p = jenkins().createProject(WorkflowJob.class, "demo");
+                p.setDefinition(new CpsFlowDefinition(join(
+                        "node {\n" +
+                                "    parallel( \n" +
+                                "        a: { \n" +
+                                "            echo('echo a');\n" +
+                                "            echo('echo a');\n" +
+                                "        }, \n" +
+                                "        b: { \n" +
+                                "            echo('echo b'); \n" +
+                                "            echo('echo b'); \n" +
+                                "        }\n" +
+                                "    )\n" +
+                                "}\n"
+                )));
+
+                startBuilding().get();
+                assertBuildCompletedSuccessfully();
+
+                // Check that the individual labeled lines are as expected
+                List<String> logLines = b.getLog(50);
+                assertGoodLabeledLogs(logLines);
+
+                // Check that the logs are printed in the right sequence e.g. that a
+                // "[a] Running: Print Message" is followed by a "[a] echo a"
+                assertGoodSequence("a", logLines);
+                assertGoodSequence("b", logLines);
+            }
+            private void assertGoodLabeledLogs(List<String> logLines) {
+                for (int i = 0; i < logLines.size(); i++) {
+                    String logLine = logLines.get(i);
+                    if (logLine.startsWith("[a] ")) {
+                        assertGoodLabeledLog("a", logLine);
+                    } else if (logLine.startsWith("[b] ")) {
+                        assertGoodLabeledLog("b", logLine);
+                    }
+                }
+            }
+            private void assertGoodLabeledLog(String label, String logLine) {
+                List<String> possibleLogLines = Arrays.asList(
+                        String.format("[%s] Running: Parallel branch: %s", label, label),
+                        String.format("[%s] Running: Print Message", label),
+                        String.format("[%s] echo %s", label, label)
+                );
+                assertTrue(possibleLogLines.contains(logLine));
+            }
+            private void assertGoodSequence(String label, List<String> logLines) {
+                String running = String.format("[%s] Running: Print Message", label);
+                String echo = String.format("[%s] echo %s", label, label);
+
+                for (int i = 0; i < logLines.size() - 1; i++) { // skip the last log line in this loop
+                    if (logLines.get(i).equals(running)) {
+                        assertEquals(echo, logLines.get(i + 1));
+                    }
+                }
+            }
+        });
+    }
 }
--- a/api/src/main/java/org/jenkinsci/plugins/workflow/actions/BodyExecutionLabelAction.java
+++ b/api/src/main/java/org/jenkinsci/plugins/workflow/actions/BodyExecutionLabelAction.java
@@ -1,0 +1,50 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2013-2014, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.jenkinsci.plugins.workflow.actions;
+
+// TODO: Seems like this should be in the step-api module alongside BodyExecution. Can't go there though because of module dependencies, which I dare not touch :)
+// I'd have thought the workflow-api module (where LabelAction lives) would be the one module that all other modules would depend on, but that's not he case.
+
+/**
+ * Body execution label.
+ * <p>
+ * Used to label the branches of an the asynchronous execution represented by a
+ * BodyExecution.
+ * </p>
+ *
+ * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
+ */
+public abstract class BodyExecutionLabelAction extends LabelAction {
+
+    private final String bodyLabel;
+
+    public BodyExecutionLabelAction(String displayName) {
+        super(displayName);
+        this.bodyLabel = displayName;
+    }
+
+    public final String getBodyLabel() {
+        return bodyLabel;
+    }
+}
--- a/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
+++ b/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
@@ -4,7 +4,7 @@ import com.cloudbees.groovy.cps.Outcome;
 import groovy.lang.Closure;
 import hudson.Extension;
 import hudson.model.TaskListener;
-import org.jenkinsci.plugins.workflow.actions.LabelAction;
+import org.jenkinsci.plugins.workflow.actions.BodyExecutionLabelAction;
 import org.jenkinsci.plugins.workflow.cps.CpsVmThreadOnly;
 import org.jenkinsci.plugins.workflow.cps.persistence.PersistIn;
 import org.jenkinsci.plugins.workflow.steps.BodyExecutionCallback;
@@ -47,11 +47,11 @@ public class ParallelStep extends Step {
     }
 
     @PersistIn(FLOW_NODE)
-    public static class ParallelLabelAction extends LabelAction {
+    public static class ParallelLabelAction extends BodyExecutionLabelAction {
         private final String branchName;
 
         ParallelLabelAction(String branchName) {
-            super(null);
+            super(branchName);
             this.branchName = branchName;
         }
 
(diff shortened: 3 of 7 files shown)


aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
```diff
@@ -331,4 +331,68 @@ public void invisibleParallelBranch() throws Exception {
...
+                                "        }\n" +
+                                "    )\n" +
+                                "}\n"
+                )));
+
+                startBuilding().get();
+                assertBuildCompletedSuccessfully();
+
+                // Check that the individual labeled lines are as expected
+                List<String> logLines = b.getLog(50);
```
jglick: Just use `JenkinsRule.getLog(b)`.

jglick: Oh, I guess you are making use of the split into separate lines.

api/src/main/java/org/jenkinsci/plugins/workflow/actions/BodyExecutionLabelAction.java
```diff
@@ -0,0 +1,50 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.jenkinsci.plugins.workflow.actions;
+
+// TODO: Seems like this should be in the step-api module alongside BodyExecution. Can't go there though because of module dependencies, which I dare not touch :)
+// I'd have thought the workflow-api module (where LabelAction lives) would be the one module that all other modules would depend on, but that's not he case.
```
jglick: `workflow-api` is the right place for this, the comment can be removed.

api/src/main/java/org/jenkinsci/plugins/workflow/actions/BodyExecutionLabelAction.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Body execution label.
+ * <p>
+ * Used to label the branches of an the asynchronous execution represented by a
+ * BodyExecution.
+ * </p>
+ *
+ * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
+ */
+public abstract class BodyExecutionLabelAction extends LabelAction {
```
jglick: Do not extend `LabelAction`, just make this an interface extending `Action`.

api/src/main/java/org/jenkinsci/plugins/workflow/actions/BodyExecutionLabelAction.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private final String bodyLabel;
```
jglick: Should be an interface, no defined storage.

jglick: (`LabelAction` was a mistake, do not follow its bad example!)

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -272,6 +292,50 @@ private void copyLogs() {
...
+            if (bodyLabelAction != null) {
+                return bodyLabelAction.getBodyLabel();
+            } else {
+                // This exec body does not have a body label. Stop now.
+                return null;
+            }
+        }
+
+        List<FlowNode> parents = node.getParents();
+        if (parents != null && !parents.isEmpty()) {
```
jglick: `getParents` may not return null, and if it is empty then the loop will be a no-op anyway, so this `if` statement could just be deleted.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -272,6 +292,50 @@ private void copyLogs() {
...
+        if (bodyInvocationAction != null) {
+            BodyExecutionLabelAction bodyLabelAction = node.getAction(BodyExecutionLabelAction.class);
+
+            if (bodyLabelAction != null) {
+                return bodyLabelAction.getBodyLabel();
+            } else {
+                // This exec body does not have a body label. Stop now.
+                return null;
+            }
+        }
```
jglick: This seems fishy. What if there is a block like `node {}` inside `parallel`? IIUC this code would fail to display any branch prefix for steps run inside that.

I do not see any particular reason to check for `BodyInvocationAction` at all.

api/src/main/java/org/jenkinsci/plugins/workflow/actions/BodyExecutionLabelAction.java
```diff
@@ -0,0 +1,50 @@
...
+package org.jenkinsci.plugins.workflow.actions;
+
+// TODO: Seems like this should be in the step-api module alongside BodyExecution. Can't go there though because of module dependencies, which I dare not touch :)
+// I'd have thought the workflow-api module (where LabelAction lives) would be the one module that all other modules would depend on, but that's not he case.
+
+/**
+ * Body execution label.
+ * <p>
+ * Used to label the branches of an the asynchronous execution represented by a
+ * BodyExecution.
```
jglick: Do we really need to be that specific? I had something simpler in mind, and more limited to the problem at hand (distinguishing parallel branches): an action which could be attached to any node (but typically a `BlockStartNode`), and which would specify a “thread name” to apply to that node and its descendants.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -243,16 +247,32 @@ private void copyLogs() {
...
+                        if (logText.isComplete()) {
+                            logText.writeRawLogTo(entry.getValue(), logger); // defend against race condition?
+                            assert !node.isRunning() : "LargeText.complete yet " + node + " claims to still be running";
+                            it.remove();
+                            modified = true;
+                        }
+                    } finally {
+                        if (prefix != null) {
+                            // It's a LogLinePrefixOutputFilter (see above). Force an EOL by closing.
+                            logger.close();
```
jglick: Would be more self-explanatory to call `LineTransformationOutputStream.forceEol` directly. I know `close` delegates to this, but you then have to dig into core sources to understand that.

jglick: (Since the normal behavior of a filtering stream in `close` is to close its delegate, which we do _not_ want here.)

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -272,6 +292,50 @@ private void copyLogs() {
...
+                if (prefix != null) {
+                    return prefix;
+                }
+            }
+        }
+
+        return null;
+    }
+
+    protected static final class LogLinePrefixOutputFilter extends LineTransformationOutputStream {
```
jglick: Can just be `private`.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -588,6 +652,21 @@ public int hashCode() {
         }
     }
 
+    private void logNodeMessage(FlowNode node, String message) {
+        PrintStream logger = listener.getLogger();
+        String prefix = getLogPrefix(node);
+        if (prefix != null) {
+            logger.println(String.format("[%s] %s", prefix, message));
```
jglick: Could be simplified to

```java
logger.printf("[%s] %s%n", prefix, message);
```

api/src/main/java/org/jenkinsci/plugins/workflow/actions/BodyExecutionLabelAction.java
jglick: IOW something like

```java
/** Applies to a {@link FlowNode} and all descendants (except those with their own such action). */
public interface ThreadNameAction extends Action {
    @Nonnull String getThreadName();
}
```

api/src/main/java/org/jenkinsci/plugins/workflow/actions/BodyExecutionLabelAction.java
tfennelly: Sorry, I thought you wanted it to extend `LabelAction`.

api/src/main/java/org/jenkinsci/plugins/workflow/actions/BodyExecutionLabelAction.java
tfennelly: ok :)

api/src/main/java/org/jenkinsci/plugins/workflow/actions/BodyExecutionLabelAction.java
jglick: (There is a possible use case for a block-scoped log prefix which is *not* a thread name, say `retry` count number, but I think it behaves differently: a given step can only be in one thread, whereas it could be nested inside several blocks of non-`parallel` steps which should probably all be reported if any are. So if and when such a feature is needed it would be a different `Action` API anyway.)

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
tfennelly: I prefer having checks like that even when, in theory, they are not needed atm (that can change if someone changes the called code). That said, I'll remove it if you feel strongly about it.

api/src/main/java/org/jenkinsci/plugins/workflow/actions/BodyExecutionLabelAction.java
tfennelly: Ok, but it's not really the thread name, is it? Or maybe it is by coincidence. Totally cool with renaming though. I do remember now that you did mention a name like that. I had just forgotten :)

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
tfennelly: Yeah maybe. Just means we need to capture it by type earlier then. I suppose casting it isn't too ugly.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
tfennelly: Sure .. I made it protected because I initially had something else in mind and thought I'd be writing tests specifically targeting it, which never happened. Thanks for spotting it.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Just add `@Nonnull` to `getParents`, at which points FindBugs would actually fail the build if you left this code in!

api/src/main/java/org/jenkinsci/plugins/workflow/actions/BodyExecutionLabelAction.java
jglick: It is not a `java.lang.Thread.name`, but we are talking about logical “threads” within the CPS VM, which should have an analogous property.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: If you do want to write tests for it—probably overkill since the hard work is done in the superclass, but who knows—then I would just give it default (“package-private”) access.

jglick:
BTW need to `git merge master`.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
tfennelly: Fair point. But it seems like I do need some way knowing when to stop searching up through the parents, otherwise I run into the situation where the `FlowEndNode` finds the `ParallelStep` as a parent (which seems very fishy to me, but I assume correct - [screenshot](https://www.dropbox.com/s/6npn2zqd3rk5w8y/Screenshot%202015-03-06%2015.20.10.png?dl=0)) and we end up thinking the end nodes are part of the parallel step, which is obviously not right.

e.g.

```
Started
Running: Allocate node : Start
Running on master in /var/folders/nw/tdwkgqsx1bj91gbwybt4d8540000gn/T/junit4918197128410133531/junit7632045101504281797/workspace/demo
Running: Allocate node : Body : Start
Running: Execute sub-workflows in parallel : Start
[a] Running: Parallel branch: a
[b] Running: Parallel branch: b
[a] Running: Print Message
[a] echo a
[a] Running: Print Message
[a] echo a
[a] Running: Execute sub-workflows in parallel : Body : End
[b] Running: Print Message
[b] echo b
[b] Running: Print Message
[b] echo b
[b] Running: Execute sub-workflows in parallel : Body : End
[a] Running: Execute sub-workflows in parallel : End
[a] Running: Allocate node : Body : End
[a] Running: Allocate node : End
[a] Running: End of Workflow
Finished: SUCCESS
```

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
```diff
@@ -47,11 +49,11 @@ public StepExecution start(StepContext context) throws Exception {
     }
 
     @PersistIn(FLOW_NODE)
-    public static class ParallelLabelAction extends LabelAction {
+    public static class ParallelLabelAction extends LabelAction implements ThreadNameAction {
         private final String branchName;
 
         ParallelLabelAction(String branchName) {
-            super(null);
+            super(branchName);
```
jglick: Revert this, passing null to the super call is intentional since we override `getDisplayName` and do not want to store two fields.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
```diff
@@ -63,6 +65,12 @@ public String getDisplayName() {
         public String getBranchName() {
             return branchName;
```
jglick: AFAIK this method is used only from `ParallelStepTest.shouldHaveParallelStepsInTheOrder`, so it could be deleted in favor of `getThreadName`.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -243,16 +247,32 @@ private void copyLogs() {
...
+                        if (logText.isComplete()) {
+                            logText.writeRawLogTo(entry.getValue(), logger); // defend against race condition?
+                            assert !node.isRunning() : "LargeText.complete yet " + node + " claims to still be running";
+                            it.remove();
+                            modified = true;
+                        }
+                    } finally {
+                        if (prefix != null) {
+                            // It's a LogLinePrefixOutputFilter (see above). Force an EOL.
+                            ((LogLinePrefixOutputFilter)logger).forceEol();
```
jglick: My advice worked: this comment is now redundant. :-)

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -272,6 +292,50 @@ private void copyLogs() {
             }
         }
     }
+
+    private String getLogPrefix(FlowNode node) {
+        BodyInvocationAction bodyInvocationAction = node.getAction(BodyInvocationAction.class);
+
+        if (bodyInvocationAction != null) {
```
jglick: As before (comment hidden because of action rename), I think the reference to `BodyInvocationAction` should be removed.

tfennelly: Fair point. But it seems like I do need some way knowing when to stop searching up through the parents, otherwise I run into the situation where the `FlowEndNode` finds the `ParallelStep` as a parent (which seems very fishy to me, but I assume correct - [screenshot](https://www.dropbox.com/s/6npn2zqd3rk5w8y/Screenshot%202015-03-06%2015.20.10.png?dl=0)) and we end up thinking the end nodes are part of the parallel step, which is obviously not right.

e.g.

```
Started
Running: Allocate node : Start
Running on master in /var/folders/nw/tdwkgqsx1bj91gbwybt4d8540000gn/T/junit4918197128410133531/junit7632045101504281797/workspace/demo
Running: Allocate node : Body : Start
Running: Execute sub-workflows in parallel : Start
[a] Running: Parallel branch: a
[b] Running: Parallel branch: b
[a] Running: Print Message
[a] echo a
[a] Running: Print Message
[a] echo a
[a] Running: Execute sub-workflows in parallel : Body : End
[b] Running: Print Message
[b] echo b
[b] Running: Print Message
[b] echo b
[b] Running: Execute sub-workflows in parallel : Body : End
[a] Running: Execute sub-workflows in parallel : End
[a] Running: Allocate node : Body : End
[a] Running: Allocate node : End
[a] Running: End of Workflow
Finished: SUCCESS
```

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
tfennelly: You're such a NIT picker ;)

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Right… I wonder if you can terminate the recursion upon encountering a `BlockEndNode`. Thus `getLogPrefix` on the join node created by the end of the `parallel` step would get `null` from the recursion on each of its parents, which is what you want.

tfennelly: OK, thanks. Need to think about that (and debug) :)

jglick: IOW try something like this:

```java
private String getLogPrefix(FlowNode node) {
    ThreadNameAction threadNameAction = node.getAction(ThreadNameAction.class);
    if (threadNameAction != null) {
        return threadNameAction.getThreadName();
    }
    if (!(node instanceof BlockEndNode)) {
        for (FlowNode parent : node.getParents()) {
            String prefix = getLogPrefix(parent);
            if (prefix != null) {
                return prefix;
            }
        }
    }
    return null;
}
```

tfennelly: @jglick Using `BlockEndNode` is better but still not quite right. See the `Running: Execute sub-workflows in parallel : Body : End` logs think they have a `ThreadNameAction` parent. Will investigate more later. Thanks Jesse.

```
Started
Running: Allocate node : Start
Running on master in /var/folders/nw/tdwkgqsx1bj91gbwybt4d8540000gn/T/junit6508583420884267681/junit4964712954863688158/workspace/demo
Running: Allocate node : Body : Start
Running: Execute sub-workflows in parallel : Start
[a] Running: Parallel branch: a
[b] Running: Parallel branch: b
[a] Running: Print Message
[a] echo a
[a] Running: Print Message
[a] echo a
[a] Running: Execute sub-workflows in parallel : Body : End
[b] Running: Print Message
[b] echo b
[b] Running: Print Message
[b] echo b
[b] Running: Execute sub-workflows in parallel : Body : End
Running: Execute sub-workflows in parallel : End
Running: Allocate node : Body : End
Running: Allocate node : End
Running: End of Workflow
Finished: SUCCESS
```

jglick: So that is pretty close at least, should be able to use `/flowGraph` to debug that.

tfennelly:
@jglick ok I think that's fixed now using `BlockEndNode` - I was just using it incorrectly.

Also removed `ParallelLabelAction.getBranchName()` as requested. Hopefully tests pass and we can give it a thumbs up ;)

Back to Basketball match now :)

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
```diff
@@ -279,7 +279,7 @@ private void shouldHaveParallelStepsInTheOrder(String... expected) {
         for (Row row : t.getRows()) {
             ParallelLabelAction a = row.getNode().getAction(ParallelLabelAction.class);
```
jglick: Actually this could now just be `ThreadNameAction`, permitting `ParallelLabelAction` to be made private.

jglick:
:+1:

PR #87, opened 2015-03-13
Work around eclipse compiler.

jtnord:
The eclipse compiler is brain dead and treats
rule.configRoundtrip(FreeStyleProject) as ambiguous.

@reviewbybees

### DIFF
--- a/step-api/src/test/java/org/jenkinsci/plugins/workflow/steps/StepConfigTester.java
+++ b/step-api/src/test/java/org/jenkinsci/plugins/workflow/steps/StepConfigTester.java
@@ -25,6 +25,7 @@
 package org.jenkinsci.plugins.workflow.steps;
 
 import hudson.Extension;
+import hudson.model.Item;
 import hudson.model.AbstractProject;
 import hudson.model.FreeStyleProject;
 import hudson.tasks.BuildStepDescriptor;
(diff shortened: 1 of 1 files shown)


step-api/src/test/java/org/jenkinsci/plugins/workflow/steps/StepConfigTester.java
```diff
@@ -30,6 +30,7 @@
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.Builder;
 import static org.junit.Assert.*;
+
```
stephenc: @jtnord if you are going to be nit picky on people using `l` as a variable identifier and the perfectly cromulent word iff (which can only embiggen your soul) *and* you insist on using a brain-dead IDE, please do not submit pull requests with gratuitous whitespace changes ;-)

step-api/src/test/java/org/jenkinsci/plugins/workflow/steps/StepConfigTester.java
```diff
@@ -57,7 +58,8 @@ public StepConfigTester(JenkinsRule rule) {
         FreeStyleProject p = rule.createFreeStyleProject();
         try {
             p.getBuildersList().add(new StepBuilder(before));
-            p = rule.configRoundtrip(p);
+            rule.submit(rule.createWebClient().getPage(p,"configure").getFormByName("config"));
+            //p = rule.configRoundtrip((Project<FreeStyleProject,FreeStyleBuild>)p);
```
jglick: Try creating a method in _this_ class like

```java
private static <P extends Job> P configRoundtrip(JenkinsRule rule, P job) throws Exception {
    rule.configRoundtrip(job);
}
```

and then change the above code to

```java
p = configRoundtrip(rule, p);
```

jglick:
:+1:

kohsuke:
:+1:

PR #85, opened 2015-03-12
Random WorkflowTest.executorStepRestart failure on CI

jglick:
[This](https://jenkins.ci.cloudbees.com/job/plugins/job/workflow-plugin/org.jenkins-ci.plugins.workflow$workflow-aggregator/1025/testReport/junit/org.jenkinsci.plugins.workflow/WorkflowTest/executorStepRestart/) for example. @reviewbybees

PR #88, opened 2015-03-13
[JENKINS-26034] add option to fail fast on parallel branch failure.

jtnord:
Added an option "failFast: true" to the parallel step options that
will terminate all running branches inside a parallel step if one of the
branches fails.

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
@@ -81,7 +81,7 @@ public class ParallelStepTest extends SingleJobTestBase {
                     "      b: { error 'died' },",
 
                         // make sure this branch takes longer than a
-                    "      a: { sleep 3; writeFile text: '', file: 'b.done' }",
+                    "      a: { sleep 3; writeFile text: '', file: 'a.done' }",
                     "    )",
                     "    assert false;",
                     "  } catch (ParallelStepException e) {",
@@ -93,7 +93,7 @@ public class ParallelStepTest extends SingleJobTestBase {
 
                 startBuilding().get();
                 assertBuildCompletedSuccessfully();
-                assert jenkins().getWorkspaceFor(p).child("b.done").exists();
+                assert jenkins().getWorkspaceFor(p).child("a.done").exists();
 
                 buildTable();
                 shouldHaveParallelStepsInTheOrder("b","a");
@@ -120,11 +120,12 @@ public class ParallelStepTest extends SingleJobTestBase {
                     "      b: { error 'died' },",
 
                         // make sure this branch takes longer than a
-                    "      a: { sleep 10; writeFile text: '', file: 'b.done' },",
+                    "      a: { sleep 25; writeFile text: '', file: 'a.done' },",
                     "      failFast: true",
                     "    )",
                     "    assert false",
                     "  } catch (ParallelStepException e) {",
+                    "    echo e.toString()",
                     "    assert e.name=='b'",
                     "    assert e.cause instanceof AbortException",
                     "  }",
@@ -133,7 +134,7 @@ public class ParallelStepTest extends SingleJobTestBase {
 
                 startBuilding().get();
                 assertBuildCompletedSuccessfully();
-                Assert.assertFalse("b should have aborted", jenkins().getWorkspaceFor(p).child("b.done").exists());
+                Assert.assertFalse("a should have aborted", jenkins().getWorkspaceFor(p).child("a.done").exists());
 
             }
         });
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
@@ -123,7 +123,7 @@ public class ParallelStepTest extends SingleJobTestBase {
                     "      a: { sleep 10; writeFile text: '', file: 'b.done' },",
                     "      failFast: true",
                     "    )",
-                    "    assert false;",
+                    "    assert false",
                     "  } catch (ParallelStepException e) {",
                     "    assert e.name=='b'",
                     "    assert e.cause instanceof AbortException",
@@ -135,8 +135,36 @@ public class ParallelStepTest extends SingleJobTestBase {
                 assertBuildCompletedSuccessfully();
                 Assert.assertFalse("b should have aborted", jenkins().getWorkspaceFor(p).child("b.done").exists());
 
-                buildTable();
-                shouldHaveParallelStepsInTheOrder("b","a");
+            }
+        });
+    }
+
+    /**
+     * FailFast should not kill branches if there is no failure.
+     */
+    @Test @Issue("JENKINS-26034")
+    public void failFast_has_no_effect_on_suceess() throws Exception {
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                p = jenkins().createProject(WorkflowJob.class, "demo");
+                p.setDefinition(new CpsFlowDefinition(join(
+                    "import "+AbortException.class.getName(),
+                    "import "+ParallelStepException.class.getName(),
+
+                    "node {",
+                    "    parallel(",
+                    "      a: { echo 'hello from a';sleep 1;echo 'goodbye from a' },",
+                    "      b: { echo 'hello from b';sleep 1;echo 'goodbye from b' },",
+                    "      c: { echo 'hello from c';sleep 1;echo 'goodbye from c' },",
+                    // make sure this branch is quicker than the others.
+                    "      d: { echo 'hello from d' },",
+                    "      failFast: true",
+                    "    )",
+                    "}"
+                )));
+
+                startBuilding().get();
+                assertBuildCompletedSuccessfully();
             }
         });
     }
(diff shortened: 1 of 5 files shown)


jglick:
Add an entry to `CHANGES.md`.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
```diff
@@ -100,6 +101,47 @@ public void failure_in_subflow_will_cause_join_to_fail() throws Exception {
...
+                    "  }",
+                    "}"
+                )));
+
+                startBuilding().get();
+                assertBuildCompletedSuccessfully();
+                Assert.assertFalse("b should have aborted", jenkins().getWorkspaceFor(p).child("b.done").exists());
+
+                buildTable();
+                shouldHaveParallelStepsInTheOrder("b","a");
```
jglick: This assertion is not particularly interesting here, could probably be omitted.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
```diff
@@ -124,15 +161,29 @@ public String getFunctionName() {
...
         @Override
         public Step newInstance(Map<String,Object> arguments) {
+            boolean earlyTermination = false;
+            Map<String,Closure<?>> closures = new LinkedHashMap<String, Closure<?>>();
             for (Entry<String,Object> e : arguments.entrySet()) {
-                if (!(e.getValue() instanceof Closure))
+                if ((e.getValue() instanceof Closure)) {
+                    closures.put(e.getKey(), (Closure<?>)e.getValue());
+                }
+                else if ("earlyTermination".equals(e.getKey())) {
```
jglick: Needs to be updated. (Preferably the variable name, too.)

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
```diff
@@ -124,15 +161,29 @@ public String getFunctionName() {
...
-            return new ParallelStep((Map)arguments);
+            return new ParallelStep((Map)closures, earlyTermination);
         }
 
         @Override public Map<String,Object> defineArguments(Step step) throws UnsupportedOperationException {
-            return new TreeMap<String,Object>(((ParallelStep) step).closures);
+            ParallelStep ps = (ParallelStep) step;
+            Map<String,Object> retVal = new TreeMap<String,Object>(ps.closures);
+            if (ps.failFast) {
+                retVal.put("earlyTermination", Boolean.TRUE);
```
jglick: Ditto here.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep/config.jelly
```diff
@@ -27,12 +27,14 @@ THE SOFTWARE.
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
     <f:block>
         <p>
-            (No visual configuration.) Takes a map from branch names to closures:
+            (No visual configuration.) Takes a map from branch names to closures and an optional argument <code>failFast</code>
```
jglick: `failFast: true`

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
```diff
@@ -91,8 +117,19 @@ private void checkAllDone() {
                 for (Entry<String,Outcome> e : handler.outcomes.entrySet()) {
                     Outcome o = e.getValue();
 
-                    if (o==null)
-                        return; // some of the results are not yet ready
+                    if (o==null) {
+                        // some of the results are not yet ready
+                        if (handler.failFast && ! handler.isStopSent()) {
```
jglick: `&& outcome.isFailure()`

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
```diff
@@ -45,17 +51,29 @@ public StepExecution start(StepContext context) throws Exception {
...
 
+    /*package*/ boolean isFailFast() {
+        return failFast;
+    }
+
+
     @PersistIn(PROGRAM)
     static class ResultHandler implements Serializable {
         private final StepContext context;
+        private final ParallelStepExecution stepExecution;
```
jglick: Unclear to me if storing a reference to the `ParallelStepExecution` like this is kosher. @kohsuke? I was already confused by why `ParallelStepExecution` is storing its `BodyExecution`s.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
```diff
@@ -124,15 +161,29 @@ public String getFunctionName() {
...
         public Step newInstance(Map<String,Object> arguments) {
+            boolean earlyTermination = false;
+            Map<String,Closure<?>> closures = new LinkedHashMap<String, Closure<?>>();
             for (Entry<String,Object> e : arguments.entrySet()) {
-                if (!(e.getValue() instanceof Closure))
+                if ((e.getValue() instanceof Closure)) {
+                    closures.put(e.getKey(), (Closure<?>)e.getValue());
+                }
+                else if ("earlyTermination".equals(e.getKey())) {
+                    earlyTermination = Boolean.valueOf(e.getValue().toString());
```
jglick: `failFast = (Boolean) e.getValue()` would suffice, I think.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
jglick: This should rather be moved down into the `if (o.isFailure())` block.

jglick: (not quite, but similar)

jtnord:
@reviewbybees - feeling confident....

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
kohsuke: `StepExecution` is persisted in program state, so this is OK. `BodyExecution`s are stored to allow cancellation of parallel branches.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
```diff
@@ -76,47 +107,70 @@ Callback callbackFor(String name) {
...
-
+            
             private static final long serialVersionUID = 1L;
         }
 
         private static final long serialVersionUID = 1L;
     }
 
+    /** Internal exception that is only used internally to abort a parallel body in the case of a failFast body failing. */
+    private static final class FailFastException extends Exception {
```
kohsuke: Users are going to see this exception, no? Don't we need to explain this like users will read it?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
```diff
@@ -76,47 +107,70 @@ Callback callbackFor(String name) {
...
                     Outcome o = e.getValue();
 
-                    if (o==null)
-                        return; // some of the results are not yet ready
+                    if (o==null) {
+                        // some of the results are not yet ready
+                        if (stepFailed && handler.failFast && ! handler.isStopSent()) {
+                            handler.stopSent();
+                            try {
+                                handler.stepExecution.stop(new FailFastException());
```
kohsuke: I'd also put the exception message and chain the original failure, since the exception can be visible.

It's bit like `InterruptedException`, that you'd often want to know who interrupted it.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
```diff
@@ -100,6 +101,76 @@ public void failure_in_subflow_will_cause_join_to_fail() throws Exception {
...
+
+            }
+        });
+    }
+
+    /**
+     * FailFast should not kill branches if there is no failure.
+     */
+    @Test @Issue("JENKINS-26034")
+    public void failFast_has_no_effect_on_suceess() throws Exception {
```
jglick: typo

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/ParallelStep.java
jtnord: No - the user sees a `FlowInterruptedException`

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/parallel/ParallelStepTest.java
```diff
@@ -80,7 +81,7 @@ public void failure_in_subflow_will_cause_join_to_fail() throws Exception {
                     "      b: { error 'died' },",
 
                         // make sure this branch takes longer than a
```
jglick: BTW the comment looks wrong too.

jglick:
:+1:

kohsuke:
:+1: 

Don't let the perfect get in the way of better

jglick:
The test failure is covered by #85.

PR #90, opened 2015-03-17
Parallel build cancel

jglick:
Added a test for a useful scenario involving orchestration of downstream builds. Note that before this change, commenting out the call to `Executor.interrupt` in `BuildTriggerStepExecution.stop` (dating to work @kohsuke did in July) did not cause test failures; now it does.

@reviewbybees


tfennelly:
:+1:

alecharp:
:+1:

why removing the `@author` on first commit?

rsandell:
:+1:

jglick:
> why removing the `@author` on first commit?

Because looking at the file I realized that at least half of what it now consists of was _not_ written by Vivek. That is why I hate `@author` tags. :-) If you actually care who did what (and when), look at Git history.

PR #91, opened 2015-03-18
[JENKINS-27474] Adding build step File Exists

nexus49:
This step verifies if a file exists in the workspace.

https://issues.jenkins-ci.org/browse/JENKINS-27474

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ReadWriteFileStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ReadWriteFileStepTest.java
@@ -25,19 +25,24 @@
 package org.jenkinsci.plugins.workflow.steps;
 
 import hudson.Functions;
+
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.junit.Rule;
 import org.junit.Test;
 import org.jvnet.hudson.test.JenkinsRule;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ReadWriteFileStepTest {
+
+	Logger LOG = LoggerFactory.getLogger(ReadWriteFileStepTest.class);
 
     @Rule public JenkinsRule r = new JenkinsRule();
 
     @Test public void basics() throws Exception {
-        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
-        boolean win = Functions.isWindows();
+        final WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        final boolean win = Functions.isWindows();
         p.setDefinition(new CpsFlowDefinition(
                 "node {\n" +
                 (win ? "  bat 'echo hello > f1'\n" : "  sh 'echo hello > f1'\n") +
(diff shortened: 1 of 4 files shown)


basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/FileExistsStep.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2014 Jesse Glick.
```
jglick: Probably you want to add your own copyright. :-)

jglick:
Would need a test. Probably could just be added to `ReadWriteFileStepTest` as a new test case.

jglick:
You forgot to make a `config.jelly`. (`StepConfigTester` can be used to test the correctness of the config form, though I generally do not bother with it for steps with such trivial configuration.)

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ReadWriteFileStepTest.java
```diff
@@ -36,8 +37,8 @@
     @Rule public JenkinsRule r = new JenkinsRule();
 
     @Test public void basics() throws Exception {
-        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
-        boolean win = Functions.isWindows();
+        final WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        final boolean win = Functions.isWindows();
```
jglick: Gratuitous, could be reverted.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ReadWriteFileStepTest.java
```diff
@@ -49,4 +50,33 @@
...
+	{
+        final WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition(
+                "node {\n" +
+                "  def exists = fileExists('test.txt')  \n" +
+                "  echo \"FileExists: ${exists}\" \n" +
+                "}"));
+
+		r.assertLogContains("FileExists: false", p.scheduleBuild2(0).get());
+		r.assertBuildStatusSuccess(p.scheduleBuild2(0));
```
jglick: This is building the job twice. Probably you meant something like

```java
r.assertLogContains("FileExists: false", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
```

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ReadWriteFileStepTest.java
```diff
@@ -49,4 +50,33 @@
...
+		final WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+		p.setDefinition(new CpsFlowDefinition(
+				"node {\n" +
+				"  writeFile file: 'test2.txt', text:'content of file' \n" +
+				"  def exists = fileExists('test2.txt')  \n" +
+				"  echo \"FileExists: ${exists}\" \n" +
+				"}"));
+
+		r.assertLogContains("FileExists: true", p.scheduleBuild2(0).get());
+		r.assertBuildStatusSuccess(p.scheduleBuild2(0));
```
jglick: (ditto here)

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ReadWriteFileStepTest.java
```diff
@@ -49,4 +50,33 @@
...
+
+	@Test
+	public void shouldTestFileExistsAndFindFile() throws Exception
+	{
+		final WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+		p.setDefinition(new CpsFlowDefinition(
+				"node {\n" +
+				"  writeFile file: 'test2.txt', text:'content of file' \n" +
+				"  def exists = fileExists('test2.txt')  \n" +
+				"  echo \"FileExists: ${exists}\" \n" +
```
jglick: Could save a few seconds of test time and shorten sources by merging these two test cases into one: `fileExists`, `echo`, `writeFile`, `fileExists`, `echo` should print false then true.

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/FileExistsStep.java
```diff
@@ -0,0 +1,75 @@
...
+
+    }
+
+    public static final class Execution extends AbstractSynchronousStepExecution<Boolean> {
+
+        @Inject private transient FileExistsStep step;
+        @StepContextParameter private transient FilePath workspace;
+
+        @Override protected Boolean run() throws Exception {
+        	return Boolean.valueOf(workspace.child(step.file).exists());
```
jglick: Can skip `Boolean.valueOf` and rely on autoboxing.

jglick:
I would also merge with `master` to pick up some test fixes.

nexus49:
Thanks Jesse for you feedback. Its my first contribution so i still figure some things out... Let me know if you have further feedback for me to incorporate.

PR #89, opened 2015-03-17
Still trying to diagnose failures on WorkflowTest.executorStepRestart

jglick:
After #85 was merged, I now see a [different failure](https://jenkins.ci.cloudbees.com/job/plugins/job/workflow-plugin/org.jenkins-ci.plugins.workflow$workflow-aggregator/1049/testReport/junit/org.jenkinsci.plugins.workflow/WorkflowTest/executorStepRestart/). Again `isBuilding` remains true, and as before we see

```
Started
Running: Allocate node : Start
Still waiting to schedule task
Waiting for next available executor
Resuming build
Resuming build
```

for a while. The duplicated message looks suspicious (hence this patch). Then we get

```
… org.jenkinsci.plugins.workflow.job.WorkflowRun finish
INFO: demo #1 completed: SUCCESS
```

and then we start seeing

```
Started
Running: Allocate node : Start
Still waiting to schedule task
Waiting for next available executor
Resuming build
Resuming build
Running on slave0 in /tmp/hudson4729640265695080163test/workspace/demo
Running: Allocate node : Body : Start
Running: Print Message
OK ran
Running: Allocate node : Body : End
Running: Allocate node : End
Running: End of Workflow
Finished: SUCCESS
```

for a while, though the build is still considered to be `building`, and then after a bit it settles into

```
Started
Running: Allocate node : Start
Still waiting to schedule task
Waiting for next available executor
Resuming build
Resuming build
Running on slave0 in /tmp/hudson4729640265695080163test/workspace/demo
Running: Allocate node : Body : Start
Running: Print Message
OK ran
Running: Allocate node : Body : End
Running: Allocate node : End
Running: End of Workflow
Finished: SUCCESS
Running on slave0 in /tmp/hudson4729640265695080163test/workspace/demo
```

which makes even less sense.

@reviewbybees

PR #92, opened 2015-03-18
Test improvements

jglick:
Changes suggested in [JENKINS-25975](https://issues.jenkins-ci.org/browse/JENKINS-25975) and related issues.

@reviewbybees

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildWatcher.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildWatcher.java
@@ -85,6 +85,9 @@ public final class BuildWatcher extends ExternalResource {
     @Extension public static final class Listener extends RunListener<Run<?,?>> {
 
         @Override public void onStarted(Run<?,?> r, TaskListener listener) {
+            if (!active) {
+                return;
+            }
             RunningBuild build = new RunningBuild(r);
             RunningBuild orig = builds.put(r.getLogFile(), build);
             if (orig != null) {
@@ -93,11 +96,12 @@ public final class BuildWatcher extends ExternalResource {
         }
 
         @Override public void onFinalized(Run<?,?> r) {
+            if (!active) {
+                return;
+            }
             RunningBuild build = builds.remove(r.getLogFile());
             if (build != null) {
-                if (active) {
-                    build.copy();
-                }
+                build.copy();
             } else {
                 System.err.println(r + " was finalized but not started?!");
             }
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildWatcher.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildWatcher.java
@@ -1,0 +1,153 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 Jesse Glick.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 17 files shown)


jglick:
The test failure is in `WorkflowTest.executorStepRestart`, which has been failing off and on for a while now; apparently #89 did not help.

recampbell:
:+1: Seems legit

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildWatcher.java
```diff
@@ -0,0 +1,153 @@
...
+    @Override protected void after() {
+        active = false;
+        thread.interrupt();
+    }
+
+    @Extension public static final class Listener extends RunListener<Run<?,?>> {
+
+        @Override public void onStarted(Run<?,?> r, TaskListener listener) {
+            RunningBuild build = new RunningBuild(r);
+            RunningBuild orig = builds.put(r.getLogFile(), build);
```
tfennelly: Why use the log file as the key (Vs say the `Run` instance or its externalisable Id) ?

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildWatcher.java
```diff
@@ -0,0 +1,153 @@
...
+            if (orig != null) {
+                System.err.println(r + " was started twice?!");
+            }
+        }
+
+        @Override public void onFinalized(Run<?,?> r) {
+            RunningBuild build = builds.remove(r.getLogFile());
+            if (build != null) {
+                if (active) {
```
tfennelly: `active` is meant to be a static, right?

tfennelly: Ok sorry... I see what this is doing.

jglick:
Confirmed my hypothesis:

```
… jenkins.InitReactorRunner$1 onAttained
INFO: Completed initialization
loading demo #1 @665c98da
loading demo #1 @a6756a5
[demo #1] Resuming build
[demo #1] Resuming build
```

I think this is [JENKINS-22767](https://issues.jenkins-ci.org/browse/JENKINS-22767).

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildWatcher.java
jglick: I did not want to use the `Run` instance because this needs to work across Jenkins restarts. Nor did I want to use the `externalizableId` because I wanted this to work as a `ClassRule` across different test cases within the same suite, which often reuse the same project name and of course build number (`p #1`) for example. Whereas the `logFile` includes the randomized `$JENKINS_HOME` prefix and so should be both consistent across restart and unique across tests.

tfennelly:
:+1:

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildWatcher.java
jglick: This is a general `@Extension`, so active even in tests that are not asking for me. In that case, `!active` and so the map is updated but nothing else happens. (Well, the log file is still opened, but nothing extra is printed.) I suppose it would be preferable to return immediately from all listener methods if `!active`; this is set pretty early as a `@ClassRule` so there should not be a race condition.

jglick:
@reviewbybees after more recent commits

jglick:
(I expect to file a separate PR copying the new utilities to `jenkinsci/jenkins/test/src/main/java/`.)

tfennelly:
:+1:

PR #93, opened 2015-03-19
CpsScript name clash

jglick:
A customer reported a build compilation failure after upgrading to 1.4, due to using a local method named `build`. The test here in fact fails in the expected way without the source patch:

```
org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
WorkflowScript: 1: The return type of java.lang.Object build() in WorkflowScript is incompatible with hudson.model.Run build() in org.jenkinsci.plugins.workflow.cps.CpsScript
. At [1:1]  @ line 1, column 1.
   def build() {20}; …
   ^

1 error
```

@reviewbybees (ref. ZD-24830)


jglick:
Note that you need to run `diff` with a low similarity index, such as `git show -M2 f46415b10badf290400f9afc6419f66dc54092d3`, to see the rename sensibly. If GitHub offers a way to pass `-M` to the diff view, I could not find it: tried `files?M=2` by analogy with the semisecret `?w=1` without success.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -319,7 +319,7 @@ public File getStorageDir() throws IOException {
     public void start() throws IOException {
         final CpsScript s = parseScript();
         scriptClass = s.getClass();
-        s.initialize();
+        s.$initialize();
```
tfennelly: I always assumed `$` would have been an invalid char in a java identifier :smile:

tfennelly:
:+1:

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: Nope, it is legal. Generally use of it is discouraged in non-generated code. This is sort of a special case. I considered also `__initialize__` etc.

PR #94, opened 2015-03-19
Update baseline to 1.596.x

jglick:
Various new APIs can be used, and workarounds removed, once we move to the next LTS baseline. Avoids the need to keep this kind of changes in feature PRs like #37.

@reviewbybees


tfennelly:
:+1:

PR #95, opened 2015-03-19
[WiP] Using API improvements around Executor/Executable

jglick:
Based on https://github.com/jenkinsci/jenkins/pull/1610. Subsumes #94 and #96.

@reviewbybees (WiP until 1.607 is in an LTS line)


oleg-nenashev:
Should be squashed into a single commit before integrating in #120

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -194,39 +195,53 @@ public WorkflowRun(WorkflowJob job, File dir) throws IOException {
...
             executionPromise.setException(x);
+            return;
         }
+        throw sleep();
     }
 
-    /**
-     * Sleeps until the run is finished, updating log messages periodically.
-     */
-    void waitForCompletion() {
```
oleg-nenashev: Is it safe? the method was not private

jglick:
I do not rebase. `git-merge` works fine.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Package-private, and anyway the `workflow-job` plugin was not intended to expose any APIs.

support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution/PlaceholderTask/PlaceholderExecutable/executorCell.jelly
```diff
@@ -26,10 +26,23 @@
 
 <j:jelly xmlns:j="jelly:core" xmlns:t="/lib/hudson" xmlns:l="/lib/layout">
     <td class="pane">
-        <div style="white-space: normal">${%Building}
-            <!-- TODO check for Item.READ permission on the accessControlled() -->
-            <a href="${rootURL}/${it.parent.url}" class="model-link"><l:breakable value="${it.parent.fullDisplayName}"/></a>
-            <t:buildProgressBar build="${it.parent}" executor="${executor}"/>
+        <div style="white-space: normal">
+            <j:set var="r" value="${it.parent.run()}"/>
+            <j:set var="rAuth" value="${r != null and h.hasPermission(r.parent, r.parent.READ) ? r : null}"/>
```
oleg-nenashev: <code>Item.DISCOVER</code> may be missing in the logic (e.g. "I can see the project, but cannot open its internals"). On the other hand, this executor cell will leak some info in such case

oleg-nenashev:
LGTM

support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution/PlaceholderTask/PlaceholderExecutable/executorCell.jelly
jglick: At any rate, this logic is basically copied from freestyle projects.

jglick:
Merged via #120 into `lts-609`.

PR #97, opened 2015-03-20
[JENKINS-27531] Run.id-related error in 1.597+

jglick:
[JENKINS-27531](https://issues.jenkins-ci.org/browse/JENKINS-27531)

Without the source patch, but with the POM baseline set to 1.597, the test fails with the expected error.

@reviewbybees


jglick:
The alternative approach is to kill `LOADING_RUNS` by making sure no object deserialized as part of a `WorkflowRun` needs to load that same run by ID/number. I am not sure offhand how much work that is, other than that I found it difficult at the time `LOADING_RUNS` was introduced. The stack trace in JENKINS-27531 demonstrates one such case that would have to be addressed: `CpsFlowExecution.createStorage` would need to pass a handle from which the actual storage directory would be loaded on demand (possibly needed at some point to handle renames/moves of running flows, which probably does not work today).

tfennelly:
:+1:

jtnord:
:+1:

PR #98, opened 2015-03-24
Updating Dockerfile

jglick:
@reviewbybees

### DIFF
--- a/demo/Dockerfile
+++ b/demo/Dockerfile
@@ -1,4 +1,4 @@
-FROM dockerfile/java
+FROM java:7
 RUN curl -o /tmp/maven.tgz http://mirror.cogentco.com/pub/apache/maven/maven-3/3.2.2/binaries/apache-maven-3.2.2-bin.tar.gz
 RUN cd /usr/local; tar xvzf /tmp/maven.tgz
 RUN mkdir -p /var/lib/jenkins/plugins /var/lib/jenkins/jobs/cd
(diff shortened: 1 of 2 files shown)


demo/Dockerfile
```diff
@@ -1,29 +1,40 @@
-FROM dockerfile/java
```
jglick: This one got deprecated.

demo/Dockerfile
```diff
@@ -1,29 +1,40 @@
...
+RUN wget -O - "http://mirrors.ibiblio.org/eclipse/jetty/$JETTY_VERSION/dist/jetty-distribution-$JETTY_VERSION.tar.gz" | tar xvfz -
+RUN ln -sv jetty-distribution-$JETTY_VERSION jetty
 RUN cd /tmp; ln -s /opt/jetty/webapps
 
-# install OpenSSH and allow root to login with password 'root'
-RUN apt-get install -y openssh-server
-RUN echo 'root:root' | chpasswd
+# https://github.com/carlossg/docker-maven/issues/3
+RUN apt-get install -y ca-certificates-java && update-ca-certificates -f
+RUN apt-get install -y openjdk-8-jdk
```
jglick: Yes a bunch of stuff is in illogical places, to avoid blowing out the Docker cache rerunning everything from the start: some of these installations are quite slow.

jglick:
A future refactoring should move everything not related to `run-snapshot` out of the `Makefile` so that the `Dockerfile` and static content files beneath it can be self-contained (suitable for Docker automated builds, for example). Among other things that means fixing versions of Jenkins core and plugins in source control.

stephenc:
:+1: from the blind man on the galloping horse... looks ok but I cant be sure

tfennelly:
:+1: though it's kinda double-dutch to me

PR #99, opened 2015-03-24
[JENKINS-26085] Handle empty credentialsId

jglick:
Fixes mistake made in #79 which caused the snippet generator to offer

```groovy
git credentialsId: '', url: '…'
```

when no credentials were selected. It should be simply

```groovy
git '…'
```

@reviewbybees


tfennelly:
:+1:

PR #100, opened 2015-03-25
Corrected help text for catchError

jglick:
Correcting help text added in 5ade0f0d7e05e91e2db7a02c1486010c2db325e6. @reviewbybees and FYI @kohsuke

### DIFF
--- a/basic-steps/src/main/resources/org/jenkinsci/plugins/workflow/steps/CatchErrorStep/help.html
+++ b/basic-steps/src/main/resources/org/jenkinsci/plugins/workflow/steps/CatchErrorStep/help.html
@@ -1,32 +1,7 @@
 <div>
-    <p>
     If the body throws an exception, mark the build as a failure, but nonetheless
-    continue to execute the workflow from the statement next to the 'catchError' step.
-    This is useful if you want to find as many errors as you can before terminating
-    the workflow. See the following example:
-
-<pre>
-catchError {
-    sh 'false'  // this step will fail
-    
-    echo 'this step will not run'
-}
-echo 'but this step will run after sh fails'
-...
-// and the workflow fails at this point after normally completing the program
-</pre>
-    
-    <p>
-    If you want to ignore a failure, use the try/catch block instead:
-
-<pre>
-try {
-    sh 'false'  // this step will fail
-
-    echo 'this step will not run'
-} catch(_) {}
-echo 'but this step will run after sh fails'
-...
-// and the workflow as a whole would succeed
-</pre>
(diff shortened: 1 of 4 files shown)


tfennelly:
:+1:

kohsuke:
Thanks for revising the text, but I strongly suggest you keep some form of code examples, based on the experience that drove me to write this.

You can see [the entire transcript here](https://gist.github.com/kohsuke/4b1a75d0c815403cd033), but at some point he was confused enough to try a script like the following:

```
catchError {
  try {
    parallel(
      dummy1: { build job: 'dummy1' },
      dummy2: { build job: 'dummy2' }, // fails
    )
  } finally {
  }
}

echo 'The End'
```

As I kept saying, Jenkins Workflow needs to be approachable to somebody who is not very familiar with Groovy. The `catchError` step has such an obvious and visible name that people naturally gravitate to it, and even I didn't know until now that this does what it does. I thought it was just a sugar of `try { ...} catch (e) { /* ignore */ }`.

jglick:
Changing the step name now is not an option, as #12 was merged in August. (Now that you can run `currentBuild.status = 'FAILURE'` the step is probably superfluous, but at the time it was the only way to make `Mailer` work.) The descriptor’s `displayName` could be adjusted to make it seem less desirable somehow.

The link in the help text does include a code sample. I guess you are suggesting moving that sample here are making the current document refer to this help text?

jglick:
Will work on just deprecating the step.

basic-steps/src/main/resources/org/jenkinsci/plugins/workflow/steps/CatchErrorStep/help.html
```diff
@@ -1,32 +1,31 @@
...
+    } catch (err) {
+        echo "Caught: ${err}"
+        currentBuild.result = 'FAILURE'
+    }
+    step([$class: 'Mailer', recipients: 'admin@somewhere'])
+}
 </pre>
-</div>
\ No newline at end of file
+    <p>For all other cases, use plain <code>try</code>-<code>catch</code>(-<code>finally</code>) blocks.
```
kohsuke: This is where I meant to leave the code example. For some people, "plain try-catch-finally block" isn't obvious. In my original version I was trying to emphasize that by having the actual snippet.

kohsuke:
I still stand by my comment, but in the interest of not blocking a progress, I'll leave :+1: here to leave it up to Jesse's discretion.

kohsuke:
:+1: because PROCESS!! /s

recampbell:
:+1:

PR #101, opened 2015-03-27
Fix to SemaphoreStep semantics

jglick:
Fixes hard-to-diagnose test failures like one encountered in #37. @reviewbybees


kohsuke:
Tangental comment: it would be good to improve the documentation of `SemaphoreStep` at some point. It's not very obvious to me how you use this.

:+1:

PR #102, opened 2015-03-27
Misc fixes

jglick:
Some changes I made in #37 while diagnosing test failures which are not actually specific to that branch and could be useful earlier (without the core baseline change, etc.). @reviewbybees


recampbell:
:+1:

kohsuke:
:+1:

PR #104, opened 2015-03-27
[JENKINS-26128] withEnv

jglick:
[JENKINS-26128](https://issues.jenkins-ci.org/browse/JENKINS-26128)

- [x] clean up docs
- [x] reconsider type of `overrides` (lack of decent databinding options is painful)

Would be useful to update `flow.groovy` in the Docker demo, but that will conflict with #98 which has not yet been reviewed and I do not care to pull that in as a dependency too.

@reviewbybees


step-api/src/test/java/org/jenkinsci/plugins/workflow/steps/StepConfigTester.java
```diff
@@ -85,6 +89,20 @@ public StepConfigTester(JenkinsRule rule) {
             @Override public String getDisplayName() {
                 return "Test step builder";
             }
+            @Override public Builder newInstance(StaplerRequest req, JSONObject formData) throws FormException {
+                // TODO f:dropdownDescriptorSelector does not seem to work sensibly: the super method uses RequestImpl.bindJSON and ignores any StepDescriptor.newInstance override.
```
jglick: @kohsuke FYI this seems like a flaw in databinding.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
```diff
@@ -0,0 +1,124 @@
...
+                    "env.NULLED = 'outside'\n" +
+                    "node {\n" +
+                    "  withEnv(['CUSTOM=override', 'NOVEL=val', 'BUILD_TAG=custom', 'NULLED=', 'FOOPATH+BALL=/opt/ball']) {\n" +
+                    "    sh 'echo inside CUSTOM=$CUSTOM NOVEL=$NOVEL BUILD_TAG=$BUILD_TAG NULLED=$NULLED FOOPATH=$FOOPATH:'\n" +
+                    "    echo \"groovy NULLED=${env.NULLED}\"\n" +
+                    "  }\n" +
+                    "  sh 'echo outside CUSTOM=$CUSTOM NOVEL=$NOVEL NULLED=outside:'\n" +
+                    "}"));
+                WorkflowRun b = story.j.assertBuildStatusSuccess(p.scheduleBuild2(0));
+                story.j.assertLogContains("inside CUSTOM=override NOVEL=val BUILD_TAG=custom NULLED= FOOPATH=/opt/ball:/opt/foos:", b);
```
jtnord: Can you normally override BUILD_TAG in freestyle with the envinject plugin?  
If you can seems like a bug - I would not expect that you can override builtin Jenkins contributed variables (admittedly I once wrote a plugin that did exactly that, but plugin code and variables are different.)

jglick: I do not know what you can do with EnvInject. Anyway I would not expect this to be useful, just testing what is possible. It seems harmless to me.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
```diff
@@ -0,0 +1,124 @@
...
+                story.j.assertLogContains("groovy NULLED=null", b);
```
jtnord: Hmmm.
nulled= should be an empty string not null.  In the shell I would expect that it sets a variable (that is if you execute set or ENV i would expect it to be returned.)  Which isn't possible with the current syntax - so maybe this should be a map (key/value) instead? 

if you want it to be unset (or nulled) then seems like withEnv(['NULLED'=null, 'FOOBAR=WIBBLE'])

_Note_: windows behaves differently to bash here - which may make my arguments pointless :-/
```
$ export BOGUS=
$ env | grep BOGUS
BOGUS=
$

C:\>set WIBBLE=
C:\>set WIBBLE
Environment variable WIBBLE not defined
```

jglick: `EnvVars.override` handles an empty string by unsetting the environment variable and I think that is what people expect. As you note, Windows does not seem to even support empty values.

Making `overrides` be a `Map<String,String>` seems attractive but I do not think it would work well with the DSL invocation syntax because it would be confused with the parameter map (unless you explicitly specified `overrides:`, which is too verbose).

jtnord:
:+1: I'm convinced

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/EnvStepTest.java
jglick: (To be clear, I prefer to have the test assert all interesting aspects of behavior, even the questionable ones, so that any future change in behavior has to be a conscious choice that touches test code.)

recampbell:
:+1:

PR #105, opened 2015-03-30
[JENKINS-27474] Noting a few places where fileExists would be useful

jglick:
Follow-ups to #91 and 205f27b059414466f83ca4de7f99921c93336821. @reviewbybees


aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/scm/MercurialStepTest.java
```diff
@@ -90,7 +90,7 @@ private static void hg(File repo, String... cmds) throws Exception {
             "        }\n" +
             "        dir('other') {\n" +
             "            checkout([$class: 'MercurialSCM', source: $/" + otherRepoU + "/$, clean: true])\n" +
-            "            try {\n" +
+            "            try {\n" + // TODO or use fileExists
```
jglick: BTW `WaitForConditionStepTest.catchErrors` also uses this idiom but catching `FileNotFoundException` there is useful for test coverage purposes as explained in a comment there.

stephenc:
:+1:

tfennelly:
:+1:

PR #106, opened 2015-03-30
[JENKINS-27631] Added SecretPickle

jglick:
Needed for [JENKINS-27631](https://issues.jenkins-ci.org/browse/JENKINS-27631). @reviewbybees


stephenc:
:+1:

tfennelly:
:+1: 

Though not sure why the constructor converts to a String and `rehydrate` converts back to a `Secret`.

jglick:
> not sure why the constructor converts to a `String` and `rehydrate` converts back to a `Secret`

Because I want the serial form to be encrypted. https://github.com/jenkinsci/credentials-binding-plugin/issues/6 demonstrates what fails if this is not done: `program.dat` includes the password value from the environment map inside the block step (until the block exits).

PR #41, opened 2015-01-22
[JENKINS-26552] Provide a full Computer.environment to @StepContextParameter EnvVars

jglick:
[JENKINS-26552](https://issues.jenkins-ci.org/browse/JENKINS-26552)

- [X] fix adequate for #37
- [x] stricter & expanded functional test

@reviewbybees

PR #110, opened 2015-03-31
Raise precedence of imperative env.VAR

jglick:
Docker demo sanity check prior to attempted 1.5 release turned up a regression, which I think was introduced by #41, though possibly by #104. Finally testing this, and also extracting previously duplicated code into a new helper method to make it easier to follow.

Really setting `env.VAR = 'value'` was a poor design to begin with (`withEnv` is better), but it is out there and needs to remain supported.

@reviewbybees

### DIFF
--- a/step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvironmentExpander.java
+++ b/step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvironmentExpander.java
@@ -25,6 +25,8 @@
 package org.jenkinsci.plugins.workflow.steps;
 
 import hudson.EnvVars;
+import hudson.model.Computer;
+import hudson.model.Run;
 import java.io.IOException;
 import java.io.Serializable;
 import javax.annotation.CheckForNull;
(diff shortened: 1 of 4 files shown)


jglick:
(I also tested this against https://github.com/jenkinsci/workflow-plugin/pull/37 and https://github.com/jenkinsci/credentials-binding-plugin/pull/6 and tests passed.)

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvironmentExpander.java
```diff
@@ -67,4 +69,32 @@ public static EnvironmentExpander merge(@CheckForNull EnvironmentExpander origin
...
+     * @return the effective environment
+     */
+    public static @Nonnull EnvVars getEffectiveEnvironment(@Nonnull EnvVars customEnvironment, @CheckForNull EnvVars contextualEnvironment, @CheckForNull EnvironmentExpander expander) throws IOException, InterruptedException {
+        EnvVars env;
+        if (contextualEnvironment != null) {
+            env = new EnvVars(contextualEnvironment);
+            env.putAll(customEnvironment);
+        } else {
+            env = new EnvVars(customEnvironment);
+        }
```
tfennelly: Perhaps could be simplified further to:

```
EnvVars env = new EnvVars();
if (contextualEnvironment != null) {
    env.putAll(contextualEnvironment);
}
env.putAll(customEnvironment);
```

tfennelly:
:+1:

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/EnvironmentExpander.java
jglick: Unfortunately those are not necessarily identical, because an `EnvVars` also holds a `platform` field which is left unset in the no-arg constructor. It is transferred in the copy constructor but only originally set in `masterEnvVars`, which in this case (assuming you are currently inside a `node` block, otherwise you do not care about the platform) would come via the `contextualEnvironment` passed from `PlaceholderExecutable` calling `Computer.getEnvironment` (which calls `getRemote`, which loads `masterEnvVars` in the slave JVM, meaning the slave’s environment variables, not the master’s).

If you forgot and started with the no-arg constructor, all of your tests would still pass, because `override` falls back to using `File.pathSeparatorChar` (“it's still more error prone that _[sic]_ I'd like” according to the comment), which from a test would be the same on a slave as on the master. But anything using `PATH+XXX` syntax might fail in the field.

Confused yet? I am, which is why I want to be done with refactoring this logic. :-)

PR #111, opened 2015-04-02
Allow specifying any commit to checkout from Git

lptr:
With Git SCM in a regular job the user can specify a commit hash, but this doesn't work with the `git` step. I guess this change should fix the problem.


jglick:
There is no test case proving that your use case works. (At least the existing `branch` use case _is_ covered by a test.) Nor is there update to help text, and passing a commit hash to a parameter named `branch` is a bit surprising.

There are various enhancements that could be made to the `git` step but my intention is to reject most of them, since you can already use the `checkout` step with `GitSCM` to cover any more unusual use cases.

jglick:
[JENKINS-27018](https://issues.jenkins-ci.org/browse/JENKINS-27018) BTW.

PR #113, opened 2015-04-08
Do not delete the test project from configRoundTrip while the test is running

jglick:
Commit comment explains why. @reviewbybees


tfennelly:
:+1: 

So how does it get deleted then? Workspace gets blown away and so project with it I assume.

jglick:
> So how does it get deleted then?

The whole `$JENKINS_HOME` directory is deleted at the end of every test case anyway.

PR #114, opened 2015-04-09
Call new Controller.stop overload

jglick:
Just using https://github.com/jenkinsci/durable-task-plugin/pull/5.

@reviewbybees


tfennelly:
:+1:

PR #119, opened 2015-04-28
Stage execute all builds

henrikcaesar:
Modified the stage step to have the option to execute all builds that enter the stage (no preemption).

Couldn't come up with a better name than `discardOldBuilds`. 
Suggestion for a better name is welcome :)

### DIFF
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/StageStep.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/StageStep.java
@@ -35,12 +35,14 @@ import org.kohsuke.stapler.DataBoundSetter;
  * Marks a flow build as entering a gated “stage”, like a stage in a pipeline.
  * Each job has a set of named stages, each of which acts like a semaphore with an initial permit count,
  * but with the special behavior that only one build may be waiting at any time: the newest.
+ * To disable this behavior and have all builds queue up and execute set discardOldBuilds to false.
  * Credit goes to @jtnord for implementing the {@code block} operator in {@code buildflow-extensions}, which inspired this.
  */
 public final class StageStep extends AbstractStepImpl {
 
     public final String name;
     @DataBoundSetter public @CheckForNull Integer concurrency;
+    @DataBoundSetter public boolean discardOldBuilds = true;
 
     @DataBoundConstructor public StageStep(String name) {
         if (name == null || name.isEmpty()) {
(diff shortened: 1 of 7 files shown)


henrikcaesar:
I ran the suites in Netbeans and it worked fine, but I am also able to reproduce the fault now when running mvn on the command line. Will fix it.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/StageStep.java
```diff
@@ -35,12 +35,14 @@
...
  * Each job has a set of named stages, each of which acts like a semaphore with an initial permit count,
  * but with the special behavior that only one build may be waiting at any time: the newest.
+ * To disable this behavior and have all builds queue up and execute set discardOldBuilds to false.
  * Credit goes to @jtnord for implementing the {@code block} operator in {@code buildflow-extensions}, which inspired this.
  */
 public final class StageStep extends AbstractStepImpl {
 
     public final String name;
     @DataBoundSetter public @CheckForNull Integer concurrency;
+    @DataBoundSetter public @CheckForNull Boolean discardOldBuilds = true;
```
jglick: Should be `boolean` not `@CheckForNull Boolean`.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/StageStepExecution.java
```diff
@@ -260,36 +274,38 @@ private static void cancel(StepContext context, StepContext newer) throws IOExce
     }
 
     private static final class Stage {
+        /** Maximum permitted size of {@link #holding}, or null for unbounded. */
+        @CheckForNull Integer concurrency;
```
jglick: Please avoid reformatting existing code. It makes the PR diff too long and hard to review. The fewer diff lines, the better.

jglick:
I have not yet had a chance to think deeply about what you are doing here, but is this basically [JENKINS-27039](https://issues.jenkins-ci.org/browse/JENKINS-27039)?

henrikcaesar:
Thanks will fix what you pointed out. I had not seen that issue This comment from the issue describes what I have implemented:

> I have a stage with "concurrency: 1" that I know isn't a bottleneck and I'd like a "patient: true" type option that would queue up all incoming builds without dropping any.

Could also be implemented as a separate step I guess...

jglick:
So I think that should really be a separate step, which would do no cancelation and probably need not add a `StageAction` either, but merely implement a mutual exclusion region. `stage` is really intended to define a fixed sequence of stages with some builds being “left behind”, which is a very different use case.

henrikcaesar:
Okay that sounds good. Keeping it in a separate step also reduces the risk of making StageStep over complicated and introducing bugs.

Do you mean something like this?

    block {
       // do some work
    }

Is block a good name? Can also think of sequentially? I can add a new issue for a feature request where this can be discussed instead of in this PR.

Do you want me to create a separate PR for the tests I added to StageTests? Those could be useful even without the `discardOldBuilds`option.

jglick:
Maybe `withMonitor`? Not sure what name is most intuitive.

A PR for expansion of `StageStepTest` would be great.

PR #121, opened 2015-04-30
[JENKINS-28063] Handle redundant upstream builds

jglick:
[JENKINS-28063](https://issues.jenkins-ci.org/browse/JENKINS-28063)

@reviewbybees

### DIFF
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
@@ -8,16 +8,12 @@ import hudson.model.listeners.RunListener;
 
 import javax.annotation.Nonnull;
 
-/**
- * @author Vivek Pandey
- */
 @Extension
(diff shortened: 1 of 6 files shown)


support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
```diff
@@ -8,16 +8,12 @@
 
 import javax.annotation.Nonnull;
 
-/**
- * @author Vivek Pandey
```
oleg-nenashev: Is it a penalty for the missing Javadoc? :)

oleg-nenashev:
:+1:

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
jglick: Ha! No, I just looked at this class and realized that essentially nothing of what he originally wrote remains unedited, so leaving the attribution there is just misleading. If you want to know what someone wrote, check Git history.

PR #124, opened 2015-05-01
[JENKINS-28179] Workspace suffix should respect hudon.slaves.WorkspaceList property

rpocase:
These changes resolve JENKINS-28179 by ensuring the suffix for concurrent builds is generated by using hudson.slaves.WorkspaceList. On the off change this isn't defined or is an empty string, it uses the previous value (@) as the workspace suffix separator. 

https://issues.jenkins-ci.org/browse/JENKINS-28179

### DIFF
--- a/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
+++ b/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
@@ -40,6 +40,7 @@ import hudson.slaves.WorkspaceList;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
+import java.util.Properties;
 import javax.inject.Inject;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.cps.persistence.PersistIn;
@@ -91,7 +92,7 @@ public class CpsScmFlowDefinition extends FlowDefinition {
             if (baseWorkspace == null) {
                 throw new IOException(node.getDisplayName() + " may be offline");
             }
-            dir = baseWorkspace.withSuffix("@script");
+            dir = getFilePathWithSuffix(baseWorkspace);
         } else { // should not happen, but just in case:
             dir = new FilePath(owner.getRootDir());
         }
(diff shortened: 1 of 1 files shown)


cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -119,6 +120,18 @@ public String getScriptPath() {
         return exec;
     }
 
+    private FilePath getFilePathWithSuffix(FilePath baseWorkspace) {
+        return baseWorkspace.withSuffix(getFilePathSuffix());
```
jglick: Uh, I think you forgot to add `+ "script"`.

rpocase: I definitely did. I'll get that fixed.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -119,6 +120,18 @@ public String getScriptPath() {
...
+    }
+
+    private String getFilePathSuffix() {
+        String defined_suffix = new Properties().getProperty("hudson.slaves.WorkspaceList");
```
jglick: This will not work. It will always be null. Did you test this patch? (`mvn -f aggregator hpi:run -Dhudson.slaves.WorkspaceList=~` or similar)

Best to simply copy the [original definition of `COMBINATOR`](https://github.com/jglick/jenkins/blob/ca3b1db4d6f761861885d03355f35ccafe7cc425/core/src/main/java/hudson/slaves/WorkspaceList.java#L291).

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -119,6 +120,18 @@ public String getScriptPath() {
         return exec;
     }
 
+    private FilePath getFilePathWithSuffix(FilePath baseWorkspace) {
+        return baseWorkspace.withSuffix(getFilePathSuffix() + "script");
+    }
+
+    private String getFilePathSuffix() {
+        String defined_suffix = new Properties().getProperty("hudson.slaves.WorkspaceList");
```
jglick: Still wrong (see comment on outdated diff).

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
rpocase: Yep, it worked on a local instance. I'll update it to you that definition.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java
```diff
@@ -40,6 +40,7 @@
 import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
+import java.util.Properties;
```
jglick: Orphaned import, but no matter.

jglick:
> Is there common code to get defined system properties?

No.

PR #126, opened 2015-05-04
Extension point for adding pseudovariables

jglick:
New extension point useful for adding DSLs.

@reviewbybees

### DIFF
--- a/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
+++ b/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
@@ -26,6 +26,7 @@ package org.jenkinsci.plugins.workflow.cps;
 
 import groovy.lang.GroovyObjectSupport;
 import hudson.EnvVars;
+import hudson.Extension;
 import hudson.model.Run;
 import hudson.util.LogTaskListener;
 import java.io.IOException;
@@ -51,7 +52,7 @@ public class EnvActionImpl extends GroovyObjectSupport implements EnvironmentAct
     private transient EnvVars ownerEnvironment; // cache
     private transient Run<?,?> owner;
 
-    EnvActionImpl() {
+    private EnvActionImpl() {
         this.env = new TreeMap<String,String>();
     }
 
@@ -106,6 +107,23 @@ public class EnvActionImpl extends GroovyObjectSupport implements EnvironmentAct
 
     @Override public void onLoad(Run<?,?> r) {
         owner = r;
+    }
+
+    @Extension public static class Binder extends GroovyShellDecorator {
+        @Override public Object getProperty(CpsScript script, String property) throws Exception {
+            if (property.equals("env")) {
+                Run<?,?> run = script.$build();
+                if (run != null) {
+                    EnvActionImpl action = run.getAction(EnvActionImpl.class);
+                    if (action == null) {
+                        action = new EnvActionImpl();
+                        run.addAction(action);
+                    }
+                    return action;
+                }
+            }
+            return null;
+        }
     }
 
 }
--- a/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
+++ b/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
@@ -109,20 +109,22 @@ public class EnvActionImpl extends GroovyObjectSupport implements EnvironmentAct
         owner = r;
     }
 
-    @Extension public static class Binder extends GroovyShellDecorator {
-        @Override public Object getProperty(CpsScript script, String property) throws Exception {
-            if (property.equals("env")) {
-                Run<?,?> run = script.$build();
-                if (run != null) {
-                    EnvActionImpl action = run.getAction(EnvActionImpl.class);
-                    if (action == null) {
-                        action = new EnvActionImpl();
-                        run.addAction(action);
-                    }
-                    return action;
+    @Extension public static class Binder extends Singleton {
+        @Override public String getName() {
+            return "env";
+        }
+        @Override public Object getProperty(CpsScript script) throws Exception {
+            Run<?,?> run = script.$build();
+            if (run != null) {
+                EnvActionImpl action = run.getAction(EnvActionImpl.class);
+                if (action == null) {
+                    action = new EnvActionImpl();
+                    run.addAction(action);
                 }
+                return action;
+            } else {
+                throw new IllegalStateException("no associated build");
             }
-            return null;
         }
     }
 
--- a/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
+++ b/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
@@ -109,11 +109,11 @@ public class EnvActionImpl extends GroovyObjectSupport implements EnvironmentAct
         owner = r;
     }
 
-    @Extension public static class Binder extends Singleton {
+    @Extension public static class Binder extends GlobalVariable {
         @Override public String getName() {
             return "env";
         }
-        @Override public Object getProperty(CpsScript script) throws Exception {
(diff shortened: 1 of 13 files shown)


kohsuke:
See relevant PoC: https://github.com/kohsuke/workflow-plugin/commit/3d09c0a0dbc02d133da75774aab7c4924aac7013

jglick:
Thanks but I do not need that PoC and it does not apply to the use case I have in mind.

kohsuke:
I'd really love to see this go one step further, so that plugins that define those additional syntaxes can do so without writing Java code. Just annotated Groovy script or Groovy class that follows the specific naming convention should do (for example, `fooBar` → `org.jenkinsci.plugins.workflow.cps.constants.FooBar`)

jglick:
:-1: on magical binding. It is tricky enough to get things just right with Java code, given the hazards of class loading, whitelists, and `CpsCallableInvocation`s getting caught in unexpected places. If in the future there turns out to be a standard pattern shared by several plugins, it can be codified into a convenience API.

jglick:
Temporarily retracing this API proposal, since I realized that as written it is not going to allow documentation (such as from the Snippet Generator), which would be really desirable. Needs to be tweaked so that the extension can advertise the variable name it binds.

Considered putting this in a lower API plugin (`workflow-api`?) as it could seemingly be language-independent. For `currentBuild` that would perhaps make sense, assuming the implementation of `$build` were essentially copied (it is short anyway), and the parameter given to the API were `FlowExecution` or `StepContext`. (`EnvActionImpl` is in `workflow-cps` anyway so it does not matter.) For stateful DSL variables, merely returning a new object is not sufficient, however; they need to _save_ the value so that the same instance is returned on subsequent calls. Allowing the extension to communicate this to the caller (`CpsScript.getProperty`) would require another flag (`isTransient`?), or perhaps a helper method to do the saving.

jglick:
Could not work out a way for this to make sense in `workflow-api`. A DSL implementation needs access to the `CpsScript`.

jglick:
Also could not see a way to pass `StepContext`, which has too many things specific to running `Step`s. `EnvActionImpl` would benefit from the `get` method rather than its hackish call to `CpsThread.current()`, but generalizing this would require a significant new API, and anyway `EnvActionImpl` is Groovy-specific (since it implements `GroovyObject.getProperty`), so does not really make sense to move outside `workflow-cps`.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
```diff
@@ -108,4 +109,21 @@
...
+        @Override public Object getProperty(CpsScript script, String property) throws Exception {
+            if (property.equals("env")) {
+                Run<?,?> run = script.$build();
+                if (run != null) {
+                    EnvActionImpl action = run.getAction(EnvActionImpl.class);
+                    if (action == null) {
+                        action = new EnvActionImpl();
+                        run.addAction(action);
+                    }
+                    return action;
```
tfennelly: Does `Run` not have a `getEnv` method wrapping access to the actions etc? I know it's API could explode if you added everything to it, but this one seems fundamental to me.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/Singleton.java
```diff
@@ -0,0 +1,54 @@
...
+import groovy.lang.GroovyObject;
+import hudson.ExtensionPoint;
+import javax.annotation.Nonnull;
+import jenkins.model.RunAction2;
+
+/**
+ * Defines a provider of a singleton (global variable) offered to flows.
+ * Should have a view named {@code help} offering usage.
+ */
+public abstract class Singleton implements ExtensionPoint {
```
tfennelly: What is the scope of this "singleton"? A `Run`? My gut tells me to stay away from using the word "Singleton" in the naked form i.e. add something to it `RunFlowSingleton`.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
```diff
@@ -108,4 +109,23 @@
...
+        }
+        @Override public Object getProperty(CpsScript script) throws Exception {
+            Run<?,?> run = script.$build();
+            if (run != null) {
+                EnvActionImpl action = run.getAction(EnvActionImpl.class);
+                if (action == null) {
+                    action = new EnvActionImpl();
+                    run.addAction(action);
+                }
+                return action;
```
tfennelly: Not relevant to this PR really but, does `Run` not have a `getEnv` method wrapping access to the actions etc? I know it's API could explode if you added everything to it, but this one seems fundamental to me.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/Singleton.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Gets or creates the singleton object.
+     * If the object is stateful, and the state should not be managed externally (such as with a {@link RunAction2}),
+     * then the implementation is responsible for saving it in the {@link CpsScript#getBinding}.
+     * @param script the script we are running
+     * @return a POJO or {@link GroovyObject}
+     * @throws Exception if there was any problem creating it (will be thrown up to the script)
+     */
+    public abstract @Nonnull Object getProperty(CpsScript script) throws Exception;
```
tfennelly: Should there be a `@Nonnull` on the `script` arg?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/Singleton.java
jglick: Yes, it is within one flow execution.

`RunFlowSingleton` does not make much sense but some variant might. Or would `GlobalVariable` be clearer?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/Singleton.java
jglick: Yes, that would be a good idea.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
jglick: There is no `Run.getEnv` method. I am not sure what you are saying.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/Singleton.java
tfennelly: Sure, I was just throwing `RunFlowSingleton` out there expecting it to not make sense. `GlobalVariable` prob would but I'd qualify what is meant by "global" (I assume `Run`)?

jglick: Typically a `Run`. Technically one `CpsScript`.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
tfennelly: I'm saying that I think it would make sense to have one Vs having to get Actions etc (not intuitive). As I said, prob not relevant to this PR other than it highlighted that it was ugly (to me).

tfennelly: My point was ... you had to jump through ugly and unintuitive code hoops in this PR, just to get the `EnvVars` for a `Run`. Seems like an obvious addition to the `Run` API.

tfennelly:
:+1:

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/EnvActionImpl.java
jglick: No, `EnvActionImpl` is very much specific to Workflow.

PR #127, opened 2015-05-05
Advanced steps

jglick:
Adding `StepDescriptor.isAdvanced` to segregate out steps which most users should not see most of the time.

Subsumes #126.

@reviewbybees

### DIFF
--- a/cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
+++ b/cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
@@ -24,12 +24,10 @@ THE SOFTWARE.
 -->
 
 <?jelly escape-by-default='true'?>
-<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:f="/lib/form" xmlns:l="/lib/layout">
-    <f:optionalBlock name="snippetizer" title="Snippet Generator" help="/${it.HELP_URL}">
-        <f:section title="Steps"/>
-        <!-- Similar to f:dropdownDescriptorSelector, but adds fallback content to block, and JENKINS-25130 adds per-selection help: -->
-        <f:dropdownList name="prototype" title="Sample Step">
-            <j:forEach var="stepDescriptor" items="${it.stepDescriptors}">
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:f="/lib/form" xmlns:l="/lib/layout" xmlns:d="jelly:define" xmlns:local="local">
+    <d:taglib uri="local">
+        <d:tag name="listSteps">
+            <j:forEach var="stepDescriptor" items="${stepDescriptors}">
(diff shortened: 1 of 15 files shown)


cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
```diff
@@ -49,6 +48,15 @@ THE SOFTWARE.
...
                 </f:dropdownListBlock>
             </j:forEach>
+        </d:tag>
+    </d:taglib>
+    <f:optionalBlock name="snippetizer" title="Snippet Generator" help="/${it.HELP_URL}">
+        <f:section title="Steps"/>
+        <!-- Similar to f:dropdownDescriptorSelector, but adds fallback content to block, and JENKINS-25130 adds per-selection help: -->
+        <f:dropdownList name="prototype" title="Sample Step">
+            <local:listSteps stepDescriptors="${it.getStepDescriptors(false)}"/>
+            <f:dropdownListBlock title="${%— Advanced/Deprecated —}"/>
```
tfennelly: "Advanced" or "Deprecated" ? So if not deprecated it is advanced? I don't get that.

jglick: Yeah, either advanced or deprecated (or both).

tfennelly: Ok, I see what you mean after looking at https://github.com/cloudbees/cloudbees-workflow-docker-plugin/pull/19.

It's probably a NIT because what you have obviously works and is not wrong, but `advanced/deprecated` does not really seem to me like the right descriptor property.

tfennelly:
:+1:

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
jglick: Would be easy to rename it (up until the next plugin release) if you have a different suggestion.

tfennelly: @jglick yeah if we can keep it private then we have options later.

I guess it wasn't obvious what it was about. The `Descriptor` interface method was called `isAdvanced`, yet the javadoc was also saying "or deprecated". "Deprecated or advanced" seem to me like a strange pairing when you don't have the full context. Maybe a bit more javadoc (with an example) would help that. Even with the context, "advanced" seems a bit odd to me because the step you want to hide might not be all that advanced as I'm sure there will be other reasons for hiding steps.

When you mentioned earlier that these steps should be hidden somehow, I was thinking of something like the `@InternalUseOnly` (forget the exact name but I know you know what I'm talking about) annotation. Maybe it's not appropriate to use the annotation, but the use of the term "InternamUseOnly" seems more accurate in itself.

PR #128, opened 2015-05-07
Need to also push the latest tag.

jglick:
@reviewbybees

### DIFF
--- a/demo/Makefile
+++ b/demo/Makefile
@@ -25,3 +25,4 @@ push:
 
 push-latest: push
 	docker tag -f $(IMAGE):$(TAG) $(IMAGE):latest
+	docker push $(IMAGE):latest


oleg-nenashev:
:+1:

demo/Makefile
```diff
@@ -25,3 +25,4 @@ push:
 
 push-latest: push
```
tfennelly: So this target is going to push 2 images?

tfennelly:
:+1:

demo/Makefile
jglick: Well, one image with two tags. Currently

```
$ curl -s https://registry.hub.docker.com/v1/repositories/jenkinsci/workflow-demo/tags | json_pp
[
   {
      "name" : "latest",
      "layer" : "6b64b99c"
   },
   {
      "layer" : "d270500e",
      "name" : "1.5"
   },
   {
      "layer" : "6b64b99c",
      "name" : "1.6"
   },
   {
      "name" : "1.6-alpha-1",
      "layer" : "956a5bcd"
   },
   {
      "name" : "1.7-alpha-1",
      "layer" : "9d6e971f"
   }
]
```

PR #131, opened 2015-05-12
[JENKINS-26135] Allow CPS global lib scripts to define global variables

kohsuke:
This allows the release engineering team to define higher-level organization specific workflow DSLs.

This is critical to use workflow as "simplified text-based UX". Such DSLs could be for example:

```
acmeCorp {
    id = '123'
    devEnvironment = 'qa-svr-1'
}
```

(where let's say there's an inventory of application ID in the central database describing where the repository is, and 'qa-svr-1' refers to the Tomcat server where the app gets deployed at the end.)

- [X] PoC
- [x] Define top-level `vars` dir and pick scripts from there, not from `src`. _(perhaps)_
- [x] Tweak `GlobalVariable` API to get rid of dynamic extension list manipulation.
- [x] Doc updates.
- [x] Fall back to plain text help
- [x] Solve [JENKINS-26135](https://issues.jenkins-ci.org/browse/JENKINS-26135)
- [X] Use [JENKINS-28586](https://issues.jenkins-ci.org/browse/JENKINS-28586) fix from https://github.com/jenkinsci/script-security-plugin/pull/13
- [x] Deal with regressions from JENKINS-28586; e.g. `LoadStepTest.basics` failure

@reviewbybees


kohsuke:
@reviewbybees

kohsuke:
I could have avoided registering/unregistering a new `Extension` on the fly if `GlobalVariable` supports a collection of global variables, not just one.

It looks like `GlobalVariable` isn't released yet, so maybe you are open to changing it.

oleg-nenashev:
* I'm aware that <code>UserDefinedGlobalVariableList</code> takes all scripts in ${Workspace}/src folder. It means that you cannot place any other Groovy script there, so it may cause a conflict. Probably you could get scripts from somewhere like ${Workspace}/src/**globalVars/** or use a predefined wildcard
* One script can contribute only one variable. I suppose it's not user-friendly in the use-case of "I retrieve an bulk info from a remote component and want to save it to 25 variables" (e.g. SCM stats, Docker server info, etc.)

jglick:
```
[INFO] Possible null pointer dereference in new org.jenkinsci.plugins.workflow.cps.global.UserDefinedGlobalVariableList() due to return value of called method [org.jenkinsci.plugins.workflow.cps.global.UserDefinedGlobalVariableList, org.jenkinsci.plugins.workflow.cps.global.UserDefinedGlobalVariableList] Dereferenced at UserDefinedGlobalVariableList.java:[line 29]Known null at UserDefinedGlobalVariableList.java:[line 29]
[INFO] Possible null pointer dereference in org.jenkinsci.plugins.workflow.cps.global.UserDefinedGlobalVariableList.init() due to return value of called method [org.jenkinsci.plugins.workflow.cps.global.UserDefinedGlobalVariableList, org.jenkinsci.plugins.workflow.cps.global.UserDefinedGlobalVariableList] Dereferenced at UserDefinedGlobalVariableList.java:[line 69]Known null at UserDefinedGlobalVariableList.java:[line 69]
```

jglick:
> It looks like `GlobalVariable` isn't released yet

Well, it is in 1.7-alpha-1, but I think it is fine to change its signature if that makes this easier.

cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariable.java
```diff
@@ -0,0 +1,72 @@
...
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Global variable backed by user-supplied script.
+ *
+ * @author Kohsuke Kawaguchi
+ */
+// not @Extension because these are manually registered
+public class UserDefinedGlobalVariable extends GlobalVariable {
```
jglick: Need not be `public` I guess.

jglick:
As with @oleg-nenashev I am a little concerned about overloading `$JENKINS_HOME/workflow-libs/src/`, which is already used for classes you intend to `import`, in a package structure. Would seem safer to use a separate subdirectory for this RFE that emphasizes that it should contain files in a flat structure (no packages) and that the file name becomes a variable. IIRC the reason we decided to use the `src/` subdirectory in 1.0 for this plugin was exactly so that we could compatibly introduce other subdirectories for different purposes.

jglick:
BTW you should add a `CHANGES.md` entry.

cps-global-lib/src/main/resources/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariable/help.jelly
```diff
@@ -0,0 +1,4 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core">
+  <j:out value="${it.helpHtml}"/>
```
jglick: Careful—this is user content emitted raw. It would be safer to use `MarkupFormatter`, or restrict to plain text. Admittedly it currently requires `RUN_SCRIPTS` to push to this repo, but see [JENKINS-26538](https://issues.jenkins-ci.org/browse/JENKINS-26538).

cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariableList.java
```diff
@@ -0,0 +1,73 @@
...
+            }
+        }
+    }
+
+    /**
+     * Generates the initial list.
+     */
+    @Initializer(fatal=false,after= InitMilestone.EXTENSIONS_AUGMENTED,before=InitMilestone.JOB_LOADED)
+    public static void init() {
+        Jenkins.getInstance().getExtensionList(UserDefinedGlobalVariableList.class).get(UserDefinedGlobalVariableList.class).rebuild();
```
jglick: Definitely this is ugly and would benefit from a change to `GlobalVariable` to return a `Set<String>` names and take a `String name` parameter to `getValue`.

(Could even do so compatibly by introducing a `GlobalVariables` extension point with that signature, then retrofitting `GlobalVariable` to implement it with `final` methods delegating to its current signature, though this seems like overkill for an API introduced only in an alpha.)

jglick:
> I retrieve an bulk info from a remote component and want to save it to 25 variables

I guess this use case, if it ever comes up, could be treated as an additional feature—so long as we use distinctive subdirectory names.

Other than the comments above this looks good to me.

BTW have you checked that this functionality is not already possible without any change? Since `src/` is already in the script “classpath”, it seems that Groovy might have interpreted references to `someVar.someMethod` as `static` calls to a class in the default package. But maybe you then run aground trying to access an implicit `CpsScript` (for steps or other global variables) from a static context.

jglick:
As an aside, the consistent feedback I hear from everyone I talk to about Workflow scripts is that they want to keep all sources in their own Git repositories (or other SCM), and thus would not use this plugin. Not blocking this PR but we should be considering how similar features could be built using external SCMs.

kohsuke:
* I thought about possibility of name conflicts, and putting those global variables into a fixed package or a separate directory (like `${globalLibs}/vars` or something like that.) But a full day conversations with users the other day made me feel that this is probably the most common use case for CPS global libs that I will be pushing to users, so occupying the root package felt justifiable. It also felt natural; if you write `org.acme.foo` then it maps to `src/org/acme/foo.groovy`, so if it's just `acme`, one would expect that it would map to `src/foo.groovy`. I'll wait for others to chime in.
* `GlobalVariable` is unsuitable for dynamically listed variables like that, but I don't think that's an issue --- certainly not a problem in this PR. It's not really meant to be used for simple values, it's meant to be used more like a namespace.

jglick:
> It also felt natural; if you write `org.acme.foo` then it maps to `src/org/acme/foo.groovy`, so if it's just `acme`, one would expect that it would map to `src/foo.groovy`.

Yet these behave in subtly different ways. One creates a class definition, one a variable. Just worried that this is adding yet another mine to the field; see the explanation of explicit vs. implicit classes in `cps-global-lib/README.md`. (BTW this PR should include a patch to that page to document the new feature.)

Do not feel strongly about it, but I still find it clearer to use a separate directory namespace for a distinct feature.

> It's not really meant to be used for simple values, it's meant to be used more like a namespace.

Not sure I followed that. Anyway this PR should do _something_ to clean up the interaction with `GlobalVariable`—whether changing its API, or introducing a new “friend API” in `workflow-cps` that `GlobalVariable` would implement as a special case. Dynamic modifications to `ExtensionList` (and new `@Initializer`s) are never acceptable unless there is no compatible alternative.

cps-global-lib/src/main/resources/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariable/help.jelly
kohsuke: All right, I'll do `MarkupFormatter` and probably name those files as just `*.txt`.

kohsuke:
> they want to keep all sources in their own Git repositories (or other SCM)

People using another Git repository as the source of truth is anticipated and well supported. They just need to "deploy" the changes by pushing it into this repo, which is the same thing Heroku and others do.

Using other SCMs would indeed go beyond the capability of this repository. Maybe we can split `cps-globa-libs` into the part that just assumes populated `$JENKINS_HOME/workflowLibs` and another that makes that a Git repository. In that way we leave options open for a future expansion that does `SCM.checkout()`.

jglick:
> In that way we leave options open for a future expansion that does `SCM.checkout()`.

Better to do nothing until we implement that future expansion and see exactly what it requires. I suspect that it should work like, or even as a part of, `CpsScmScriptDefinition`: when the build starts, multiple repos are checked out and added to the “classpath”.

jglick:
Are you still working on this?

kohsuke:
For the past two days I've been away. My understanding based on the conversation here is that the following tasks remain:

* Define top-level `vars` dir and pick scripts from there, not from `src`.
* Tweak `GlobalVariable` API to get rid of dynamic extension list manipulation.
* Doc updates.

jglick:
> Define top-level `vars` dir

Your judgment call; I am only -0 on the current placement.

jglick:
BTW I took the liberty of moving your list into a GH tasklist for easier tracking.

jglick:
Just remembered [JENKINS-26135](https://issues.jenkins-ci.org/browse/JENKINS-26135), which requests automatically loaded global _functions_ (opposed to _variables_) here. Seems like it would not be hard to generalize this a bit to allow you to hook into `CpsScript.invokeMethod` as well as `.getProperty`, right? Or perhaps you can just define a variable whose value is actually a `Closure`; Groovy would then let you call it like a function, I think (but needs to be proven in a test).

kohsuke:
Good point about JENKINS-26135. I think I should be able to solve it at the same time.

kohsuke:
This PR is feature complete and ready to be merged.

kohsuke:
@reviewbybees to ping the team again.

kohsuke:
This fix requires [script-security-plugin #13](https://github.com/jenkinsci/script-security-plugin/pull/13)

kohsuke:
[Change in groovy-cps](https://github.com/cloudbees/groovy-cps/compare/groovy-cps-1.3...groovy-cps-1.4) for your convenience, which is independent but required for this to work.

kohsuke:
@jglick is kindly going to take over this one.

jglick:
Demonstration of the problem is that this patch fixes the most obvious failing test:

```diff
diff --git a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/steps/LoadStepTest.java b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/steps/LoadStepTest.java
index 941b926..942e7ea 100644
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/steps/LoadStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/steps/LoadStepTest.java
@@ -25,7 +25,7 @@ public class LoadStepTest {
         p.setDefinition(new CpsFlowDefinition(
             "node {\n" +
             "  writeFile text: 'println(21*2)', file: 'test.groovy'\n" +
-            "  println 'something printed'\n" +// make sure that 'println' in groovy script works
+            "  echo 'something printed'\n" +// make sure that 'println' in groovy script works
             "  load 'test.groovy'\n" +
             "}", true));
         WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
```

jglick:
Specifically `LoadStepTest.basics` fails with

```
org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use method groovy.lang.GroovyObject invokeMethod java.lang.String java.lang.Object (com.cloudbees.groovy.cps.impl.CpsClosure println java.lang.String)
 at org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.StaticWhitelist.rejectMethod(StaticWhitelist.java:155)
 at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxInterceptor.onMethodCall(SandboxInterceptor.java:77)
 at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxInterceptor.onMethodCall(SandboxInterceptor.java:60)
 at org.kohsuke.groovy.sandbox.impl.Checker$1.call(Checker.java:148)
 at org.kohsuke.groovy.sandbox.impl.Checker.checkedCall(Checker.java:145)
 at com.cloudbees.groovy.cps.sandbox.SandboxInvoker.methodCall(SandboxInvoker.java:15)
 at WorkflowScript.run(WorkflowScript:3)
 at …
```

The problem seems to be in the handling of `GroovyObject` method calls in `script-security`. When there is no matching method on the apparent receiver, we first check for `invokeMethod` before proceeding to GDK methods and then failure. But when the receiver _does_ have `invokeMethod`, yet it is not whitelisted, the nested call throws `RejectedAccessException` and the search ends there. We should perhaps rather throw a `MissingMethodException` (with a `RejectedAccessException` as `cause`, for diagnosis), and let Groovy continue its search for another scope (in this case from `CpsClosure` to the enclosing `CpsScript`). `onGetProperty` already does something superficially similar, though in this case it is not safe to call `super.onMethodCall`.

jglick:
@kohsuke https://github.com/jenkinsci/script-security-plugin/pull/16 seems to fix the `LoadStepTest` failure. It does not fix the `TimeoutStepTest` failure; this test does not even use the sandbox, so it seems to be a regression in `groovy-cps`.

kohsuke:
@jglick 7b6d786, in conjunction with [cloudbees/groovy-cps #13](https://github.com/cloudbees/groovy-cps/pull/13), fixes `TimeoutStepTest`.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsClosure2.java
```diff
@@ -0,0 +1,36 @@
...
+ * {@link CpsClosure} that intercepts the {@code sleep} call so that it gets handled via SleepStep,
+ * instead of {@link DefaultGroovyStaticMethods#sleep(Object, long)} that Groovy adds to {@code Object}.
+ *
+ * <p>
+ * We had to do this because we made a mistake of having the 'sleep' step pick the same name as
+ * a method defined on {@code Object}. Granted, it is a method added by Groovy, not by JDK, but the end result
+ * is still the same, and the consequence is as severe as trying to override {@code hashCode()} method
+ * and use it for something completely different. We ended up doing this because when we did it, a bug masked
+ * the severity of the problem. In hindsight, we should have thought twice before adding a work around
+ * like {@link CpsScript#sleep(long)}.
```
jglick: FWIW, it was quite intentional to override `Object.sleep`, since the original caused such a problem in the CPS VM.

kohsuke:
All the regressions are fixed in conjunction with https://github.com/jenkinsci/script-security-plugin/pull/19

kohsuke:
The build failure is due to the use of SNAPSHOT in POM to pick up https://github.com/jenkinsci/script-security-plugin/pull/19. This change is ready to merge.

cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariable.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * Loads help from user-defined file, if available.
+     */
+    public String getHelpHtml() throws IOException {
+        File help = source(".txt");
+        if (!help.exists())     return null;
+
+        return new EscapedMarkupFormatter().translate(FileUtils.readFileToString(help, Charsets.UTF_8));
```
jglick: This is simply a roundabout way of saying `Util.escape`. But do we not want to use `Jenkins.getInstance().getMarkupFormatter().translate(…)` instead?

(The `*.txt` extension would not always be appropriate in that case, of course, but close enough.)

aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryTest.java
```diff
@@ -86,4 +92,55 @@ public void globalLib() throws Exception {
...
+                uvl.rebuild();
+
+                WorkflowJob p = jenkins.createProject(WorkflowJob.class, "p");
+
+                p.setDefinition(new CpsFlowDefinition(
+                        "acme.hello('Workflow');" +
+                        "acme.foo('seed');" +
+                        "echo '['+acme.bar()+']';"+
+                        "acme(1,2);"+
+                        "acme.bodyCall { title = 'yolo' }"));
```
jglick: This should be verifying sandbox mode. :bug:

jglick:
Build fails because of

    Could not find artifact com.cloudbees:groovy-cps:jar:1.5-SNAPSHOT in repo.jenkins-ci.org

I have manually `deploy`ed this snapshot, but it went to https://oss.sonatype.org/content/repositories/snapshots/ which I guess is not a configured snapshot source. Trying again with

    mvn clean deploy -DaltDeploymentRepository=maven.jenkins-ci.org::default::http://maven.jenkins-ci.org:8081/content/repositories/snapshots

aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryTest.java
kohsuke: Yes, agreed, this is indicative of something bigger deeper that's wrong.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryTest.java
```diff
@@ -86,4 +92,55 @@ public void globalLib() throws Exception {
...
+     */
+    @Test
+    public void userDefinedGlobalVariable() throws Exception {
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                File f = new File(repo.workspace, UserDefinedGlobalVariable.PREFIX+"/acme.groovy");
+                f.getParentFile().mkdirs();
+
+                FileUtils.writeStringToFile(f, StringUtils.join(Arrays.asList(
+                        "def hello(name) { node { sh \"echo Hello ${name}\" } }",
```
jtnord: just a reminder that sh in tests won't work on windows...

jglick: Indeed :bug: this should simply be

```groovy
def hello(name) {echo "Hello ${name}"}
```

jglick:
:bug: until `Closure` handling is cleared up.

(Note that `Closure` handling is apparently already wrong in `master`. But whatever the problems are, they are not manifested in any tested use cases prior to this PR; whereas the stated ultimate goal of this PR is to allow creation of a DSL based on the Groovy builder pattern, which is exactly what does not seem to work, at least if you enable the sandbox.)

jglick:
(still :bug: for `Closure`, just seems I have to add a fresh comment so the bot pesters @kohsuke not me)

kohsuke:
I fixed the handling of property access to Map in https://github.com/kohsuke/groovy-sandbox/commit/2a934790deabc128b4039eaf1d6953ba27546512 and that allows the test case in question to work correctly in the sandboxed mode.

Note that this change now requires https://github.com/jenkinsci/script-security-plugin/pull/22

kohsuke:
Removing wip label because now I consider this change to be complete.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/CpsFlowExecutionTest.java
```diff
@@ -56,24 +56,22 @@
...
-                    f.setAccessible(true);
-                    f.set(null, null);
-                }
-                MemoryAssert.assertGC(LOADER);
-            }
-        });
+    @Test public void loaderReleased() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition(CpsFlowExecutionTest.class.getName() + ".register(this)"));
+        r.assertBuildStatusSuccess(p.scheduleBuild2(0));
```
jglick: Uncompilable. Needs to be wrapped in `story.addStep`.

jglick: Seems that b282c9cec00d195582ba56b01b75f040b413be46 was a bad merge.

cps-global-lib/src/main/resources/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariable/help.jelly
jglick: (resolved)

jglick:
:bee:

cps-global-lib/README.md
```diff
@@ -59,28 +65,133 @@ Thereafter it should suffice to run:
...
 You can then call such function from your main workflow script like this:
 
-    def z = new org.foo.Zot()
-    z.checkOutFrom(repo)
+```groovy
+def z = new org.foo.Zot()
+z.checkOutFrom(repo)
+```
+
+### Defining global functions
```
daniel-beck: One of the use cases of this seem to be a friendlier syntax for workflow-compatible but not explicitly workflow-supporting plugins that would use generic steps. Maybe have a specific example of that here?

cps-global-lib/README.md
```diff
@@ -13,18 +13,24 @@ Every workflow script in your Jenkins see these shared library scripts in their
...
-             +- foo
-                 +- Bar.groovy  # for org.foo.Bar class
-
+     +- src                     # groovy source files
+     |   +- org
+     |       +- foo
+     |           +- Bar.groovy  # for org.foo.Bar class
+     +- vars
+         +- foo.groovy          # for global 'foo' variable/function 
+         +- foo.txt             # help for 'foo' variable/function
```
daniel-beck: Not documented sufficiently, should mention transform of configured markup formatter.

daniel-beck:
What happens when a file name such as `foo.bar.groovy` is used? It defines a variable that is listed but cannot actually be accessed?

daniel-beck:
:bug: until the `foo.bar` question is answered, and the documentation for the help file is improved.

kohsuke:
`foo.bar` is an invalid class name, and so you will not be able to load a script from `foo.bar.groovy`.

cps-global-lib/README.md
kohsuke: Hmm, for me the syntax simplification for the generic step is a problem on its own. You are right that this could be a workaround while we solve that, but not sure if we want to advertise it in README.

jglick: Right that is a separate (filed) issue, not really related to this PR.

jglick:
@kohsuke you _closed_ the repository for `com.cloudbees.groovy-cps:1.5` but you did not _release_ it (I did so just now), so I cannot yet push the change that switches to non-`SNAPSHOT` dependencies. Will do that before merge, once it makes its way to Central. (BTW it would be great if the CI builder could pick up `https://oss.sonatype.org/content/groups/public/` so that we do not need to wait several hours for Central propagation after releasing libraries.)

cps-global-lib/README.md
jglick: (resolved, I think)

daniel-beck:
> `foo.bar` is an invalid class name, and so you will not be able to load a script from `foo.bar.groovy`.

@kohsuke It looks like it would show up in the last though when it should not.

cps-global-lib/README.md
```diff
@@ -13,18 +13,27 @@ Every workflow script in your Jenkins see these shared library scripts in their
...
 
 The `src` directory should look like standard Java source directory structure.
-This directory is added to the classpath when executing workflows. The groovy
-source files in this directory get the same sandbox / CPS transformation
-just like your workflow scripts.
+This directory is added to the classpath when executing workflows.
+
+The `vars` directory hosts scripts that define global variables accessible from
+workflow scripts.
+The basename of each `*.groovy` file should be a Groovy (~ Java) identifier, conventionally `camelCased`.
```
daniel-beck: *must*, if you want it to work.

jglick:
> It looks like it would show up in the [list?]

It would, but oh well. Just do not do that.

daniel-beck:
Right. Good enough I suppose.

:bee:

cps-global-lib/README.md
```diff
@@ -13,18 +13,27 @@ Every workflow script in your Jenkins see these shared library scripts in their
...
+The matching `*.txt`, if present, can contain documentation, processed through the system’s configured markup formatter
```
jtnord: This seems bad - as this can be changed independently and then this would affect the help rendering here.
I guess no different to all the job descriptions though.

jglick: Yes, the same problem applies to any saved markup I am afraid.

The forced `*.txt` extension is ugly but `MarkupFormatterDescriptor` lacks an API for specifying a conventional file extension.

cps-global-lib/README.md
```diff
@@ -59,28 +68,133 @@ Thereafter it should suffice to run:
...
+windows {
+    bat "cmd /?"
+}
+```
+
+See [the closure chapter of Groovy language reference](http://www.groovy-lang.org/closures.html) for more details
+about the block syntax in Groovy.
+
+### Defining global variables
+Internally, scripts in the `vars` directory are instantiated as a singleton on-demand, when used first.
```
jtnord: OT: is there an equivalent for files not stored in the cps-global-lib?

cps-global-lib/README.md
```diff
@@ -13,18 +13,27 @@ Every workflow script in your Jenkins see these shared library scripts in their
...
+workflow scripts.
+The basename of each `*.groovy` file should be a Groovy (~ Java) identifier, conventionally `camelCased`.
+The matching `*.txt`, if present, can contain documentation, processed through the system’s configured markup formatter
+(so may really be HTML, Markdown, etc., though the `txt` extension is required).
 
-Directories other than "src" is reserved for future enhancements.
+The groovy source files in these directories get the same sandbox / CPS
+transformation just like your workflow scripts.
+
+Other directories under the root are reserved for future enhancements.
```
tfennelly: Not part of this PR, but I think the docs for this overall feature should clarify what is meant by "a 'shared library script' Git repository inside Jenkins". Specifically, I have no idea what it means when saying "inside Jenkins".

I'm also wondering why a `src` and a `vars` folder is needed in the repo i.e. why doesn't the `src` folder store everything?

jtnord:
:bee: couldn't see anything obviously wrong.

cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariable.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * Loads help from user-defined file, if available.
+     */
+    public String getHelpHtml() throws IOException {
+        File help = source(".txt");
+        if (!help.exists())     return null;
+
+        return Jenkins.getActiveInstance().getMarkupFormatter().translate(FileUtils.readFileToString(help, Charsets.UTF_8));
```
tfennelly: I assume there's a default Markup Formatter ?

cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariable.java
```diff
@@ -0,0 +1,84 @@
...
+        return new File(repo.workspace, PREFIX+"/"+ name + extension);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        UserDefinedGlobalVariable that = (UserDefinedGlobalVariable) o;
+        return name.equals(that.name);
```
tfennelly: Looks like `name` can be `null`?

cps-global-lib/README.md
tfennelly: Okay, I see later how it works (`WorkflowLibRepository`, `FileBackedHttpGitRepository` etc).

tfennelly:
:bee: 

Insofar as I get it. Only potential problem I saw was the potential NPE in `UserDefinedGlobalVariable`, which I'm betting is not really possible anyway.

cps-global-lib/README.md
jglick: No, and to my mind that is a crucial drawback (@kohsuke seems to think it is a minor enhancement). Requires some design thought. Probably a lot of the infrastructure work done here could be reused.

cps-global-lib/README.md
jglick: > I have no idea what it means when saying "inside Jenkins".

Does the _Accessing repository_ section below clear it up? Probably we need some docs reorganization (CC @estheryouhana).

> why doesn't the src folder store everything?

Because “variables” are accessed in a different way than importable classes with a package structure.

cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariable.java
jglick: Yes, it is `@Nonnull`.

cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariable.java
jglick: I hope not. The getter is `@Nonnull` at least. The constructor is called with `FilenameUtils.getBaseName`, which should return nonnull when passed nonnull, which we do (`File.getName`).

cps-global-lib/README.md
jtnord: (OT:)  at scale (and even not at scale) you likely want code review (e.g. Gerrit / GitHub PR) and may have multiple groups using a single Jenkins instance - so a global cps-global-lib becomes less useful and the desire to store these elsewhere becomes greater.

jglick: Exactly my argument. You _can_ keep your libraries in a proper SCM, and just “deploy” them as needed to Jenkins using `git push` (possibly after exporting/importing from another SCM). Awkward.

Crucially (to me; @kohsuke thinks it is minor), there is no way of specifying a library version from a single project’s `Jenkinsfile`. It is like having a `pom.xml` where every plugin is using `<version>LATEST</version>` and the build could start failing without warning.

PR #136, opened 2015-05-27
[WiP] [FIXED JENKINS-26761] - Properly handle null WorkflowBuild::checkouts

oleg-nenashev:
Fixes the case with NPE in <code>WorkflowJob:: getSCMs()</code>
This is a hotfix without tests and similar fixes for null values of other fields.

@reviewbybees

### DIFF
--- a/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -395,8 +395,10 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements B
             return Collections.emptySet();
         }
         List<SCM> scms = new LinkedList<SCM>();
-        for (WorkflowRun.SCMCheckout co : b.checkouts) {
-            scms.add(co.scm);
+        if (b.checkouts != null) {
+            for (WorkflowRun.SCMCheckout co : b.checkouts) {
+                scms.add(co.scm);
+            }
         }
         return scms;
     }
@@ -418,6 +420,9 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements B
         WorkflowRun b = getLastCompletedBuild();
         if (b == null) {
             listener.getLogger().println("no previous build to compare to");
+            return PollingResult.NO_CHANGES;
+        }
+        if (b.checkouts == null) {
             return PollingResult.NO_CHANGES;
         }
         for (WorkflowRun.SCMCheckout co : b.checkouts) {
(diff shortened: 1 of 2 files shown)


jglick:
Well this would probably address the symptom, but there is some underlying bug, perhaps serious, which is unaddressed. And this patch may make it harder to tell if you have successfully reproduced the bug. Or am I missing something?

oleg-nenashev:
The error may arise if a workflow-job run attempt fails before the initialization of checkouts list during the run procedure. 

Possible cause examples:
* No flow definition in the job
* The build gets interrupted in the early beginning (e.g. when it runs for the completion of the previous workflow)

jglick:
> No flow definition in the job

Then the right fix is just to move the initialization of the field earlier.

Anyway I am not convinced that is the root cause seen in the field, which seems to involve Jenkins restarts and thus deserialization, but it is unclear.

> when it runs for the completion of the previous workflow

Not sure I followed that.

oleg-nenashev:
> Then the right fix is just to move the initialization of the field earlier...
> Anyway I am not convinced that is the root cause seen in the field, which seems to involve Jenkins restarts and thus deserialization, but it is unclear.

Technically the field will have the null value between the WorkflowBuild constructor call and WorkflowJob::run() call, so I think the change is valid for corner-cases. I'll move the initialization to the earlier stage to address the first case.

In any case, I'm going to rework the PR in order to get rid of the direct <code>checkouts</code> field usage at all.

> Not sure I followed that.
It was a weird sentence BTW. Several methods like <code> updateSymlinks()</code> may throw interrupted exception before the field gets initialized. I'll fix it by moving the list as well.

jglick:
> I'm going to rework the PR in order to get rid of the direct `checkouts` field usage at all.

Careful, this is in the serial form of every build. Moving the definition up a few lines would help in corner cases such as a syntactically malformed flow, but for purposes of fixing the real problem I think this is a dead end.

Basically my problem here is that while it is obvious how to avoid the NPE symptom—just return an empty list—it is not obvious how people get into the state where this field is null. The descriptions in [JENKINS-26761](https://issues.jenkins-ci.org/browse/JENKINS-26761) strongly suggest that `WorkflowRun.checkouts` _is_ set initially, and also gets saved to `build.xml`, but after a restart the field is not loaded. That in turn suggests some serious problem with XStream usage, which might be affecting lots of fields, not just `checkouts`. So what I really want to see is a way to reproduce from scratch (in the restart scenario), or failing that, more thorough logging of suspicious conditions.

jglick:
There has been no progress on reproducing or analyzing the problem, so I am going to work on a PR similar to this one but with better diagnostics.

PR #137, opened 2015-05-29
[JENKINS-28649] Api call with AfterRestartTask throws ExportedBean annotation exception

christ66:
- [ ] functional test using HtmlUnit to retrieve (e.g.) `api/json`
- [X] `AfterRestartTask`
- [ ] `PlaceholderTask`

@reviewbybees

### DIFF
--- a/job/src/main/java/org/jenkinsci/plugins/workflow/job/AfterRestartTask.java
+++ b/job/src/main/java/org/jenkinsci/plugins/workflow/job/AfterRestartTask.java
@@ -33,6 +33,7 @@ import hudson.model.queue.AbstractQueueTask;
 import hudson.model.queue.CauseOfBlockage;
 import java.io.IOException;
 import org.acegisecurity.Authentication;
+import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
 
 /**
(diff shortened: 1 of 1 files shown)


job/src/main/java/org/jenkinsci/plugins/workflow/job/AfterRestartTask.java
```diff
@@ -33,11 +33,13 @@
...
+import org.kohsuke.stapler.export.ExportedBean;
 
 /**
  * Represents a {@link WorkflowRun} still running after a Jenkins restart.
  * Could be a {@code ContinuedTask}, though not really necessary since it is a {@link FlyweightTask} which would never be blocked anyway.
  */
+@ExportedBean
 class AfterRestartTask extends AbstractQueueTask implements Queue.FlyweightTask, Queue.TransientTask {
 
     private final WorkflowRun run;
```
jglick: This is what should be `@Exported` I think; this class is just a proxy for the run.

christ66: Fixed.

jglick:
Would be appropriate to have test coverage for this, but did you at least test this (as of the second commit) interactively to verify that (a) no exception is thrown, (b) usable information is available from `/queue/api/xml`?

kaihowl:
This seems to be possible fix for JENKINS-28649 (critical). What work is left to merge this in? We would very much like to be able to see the build queue in the REST API again. Thank you for working on this. :)

amuniz:
This PR has been completed in #298 

@christ66 your commits have been included in the other PR, could you close this one please?

jglick:
@amuniz there is no need to close this one. If and when #298 is merged, this one will be shown as merged automatically as well: its latest commit will be an ancestor of the target branch.

PR #139, opened 2015-06-04
[JENKINS-28756] GenericSCMStep.getApplicableDescriptors buglet

jglick:
[JENKINS-28756](https://issues.jenkins-ci.org/browse/JENKINS-28756)

@reviewbybees

### DIFF
--- a/step-api/src/main/java/org/jenkinsci/plugins/workflow/util/StaplerReferer.java
+++ b/step-api/src/main/java/org/jenkinsci/plugins/workflow/util/StaplerReferer.java
@@ -22,7 +22,7 @@
  * THE SOFTWARE.
  */
 
-package org.jenkinsci.plugins.workflow.support.steps.build;
+package org.jenkinsci.plugins.workflow.util;
 
 import hudson.model.Item;
 import java.net.URI;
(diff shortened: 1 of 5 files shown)


step-api/src/main/java/org/jenkinsci/plugins/workflow/util/StaplerReferer.java
```diff
@@ -38,8 +38,9 @@
 
 /**
  * JENKINS-19413 workaround.
+ * Will be deleted in the future.
  */
-class StaplerReferer {
+public class StaplerReferer {
```
oleg-nenashev: <code>@Restricted</code> or <code>@Deprecated</code> would be useful if you're going to delete it at some point

oleg-nenashev:
:+1:

step-api/src/main/java/org/jenkinsci/plugins/workflow/util/StaplerReferer.java
jglick: Cannot be `@Restricted` because I need to use it from sister plugins, unfortunately.

PR #140, opened 2015-06-17
[JENKINS-26107] Labeled block

jglick:
[JENKINS-26107](https://issues.jenkins-ci.org/browse/JENKINS-26107)

- [X] basic impl
- [X] test
- [ ] fix [JENKINS-26156](https://issues.jenkins-ci.org/browse/JENKINS-26156)
- [ ] ensure that failure notifications make note of the label somehow

@reviewbybees

### DIFF
--- a/basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/LabelStep.java
+++ b/basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/LabelStep.java
@@ -1,0 +1,80 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 Jesse Glick.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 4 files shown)


basic-steps/src/main/resources/org/jenkinsci/plugins/workflow/steps/LabelStep/config.jelly
```diff
@@ -0,0 +1,31 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-->
+
+<?jelly escape-by-default='true'?>
```
KostyaSha: imho better do groovy UIs instead of xml based jelly.

jglick: I do when the view needs to run any real logic. In the usual case that it is just using standard form controls, I tend to go with the more familiar option.

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/LabelStep.java
```diff
@@ -0,0 +1,80 @@
...
+            return false;
+        }
+
+        @Override public void stop(Throwable cause) throws Exception {
+            // should not be called
+        }
+
+    }
+
+    @Extension public static class DescriptorImpl extends AbstractStepDescriptorImpl {
```
amuniz: Probably it needs to extend `AbstractSynchronousNonBlockingStepExecution` rather than `AbstractStepDescriptorImpl` (which did not exist at the time of writing this code).

jglick: No, because this is an _asynchronous_ step.

jglick:
Refiled as https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/1.

PR #141, opened 2015-06-17
[JENKINS-25550] Hard kill

jglick:
[JENKINS-25550](https://issues.jenkins-ci.org/browse/JENKINS-25550)

- [X] basic implementation
- [x] remove special case handling in `DurableTaskStep` and (cf. #176) `AbstractSynchronousNonBlockingStepExecution`
- [x] make second abort attempt forcibly stop just running steps
- [x] confirmation UI
- [x] some problem with zombie builds coming back to life after restart

@reviewbybees


oleg-nenashev:
If I understand correctly, this thing does not abort the flow execution, hence there may be collisions between the hanging step and the newly launched one if they somehow use a shared resource.

jglick:
This only happens after the flow execution has been “softly” aborted twice and failed to exit, generally due to a failure of a `StepExecution.stop` to respond reliably. There are probably undesirable effects associated with a hard kill in some circumstances, but the alternative is very bad. A future revision of `CpsFlowExecution.abort` might do a better job of aborting steps within the language.

oleg-nenashev:
I have a weak :-1: regarding such solution.
What about adding "Hard kill" action with a confirmation dialog, which would allow users to abort the build if they are sure and take the responsibility for all effects, which may happen?

jglick:
That might be a better UI, assuming the option appeared only after a regular `Executor.stop` had been run at least once and (say) fifteen seconds had elapsed without the build ending on its own.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
```diff
@@ -201,4 +211,48 @@ public void contextInjectionOfSubParameters() throws Exception {
...
+        ex.interrupt();
+        JenkinsRuleExt.waitForMessage("bwahaha org.jenkinsci.plugins.workflow.steps.FlowInterruptedException #1", b);
+        ex.interrupt();
+        JenkinsRuleExt.waitForMessage("bwahaha org.jenkinsci.plugins.workflow.steps.FlowInterruptedException #2", b);
+        ex.interrupt();
+        JenkinsRuleExt.waitForMessage("Hard kill!", b);
+        JenkinsRuleExt.waitForCompletion(b);
+        r.assertBuildStatus(Result.ABORTED, b);
+    }
+    public static class Zombie extends AbstractStepImpl {
```
KostyaSha: add new line. With `@Test` without new line this block of code looks very unparsable by eyes.

jglick: Prefer to keep all the lines relating to this one test case together.

durable-task-step/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -136,15 +135,9 @@ protected DurableTaskStepDescriptor() {
...
-                FilePath workspace = getWorkspace();
-                if (workspace != null) {
-                    controller.stop(workspace, getContext().get(Launcher.class));
-                }
-            } finally {
-                if (stopAttempt++ == 1) { // second attempt
-                    getContext().onFailure(cause);
-                }
+            FilePath workspace = getWorkspace();
+            if (workspace != null) {
```
amuniz: Under what conditions could this be `false`?

jglick: That the slave is offline.

(This code was already there, of course, just reindented.)

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
```diff
@@ -65,7 +65,6 @@ public void stop(Throwable cause) throws Exception {
         if (task != null) {
             task.cancel(true);
         }
-        getContext().onFailure(cause);
```
amuniz: Where is this called now? In [doTerm](https://github.com/jenkinsci/workflow-plugin/pull/141/files#diff-16ebc446122b84ab0f87d8727295f655R287)? So any `AbstractSynchronousNonBlockingStepExecution` is now virtually not cancellable without requiring a click on `doTerm` link, right?

amuniz:
:bee:

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
jglick: The regular abort button should call `task.cancel(true)` which ought to wind up calling `Thread.interrupt`, as `SynchronousNonBlockingStepTest` still demonstrates. The `doTerm` fallback should only be necessary if that thread ignores interrupts (or catches it but then retries its task, etc.).

PR #142, opened 2015-06-17
[JENKINS-26860] WorkflowJob.concurrentBuild

jglick:
[JENKINS-26860](https://issues.jenkins-ci.org/browse/JENKINS-26860)

@reviewbybees

### DIFF
--- a/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -35,6 +35,7 @@ import hudson.model.BuildableItem;
 import hudson.model.Cause;
 import hudson.model.Computer;
 import hudson.model.Descriptor;
+import hudson.model.Executor;
 import hudson.model.Item;
 import hudson.model.ItemGroup;
 import hudson.model.Items;
@@ -44,6 +45,7 @@ import hudson.model.Label;
 import hudson.model.Node;
 import hudson.model.Queue;
 import hudson.model.ResourceList;
+import hudson.model.Run;
 import hudson.model.RunMap;
 import hudson.model.TaskListener;
 import hudson.model.TopLevelItem;
@@ -76,6 +78,7 @@ import jenkins.model.ParameterizedJobMixIn;
 import jenkins.model.lazy.LazyBuildMixIn;
 import jenkins.triggers.SCMTriggerItem;
 import jenkins.util.TimeDuration;
+import net.sf.json.JSONObject;
 import org.acegisecurity.Authentication;
 import org.jenkinsci.plugins.workflow.flow.FlowDefinition;
 import org.kohsuke.accmod.Restricted;
@@ -83,6 +86,7 @@ import org.kohsuke.accmod.restrictions.DoNotUse;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.StaplerResponse;
+import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.interceptor.RequirePOST;
 
 @SuppressWarnings({"unchecked", "rawtypes"})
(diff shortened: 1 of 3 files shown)


job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -161,10 +168,11 @@ public void setDefinition(FlowDefinition definition) {
         } else {
             quietPeriod = null;
         }
-        triggers.rebuild(req, req.getSubmittedForm(), Trigger.for_(this));
+        triggers.rebuild(req, json, Trigger.for_(this));
         for (Trigger t : triggers) {
             t.start(this, true);
         }
+        concurrentBuild = json.optBoolean("concurrentBuild") ? null : false;
```
oleg-nenashev: Javadoc states the field defaults to "true"

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -291,13 +299,39 @@ public void setQuietPeriod(Integer seconds) throws IOException {
...
 
+    @Exported
     @Override public boolean isConcurrentBuild() {
-        // TODO should this be configurable?
-        return true;
+        return !Boolean.FALSE.equals(concurrentBuild);
+    }
+    
+    public void setConcurrentBuild(boolean b) throws IOException {
+        concurrentBuild = b ? null : false;
```
oleg-nenashev: Hm?

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
oleg-nenashev: OK, I see it in the getter method

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -291,13 +299,39 @@ public void setQuietPeriod(Integer seconds) throws IOException {
...
 
     @Override public CauseOfBlockage getCauseOfBlockage() {
-        // TODO is there any legitimate reason to block a flow? maybe !isConcurrentBuild()?
+        if (isLogUpdated() && !isConcurrentBuild()) {
+            return new BecauseOfBuildInProgress(getLastBuild());
+        }
         return null;
     }
+    // TODO copied from AbstractProject
+    public static class BecauseOfBuildInProgress extends CauseOfBlockage {
```
oleg-nenashev: I suppose the class in the core is a subject for the update. <code>@Restricted</code> may be useful here

jglick: True I should file a PR in core to make it more generic.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -291,13 +299,39 @@ public void setQuietPeriod(Integer seconds) throws IOException {
     }
 
     @Override public CauseOfBlockage getCauseOfBlockage() {
-        // TODO is there any legitimate reason to block a flow? maybe !isConcurrentBuild()?
+        if (isLogUpdated() && !isConcurrentBuild()) {
```
oleg-nenashev: I still suspect <code>isLogUpdated()</code> is not reliable for parallel builds as well as any method using <code>getLastBuild()</code>. If build10 finishes before build9, the result may be errorneous

jglick: > If build10 finishes before build9

I think the idea is that this should be an impossible situation, assuming `isConcurrentBuild()` when build 10 was scheduled—it should have waited for 9 to finish.

At any rate, here I am just using the same logic as in `AbstractProject` so it is probably good enough. In the future it might be better to pull this up into `Job` or refactor into a helper.

oleg-nenashev: Yes, this core's code is very suspicious. I've investigated it several times, but didn't find obvious issues caused by it in JIRA.

oleg-nenashev:
:+1:

PR #144, opened 2015-06-25
Checking that binding['name'] syntax works for parameters

jglick:
@reviewbybees but this is for @harniman who claimed it did not work (AFAICT it does).


oleg-nenashev:
:+1:

PR #146, opened 2015-06-29
[JENKINS-26399] Standard test utils

jglick:
[JENKINS-26399](https://issues.jenkins-ci.org/browse/JENKINS-26399)

Use standard test utilities now that we can. @reviewbybees


stephenc:
:bee:

jglick:
@recena BTW technical tip: comment on lines in the PR diff tab (`…/files` URL) rather than on containing commits. This ensures that if and when the line comment is requested, GitHub will mark it as a “comment on an outdated diff”.

jglick:
@recena not sure what you mean. [This line](https://github.com/jglick/workflow-plugin/blob/a3ed2b49c91c5e62b179eb3f378068e7c1e7601e/aggregator/src/test/java/org/jenkinsci/plugins/workflow/JenkinsRuleExt.java#L64) uses the import.

recena:
@jglick What is a right way to add a comment to line of code?

recena:
@jglick Ok, after to talk with @amuniz I understand what do you want to tell me. Thanks.

recena:
@jglick Try to compile after removing the import and you'll be surprised ;)

varmenise:
:bee:

oleg-nenashev:
:bee:

PR #147, opened 2015-06-29
[JENKINS-26129] Multibranch project

jglick:
Downstream of https://github.com/jenkinsci/branch-api-plugin/pull/4.

- [X] new plugin (in the main repo for now for convenience; later could be split into an independent repo)
- [X] skeleton of project type
- [x] branch job property
- [x] run `jenkins.groovy` from SCM
- [x] allow `checkout scm` to work
- [x] disable configuration UI for branch jobs
- [x] `getSCMSourceCriteria`
- [x] smoke tests
- [x] sanity check Mercurial
- [x] sanity check Git
- [x] sanity check Subversion
- [x] reuse `ParametersDefinitionBranchProperty`
- [x] suppress `UntrustedBranchProperty`
- [x] ensure that `SCMBinder` fixes a particular commit rather than naming the branch

@reviewbybees esp. @stephenc


stephenc:
:bee: so far

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowProjectFactoryImpl.java
```diff
@@ -0,0 +1,99 @@
...
+import jenkins.branch.Branch;
+import jenkins.branch.BranchProjectFactory;
+import jenkins.branch.BranchProjectFactoryDescriptor;
+import jenkins.branch.MultiBranchProject;
+import org.jenkinsci.plugins.workflow.cps.CpsScmFlowDefinition;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+public class WorkflowProjectFactoryImpl extends BranchProjectFactory<WorkflowJob,WorkflowRun> {
```
tfennelly: To me, `WorkflowBranchProjectFactory` would seem like a better name for this class (or `WorkflowBranchProjectFactoryImpl`, but the `Impl` suffix is not needed imo) ?

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowProjectFactoryImpl.java
```diff
@@ -0,0 +1,99 @@
...
+        BranchJobProperty property = project.getProperty(BranchJobProperty.class);
+        try {
+            if (property == null) {
+                property = new BranchJobProperty();
+                setBranch(property, branch, project);
+                project.addProperty(property);
+            } else if (!property.getBranch().equals(branch)) {
+                setBranch(property, branch, project);
+                project.save();
+            }
```
tfennelly: Would it really be ok for a `BranchJobProperty` to already have a branch name set on it? And would it be okay to change it?

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowProjectFactoryImpl.java
```diff
@@ -0,0 +1,99 @@
...
+            LOGGER.log(Level.WARNING, null, x);
+        }
+        return project;
+    }
+
+    private void setBranch(BranchJobProperty property, Branch branch, WorkflowJob project) {
+        property.setBranch(branch);
+        // TODO DRY would perhaps be better for getDefinition to consult the property
+        project.setDefinition(new CpsScmFlowDefinition(branch.getScm(), SCRIPT));
+    }
```
tfennelly: NIT: Not sure I like `project.setDefinition(...)` in this setter. It seems to be overloading the method a bit.

tfennelly: How about putting `setBranch` on `BranchJobProperty`?

tfennelly:
I performed an SVN sanity check and it seems to work fine:

screenshot 2015-07-03 12 14 08

And the job config I used:

screenshot 2015-07-03 12 18 03

KostyaSha:
> And the job config I used:

Can "include branches" be expandableTextBox with new line delimiter?

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowProjectFactoryImpl.java
jglick: I am not sure about either question. I was merely implementing the required methods from the interface definition, following the lead of the `literate` plugin. Whether the interface design is right is another matter.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowProjectFactoryImpl.java
jglick: > It seems to be overloading the method a bit.

I do not think so. If you set the branch property, you must also ensure that the script definition matches it.

> How about putting `setBranch` on `BranchJobProperty`?

There is already such a method. But here we need to adjust the `project`, not just the `property`, and the `project` owns the `property` so a method on the property should not be modifying the project.

jglick:
> Can "include branches" be expandableTextBox with new line delimiter?

Might make sense, but anyway that configuration is taken from `subversion-plugin`.

jglick:
@Vlatombe IIRC you had experience in Literate and may be interested in following this.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowProjectFactoryImpl.java
jglick: Yeah, was just following a convention from `literate` IIRC. I will rename it.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowProjectFactoryImpl.java
jglick: Confirmed that `setBranch` is indeed called from several places, for whatever reason.

jglick:
Taking

> consolidate redundant `checkouts` in build metadata

off the list as this could be handled separately in [JENKINS-29326](https://issues.jenkins-ci.org/browse/JENKINS-29326).

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/scm/AbstractSampleDVCSRepoRule.java
```diff
@@ -0,0 +1,71 @@
...
+        args.addAll(Arrays.asList(cmds));
+        run(false, sampleRepo, args.toArray(new String[args.size()]));
+    }
+
+    public final String bareUrl() throws UnsupportedEncodingException {
+        return URLEncoder.encode(toString(), "UTF-8");
+    }
+
+    public final String fileUrl() {
+        return sampleRepo.toURI().toString();
```
oleg-nenashev: <code>toURI</code> throws SecurityException . probably it makes sense to handle it here

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJobProperty.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Job property which can specially affect behavior of the project.
+ */
+public abstract class WorkflowJobProperty extends JobProperty<WorkflowJob> {
+
+    /**
+     * Allows a property to control permissions related to the job.
+     */
+    public @Nonnull ACL decorateACL(@Nonnull ACL acl) {
```
oleg-nenashev: Makes sense to add it to Jenkins core as well

oleg-nenashev:
My head is spinning, but the most of the code looks good to me. :bee: 
Just for the future, it would be better to submit massive test framework reworks in separate pull reuqests

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/scm/AbstractSampleDVCSRepoRule.java
jglick: Only if a `SecurityManager` were present, which one will not be. Anyway even if it were thrown, the test would just fail and someone would fix it.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJobProperty.java
jglick: Perhaps. I am not completely comfortable with doing this via a `JobProperty`-based API, just following a convention based on behavioral hooks in `JobProperty` itself, not to mention `FolderProperty`. The more flexible approach is to use an `@Extension` which can choose to look for a particular `JobProperty` or use any other criterion. I am open to switching to that idiom here. (Though it needs to be verified that there is minimal overhead in iterating the extensions, so as not to adversely impact jobs not using the property. I _think_ the extension lists get cached effectively.)

For the future, if we are talking about things to be baked into core, I think a more general set of APIs is needed to allow external customization of ACL decisions. CJP-2326 tracks some thoughts on this, which if accepted would need to be written up into a more concrete proposal in JENKINS.

jglick:
> it would be better to submit massive test framework reworks in separate pull requests

Yes, it might have been better to do that refactoring in an upstream PR. I just did not realize in advance I was going need so much refactoring of existing tests.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
```diff
@@ -0,0 +1,75 @@
...
+public class BranchJobProperty extends WorkflowJobProperty {
+
+    private /*@Nonnull once initialized */ Branch branch;
+
+    public synchronized Branch getBranch() {
+        return branch;
+    }
+
+    synchronized void setBranch(@Nonnull Branch branch) {
+        branch.getClass();
```
amuniz: Needed?

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
```diff
@@ -0,0 +1,91 @@
...
+        return project.getProperty(BranchJobProperty.class).getBranch();
+    }
+
+    @Override public WorkflowJob setBranch(WorkflowJob project, Branch branch) {
+        BranchJobProperty property = project.getProperty(BranchJobProperty.class);
+        try {
+            if (property == null) {
+                property = new BranchJobProperty();
+                property.setBranch(branch);
+                project.addProperty(property);
```
amuniz: Isn't ```project.save()``` required here?

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
jglick: Enforces `@Nonnull`. Just copied from the convention used in `literate`.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactory.java
jglick: [no](https://github.com/jenkinsci/jenkins/blob/e55f356ed0132b966200e5266a5e382a1c019388/core/src/main/java/hudson/model/Job.java#L512)

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
stephenc: My preferred style is to ensure that non-null parameters actually are... as other users may not be running findbugs and if they pass a null then all manner of issues can arise.

Some time back, the fastest way to check for a null was to just call `getClass()`. Current benchmarking of the JVM shows that this is no longer the case as of at least more recent builds of JVM 7. The only reason to use it though is on Java 1.6 codebases it does not leak either a branch point (which `if (x==null) throw NPE;` would do) or require Java 1.7 where `Objects.requireNonNull(x)` is the fastest.

stephenc: If you want to leave a `x.getClass(); // throws NPE if null TODO replace with Objects.requireNonNull(x) once Java 1.7+` comment that is fine by me

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        @Override public TopLevelItem newInstance(ItemGroup parent, String name) {
+            return new WorkflowMultiBranchProject(parent, name);
+        }
+
+        @Override public List<SCMDescriptor<?>> getSCMDescriptors() {
+            // TODO this is a problem. We need to select only those compatible with Workflow.
+            // Yet SCMDescriptor.isApplicable requires an actual Job, whereas we can only pass WorkflowJob.class!
+            // Can we create a dummy WorkflowJob instance?
```
amuniz: I think it would be ok to pass ```new WorkflowJob()``` if the instance doesn't break any ```@NotNull``` restriction. But we only have ```new WorkflowJob(ItemGroup, String)```... so could be Jenkins.getInstance() the item group?

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
```diff
@@ -0,0 +1,192 @@
...
+                WorkflowRun b1 = p.getLastBuild();
+                assertNotNull(b1);
+                assertEquals(1, b1.getNumber());
+                story.j.assertLogContains("initial content", story.j.waitForCompletion(b1));
+            }
+        });
+    }
+
+    @Test public void exactRevisionGit() throws Exception {
+        story.addStep(new Statement() {
```
amuniz: Since there is only one step, is needed a ```RestartableJenkinsRule``` here?

pom.xml
```diff
@@ -88,7 +89,11 @@
                 <artifactId>subversion</artifactId>
                 <version>2.5</version>
             </dependency>
-
+            <dependency>
+                <groupId>org.jenkins-ci.plugins</groupId>
+                <artifactId>mercurial</artifactId>
+                <version>1.54</version>
```
amuniz: Should it be scoped to ```test```?

amuniz:
:bee:

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
jglick: In my experience creating phony `Item`s is a disaster. Probably `SCMDescriptor` in core would require an API change.

Anyway `MultiBranchProjectDescriptor.getSCMDescriptors` appears to be unused. It claims to be used by `SingleSCMSource` (a single fixed branch—not the normal use case), but all I can find is [this](https://github.com/jenkinsci/scm-api-plugin/blob/dfd212743862b9d1f7a12255723f8358bbb1bf78/src/main/java/jenkins/scm/impl/SingleSCMSource.java#L176-204), which would ultimately call `MultiBranchProjectDescriptor.isApplicable(Descriptor)`, which is not overridden. Presumably that ought to be calling `getSCMDescriptors`, right @stephenc?

pom.xml
jglick: No, this is just `dependencyManagement`; the actual `dependency`s are `scope`d to `test`.

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
jglick: The `scmPickle` test uses two steps.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/BranchJobProperty.java
jglick: Could probably be made into an automated refactoring for the future, since I see this idiom a lot of places.

pom.xml
amuniz: Just below, ```mockito-all``` has scope ```test```, it should be removed then.

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @ClassRule public static BuildWatcher buildWatcher = new BuildWatcher();
+    @Rule public JenkinsRule r = new JenkinsRule();
+    @Rule public GitSampleRepoRule sampleRepo = new GitSampleRepoRule();
+
+    @Test public void basicBranches() throws Exception {
+        sampleRepo.init();
+        sampleRepo.write("jenkins.groovy", "node {checkout scm; echo readFile('file')}");
+        sampleRepo.write("file", "initial content");
+        sampleRepo.git("add", "jenkins.groovy");
```
tfennelly: Looks like you are calling `sampleRepo.write` for 2 files above, but only adding 1 of them to the repo ("jenkins.groovy"). Was that intentional?

pom.xml
jglick: Maybe so, but not on topic for this PR.

tfennelly:
General point ...

The following code pattern appears a number of times:
```
sampleRepo.write("jenkins.groovy", "<script-code>");
sampleRepo.write("file", "initial content");
sampleRepo.git("add", "jenkins.groovy");
sampleRepo.git("commit", "--all", "--message=flow");
```

How about a repo utility function that dries this down to reduce the noise in the test (assuming it's only test setup)? E.g.:
```
sampleRepo.commit("jenkins.groovy", "<script-code>");
```

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectTest.java
jglick: Yes, because `file` was already `add`ed by `init()`, and `--all` (below) adds it to the index too.

jglick:
> How about a repo utility function that dries this down to reduce the noise in the test

Would be a nice refactoring. I am not going to bother with it now.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+        WorkflowJob job = ((WorkflowRun) exec).getParent();
+        BranchJobProperty property = job.getProperty(BranchJobProperty.class);
+        if (property == null) {
+            throw new IllegalStateException("inappropriate context");
+        }
+        Branch branch = property.getBranch();
+        ItemGroup<?> parent = job.getParent();
+        if (!(parent instanceof WorkflowMultiBranchProject)) {
+            throw new IllegalStateException("inappropriate context");
```
tfennelly: NIT: How cryptic will the above `IllegalStateException("inappropriate context")` exceptions be if they appear to the user? Or will the user ever see them?

tfennelly:
:bee: 

I just reviewed the commits since the last time I looked at this and I did it kinda quickly.

`GitSampleRepoRule` is great .... I like how it cleans things up. Nit: I know it's a `@Rule`, but does the name need to include that? Plain `GitSampleRepo` seems cleaner to me.

Nit: I can follow the tests and what they are doing, but I sometimes find them a bit hard to read because there's no comments to tell the reader what the test is doing, expecting etc. E.g. WorkflowMultiBranchProjectTest could be easier to read.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
jglick: It is an `IllegalStateException` because the user should never see them unless I messed up somehow (or they did something weird with the script console, etc.).

jglick:
> I know it's a `@Rule`, but does the name need to include that? Plain `GitSampleRepo` seems cleaner to me.

No opinion, but consider `JenkinsRule`.

Tests could probably be more thoroughly commented. I suppose this is something else that could be improved in the future, though.

PR #149, opened 2015-07-03
[FIXED JENKINS-29221] WorkflowRun assigned FlowExecution when execution start fails

tfennelly:
[See JENKINS-29221](https://issues.jenkins-ci.org/browse/JENKINS-29221).

@reviewbybees

### DIFF
--- a/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -180,13 +180,18 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements Q
             }
             checkouts = new LinkedList<SCMCheckout>();
             Owner owner = new Owner(this);
-            execution = definition.create(owner, listener, getAllActions());
+            
+            FlowExecution newExecution = definition.create(owner, listener, getAllActions());
             FlowExecutionList.get().register(owner);
-            execution.addListener(new GraphL());
+            newExecution.addListener(new GraphL());
             completed = new AtomicBoolean();
             logsToCopy = new LinkedHashMap<String,Long>();
-            execution.start();
-            executionPromise.set(execution);
+            newExecution.start();
+            
+            // It's only okay to initialise the global once the new FlowExecution
+            // has successfully started.
+            execution = newExecution;
+            executionPromise.set(newExecution);
         } catch (Throwable x) {
             if (listener == null) {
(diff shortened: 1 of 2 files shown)


job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -180,13 +180,18 @@ public WorkflowRun(WorkflowJob job, File dir) throws IOException {
...
             completed = new AtomicBoolean();
             logsToCopy = new LinkedHashMap<String,Long>();
-            execution.start();
-            executionPromise.set(execution);
+            newExecution.start();
+            
+            // It's only okay to initialise the global once the new FlowExecution
+            // has successfully started.
+            execution = newExecution;
+            executionPromise.set(newExecution);
```
tfennelly: @kohsuke in the situation where the `execution` fails to start, does the instance need to be cleaned up in some way?

tfennelly:
@kohsuke KK, would you be able to review this please?

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
kohsuke: No, no such cleanup needed.

kohsuke:
:bee: this change is good from my PoV.

tfennelly:
@reviewbybees done

tfennelly:
@KostyaSha Yeah... I'm just following the pattern already established in that code (see other `@Test` methods). While I too (like you) don't like the "all on one line" style, I don't agree with breaking with what's already there and everyone doing what they think is "right".

KostyaSha:
@tfennelly seems one line annotation is only @jglick coding style. Just want to see some consistency... :(

tfennelly:
@KostyaSha right, I'm not sure of anyone else doing it. I don't love it, but it doesn't bother me that much either. I've seen lots of different code style conventions used in Jenkins code (core and in plugins), so I've adopted the policy of following/respecting whatever the "local" convention is as best I can.

jglick:
FWIW my convention is to put only determiner annotations like `@Test` (where in English you could put the word “the” in front of the annotation: “the test `worksOnMyMachine`”—there can be only one such) on the same line, but to put descriptive annotations like `@Issue` on their own lines.

jglick:
Do we not need to call some variant of `finish` if creating the execution fails? To fire build finish events, etc. That is what I expected the fix of this issue to involve.

PR #150, opened 2015-07-08
[FIXED JENKINS-29289] InputStep doesn't support File Parameters

tfennelly:
[See JENKINS-29289](https://issues.jenkins-ci.org/browse/JENKINS-29289)

- [x] Support `multipart/form-data`
- [x] `InputStepExecution` to copy File Param file to Workspace instead of build dir
- [x] Add a test

@reviewbybees (esp @kohsuke)

There's an ugly hack in order to get at the `parsedFormData` in the stapler `RequestImpl`. `RequestImpl.getFileItem` filters the items on behalf of the caller, which sucks imo.

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepFileParameterTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepFileParameterTest.java
@@ -1,0 +1,149 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.jenkinsci.plugins.workflow.steps.input;
+
+import hudson.model.queue.QueueTaskFuture;
+import org.apache.commons.httpclient.HttpClient;
+import org.apache.commons.httpclient.methods.PostMethod;
+import org.apache.commons.httpclient.methods.multipart.FilePart;
+import org.apache.commons.httpclient.methods.multipart.MultipartRequestEntity;
+import org.apache.commons.httpclient.methods.multipart.Part;
+import org.apache.commons.httpclient.methods.multipart.StringPart;
+import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
+import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.jenkinsci.plugins.workflow.support.steps.input.InputAction;
+import org.jenkinsci.plugins.workflow.support.steps.input.InputStepExecution;
+import org.junit.Assert;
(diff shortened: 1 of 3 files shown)


support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/InputStepExecution.java
```diff
@@ -296,4 +313,62 @@ private Object convert(String name, ParameterValue v) throws IOException, Interr
...
+        
+        if (item != null) {
+            return item;
+        }
+        if (!(request instanceof RequestImpl)) {
+            return null;
+        }
+
+        try {
+            Field parsedFormDataField = request.getClass().getDeclaredField("parsedFormData");
```
jglick: Prefer use of `RequestImpl.class`.

tfennelly: @jglick Right .. I added the earlier explicit check for `RequestImpl` after that `request.getClass()` was added ... should have changed that at the same time.

Better again though would be to not have to do shit like this at all ;) I doubt we ever do hacks like this, or is it acceptable as a stopgap while we wait for changes to come through from Stapler i.e. for me to make a change there to make it possible to get the proper `FileItem` from `StaplerRequest` (not done, but will do if we think this is how to go).

jglick:
Looks like this approach should work, though I had also considered a very different approach in [JENKINS-27413](https://issues.jenkins-ci.org/browse/JENKINS-27413). No strong opinion at the moment.

tfennelly:
@jglick so is that a bee then? We can always follow up and improve the impl with whatever you had in mind, no? I just added a test this morning.

@kohsuke KK, would you mind reviewing and giving go/no-go please?

amuniz:
:bee:

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepFileParameterTest.java
```diff
@@ -0,0 +1,149 @@
...
+                "\n" +
+                "    // File should now be in the workspace... \n" +
+                "\n" +
+                "    if (!file.exists()) {\n" +
+                "        throw new IllegalStateException(\"File should exist\");\n" +
+                "    } else {\n" +
+                "        echo \"${file} is in workspace. OK!!\"\n" +
+                "    }\n" +
+                "\n" +
+                "}"));
```
amuniz: I think having the groovy script in a test resource would be clearer.

tfennelly: Hmmmm.... do you really think so? I prefer being able to see it all in one place if possible.

amuniz: I think inlining works for small scripts, but this one is difficult to read with a lot of scape and new line chars. It's only my perception, I don't have strong feelings on it.

oleg-nenashev:
:bee:

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/InputStepExecution.java
```diff
@@ -287,13 +302,77 @@ private Object parseValue(StaplerRequest request) throws ServletException, IOExc
...
+            }
+        }
+
+        return false;
+    }    
+    
+    // TODO: total hack to gain access to all parsedFormData from Stapler RequestImpl
+    // It's a pity that RequestImpl makes the decision it does wrt isFormField FileItem filtering... why not let the 
+    // caller do that kind of filtering if they need it !!    
+    private static FileItem getFileItem(StaplerRequest request, String name) throws ServletException, IOException {
```
kohsuke: My apologies, I originally didin't understand where @tfennelly wanted me to comment on. I think you were interested in my providing the background on why `StaplerRequest.getFileItem` checks `item.isFormField()` and returns null.

The idea behind the code is that it didn't make sense in my mind that `FileItem` represents something other than a file, like a value from `<input type=text>` field. The way I see it, `StaplerRequest.getParameter` should return such a string value --- IOW I wanted to 'enhance' `StaplerRequest` to natively understand multipart form submissions, so that the applications don't need to differentiate those two submission content types. I see that `RequestImpl.getParameter` actually doesn't do any of that, so you are caught between a rock and a hard place. My apologies.

I'll improve Stapler so that `getParameter` family of methods do the right thing. In the mean time, you can either keep this code as-is, or you can use `StaplerRequest.getSubmittedForm()` to parse a submission into JSON, which provides a consistent JSON object regardless of whether the submission is multipart/form-data or application/x-www-form-urlencoded.

kohsuke: I filed https://github.com/stapler/stapler/issues/52 to keep track of this task in stapler.

jglick:
> so is that a bee then?

It was not; I still have not had time to think about this deeply enough to really comment.

> We can always follow up and improve the impl

No, we cannot, it is not just implementation, it is a definition of how parameters work, and any change would break scripts using the initial implementation.

jglick:
To be clear, my hope was to avoid any special-case handling of `FileParameter` in `workflow-plugin`, so that all the desired behavior would come from the parameter itself, as happens for freestyle projects, allowing any other plugin-provided parameters to be equally expressive. Obviously that would require core API changes. Perhaps it would wind up being so awkward it is not worth it—needs some investigation.

jglick:
Also I would suggest implementing `FileParameter` support for build parameters at the same time, as requested in JENKINS-27413, so that we can see how a particular design choice holds up.

tfennelly:
@jglick I'll close this PR seeing as it's nowhere near what you want.

jglick:
Well, TBD. As I said, it might be the only approach that actually works.

PR #152, opened 2015-07-13
[FIXED JENKINS-29396] - Support the customizable BUILD_NOW text

oleg-nenashev:
JENKINS-26147 blocks a solution in Jenkins core, so we need an alternative approach for workflow. The best way would be to have an internal implementation of WorkflowJob.BUILD_NOW_TEXT.

https://issues.jenkins-ci.org/browse/JENKINS-29396

@reviewbybees

### DIFF
--- a/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -30,6 +30,7 @@ import hudson.FilePath;
 import hudson.Launcher;
 import hudson.init.InitMilestone;
 import hudson.init.Initializer;
+import hudson.model.AbstractProject;
 import hudson.model.Action;
 import hudson.model.BuildableItem;
 import hudson.model.Cause;
@@ -100,6 +101,8 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements B
     private transient LazyBuildMixIn<WorkflowJob,WorkflowRun> buildMixIn;
     /** defaults to true */
     private @CheckForNull Boolean concurrentBuild;
+    
+    public static final AlternativeUiTextProvider.Message<WorkflowJob> BUILD_NOW_TEXT = new AlternativeUiTextProvider.Message<WorkflowJob>();
 
     public WorkflowJob(ItemGroup parent, String name) {
         super(parent, name);
@@ -285,7 +288,8 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements B
     }
 
     @Override public String getBuildNowText() {
-        return createParameterizedJobMixIn().getBuildNowText();
+        //TODO Get rid of it when JENKINS-26147 gets fixed
+        return AlternativeUiTextProvider.get(BUILD_NOW_TEXT, this, createParameterizedJobMixIn().getBuildNowText());
     }
 
     @Override public boolean isBuildBlocked() {


oleg-nenashev:
Retriggering the build to see findbugs

tfennelly:
:bee: 

`AlternativeUiTextProvider` etc makes me feel a bit nauseous though.

jglick:
:bug: pending a new API from [JENKINS-26147](https://issues.jenkins-ci.org/browse/JENKINS-26147) I would rather just use the existing text from `AbstractProject`, which would allow providers such as the one in `cloudbees-template` to work unmodified.

oleg-nenashev:
@jglick 
See my comments in the issue. Text from <code>AbstractProject</code> won't work, because <code>AlternativeUiTextProvider.Message<XXX>()</code> is templatized => we cannot pass the instance of WorkflowJob to AbstractProject

jglick:
In fact you can:

```diff
diff --git a/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java b/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
index 5680c43..c61a33c 100644
--- a/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
+++ b/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
@@ -30,6 +30,7 @@ import hudson.FilePath;
 import hudson.Launcher;
 import hudson.init.InitMilestone;
 import hudson.init.Initializer;
+import hudson.model.AbstractProject;
 import hudson.model.Action;
 import hudson.model.BuildableItem;
 import hudson.model.Cause;
@@ -285,7 +286,8 @@ public final class WorkflowJob extends Job<WorkflowJob,WorkflowRun> implements B
     }
 
     @Override public String getBuildNowText() {
-        return createParameterizedJobMixIn().getBuildNowText();
+        AlternativeUiTextProvider.Message text = AbstractProject.BUILD_NOW_TEXT;
+        return AlternativeUiTextProvider.get(text, this, createParameterizedJobMixIn().getBuildNowText());
     }
 
     @Override public boolean isBuildBlocked() {
```

Not pretty but good enough until the core API change is available.

oleg-nenashev:
@jglick 
If I remember the code correctly, <code>this</code> (second method parameter) must be an instance of <code>AbstractProject</code> to get this approach working

jglick:
> `this` (second method parameter) must be an instance of `AbstractProject`

Hence the use of the rawtype. What happens at runtime is another matter. `AbstractProject.BUILD_NOW_TEXT.cast` will succeed, but the caller’s bytecode may then be casting to `AbstractProject`.

If so, I would just say this PR should be closed and JENKINS-29396 not fixed. Wait for the core API to be available, and when it is, implement it in Workflow. Adding this new API here when we know we are going to delete it is not a good idea.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -100,6 +101,8 @@
     private transient LazyBuildMixIn<WorkflowJob,WorkflowRun> buildMixIn;
     /** defaults to true */
     private @CheckForNull Boolean concurrentBuild;
+    
+    public static final AlternativeUiTextProvider.Message<WorkflowJob> BUILD_NOW_TEXT = new AlternativeUiTextProvider.Message<WorkflowJob>();
```
KostyaSha: Why not `RUN_NOW_TEXT`?

jglick: Because it is generalizing a field already named `BUILD_NOW_TEXT`, which is the text normally called (in English) _Build Now_.

jglick:
Better to just implement the API behind JENKINS-26147 in core and use it when available.

PR #153, opened 2015-07-14
Enable FindBugs in workflow parent pom

oleg-nenashev:
@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,7 +101,7 @@
         <dependency>
             <groupId>com.google.code.findbugs</groupId>
             <artifactId>annotations</artifactId>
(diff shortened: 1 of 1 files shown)


stephenc:
:bee:

jtnord:
:bee: 
although I would generally prefer something not related to tests to not be related to the `skipTests` flag this was already existing.

jglick:
> I would generally prefer something not related to tests to not be related to the `skipTests` flag

Wanted to make sure that a

    mvn -DskipTests clean install

would do the minimum work necessary to create snapshot binaries and not run testing-like activity, into which category FindBugs falls.

PR #154, opened 2015-07-16
Wrap PrintWriter around StringWriter because print stack trace does n…

ganthore:
…ot accept StringWriter as a method.

### DIFF
--- a/basic-steps/CORE-STEPS.md
+++ b/basic-steps/CORE-STEPS.md
@@ -50,7 +50,7 @@ node {
         mail subject: 'all well', to: 'admin@somewhere', body: 'All well.'
     } catch (e) {
         def w = new StringWriter()
-        e.printStackTrace(w)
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
PrintWriter does not specify the encoding. It may be a bad approach to recommend it in docs. http://stackoverflow.com/questions/2822005/java-difference-between-printstream-and-printwriter

stephenc:
@oleg-nenashev given that this is a `StringWriter` that is being wrapped, you do not need to specify encoding. You only need to specify the encoding if you are pushing all the way down to an encoded `byte[]` but as this is going to a `String` then `PrintWriter` is actually perfectly acceptable.

oleg-nenashev:
OK, then :bee:

jglick:
:bee: though it occurs to me that `hudson.Functions.printThrowable` could be used instead.

PR #156, opened 2015-07-21
[JENKINS-29534] WorkflowJob now extends AbstractProject instead of Job

aduprat:


jglick:
No, extending `Job` was a fundamental design decision.

PR #157, opened 2015-07-22
[JENKINS-29571] Handle build records after rename

jglick:
[JENKINS-29571](https://issues.jenkins-ci.org/browse/JENKINS-29571)

If the existence of this bug a year in is not sufficient evidence of the need for QA, I am not sure what is…@reviewbybees but esp. @kohsuke whose exotic `unmarshal` implementation this was.

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
@@ -56,7 +56,6 @@ import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
 import static org.junit.Assert.*;
 import org.junit.ClassRule;
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.jvnet.hudson.test.BuildWatcher;
@@ -231,7 +230,6 @@ public class WorkflowRunTest {
         r.assertBuildStatusSuccess(r.waitForCompletion(b));
     }
 
-    @Ignore("TODO as per JENKINS-27704, CpsFlowExecution.owner should be transient")
     @Issue("JENKINS-29571")
     @Test public void buildRecordAfterRename() throws Exception {
         {
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
@@ -36,19 +36,27 @@ import hudson.model.queue.QueueTaskFuture;
 import hudson.security.ACL;
 import hudson.security.GlobalMatrixAuthorizationStrategy;
 import hudson.security.Permission;
+import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import jenkins.model.Jenkins;
+import org.apache.commons.io.FileUtils;
 import org.jenkinsci.plugins.scriptsecurity.scripts.ScriptApproval;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
+import org.jenkinsci.plugins.workflow.cps.nodes.StepNode;
+import org.jenkinsci.plugins.workflow.flow.FlowExecution;
+import org.jenkinsci.plugins.workflow.graph.FlowGraphWalker;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
-
 import static org.junit.Assert.*;
 import org.junit.ClassRule;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.jvnet.hudson.test.BuildWatcher;
@@ -221,6 +229,37 @@ public class WorkflowRunTest {
         WorkflowRun b = p.getLastBuild();
         assertNotNull(b);
         r.assertBuildStatusSuccess(r.waitForCompletion(b));
+    }
+
+    @Ignore("TODO as per JENKINS-27704, CpsFlowExecution.owner should be transient")
+    @Issue("JENKINS-29571")
+    @Test public void buildRecordAfterRename() throws Exception {
+        {
+            WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p1");
+            p.setDefinition(new CpsFlowDefinition("echo 'hello world'"));
+            r.assertBuildStatusSuccess(p.scheduleBuild2(0));
+            p.renameTo("p2");
+        }
+        r.jenkins.reload();
+        {
+            WorkflowJob p = r.jenkins.getItemByFullName("p2", WorkflowJob.class);
+            assertNotNull(p);
+            WorkflowRun b = p.getLastBuild();
+            assertNotNull(b);
+            System.out.println(FileUtils.readFileToString(new File(b.getRootDir(), "build.xml")));
+            r.assertLogContains("hello world", b);
+            FlowExecution exec = b.getExecution();
+            assertNotNull(exec);
+            FlowGraphWalker w = new FlowGraphWalker(exec);
+            List<String> steps = new ArrayList<String>();
+            FlowNode n;
+            while ((n = w.next()) != null) {
+                if (n instanceof StepNode) {
+                    steps.add(((StepNode) n).getDescriptor().getFunctionName());
+                }
+            }
+            assertEquals("[echo]", steps.toString());
+        }
     }
 
 }
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
@@ -48,7 +48,6 @@ import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
 
 import static org.junit.Assert.*;
-import org.junit.Before;
 import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
@@ -62,15 +61,8 @@ public class WorkflowRunTest {
     @ClassRule public static BuildWatcher buildWatcher = new BuildWatcher();
     @Rule public JenkinsRule r = new JenkinsRule();
 
-    WorkflowJob p;
-
-    @Before
-    public void setUp() throws Exception {
-        r.jenkins.getInjector().injectMembers(this);
-    }
-
     @Test public void basics() throws Exception {
-        p = r.jenkins.createProject(WorkflowJob.class, "p");
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
         p.setDefinition(new CpsFlowDefinition("println('hello')"));
         WorkflowRun b1 = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
         assertFalse(b1.isBuilding());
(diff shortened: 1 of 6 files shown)


amuniz:
Basically you are excluding ```owner``` from serialization, so it was not needed since the beginning, right?

jglick:
> you are excluding `owner` from serialization

Right. It violated the DRY principle: `$JENKINS_HOME/jobs/foo/builds/17/build.xml` should not have any nested objects specifying that they are owned by foo 17.

(There could be other bugs involving builds running _during_ a rename, since it is common for serialized program state to include a `FlowExecutionOwner`. However it is generally a bad idea to rename or move a job while it is being built. `cloudbees-folders-plus` makes _Move_ wait until all running builds are finished. Not sure if there is a similar sanity check out there for renaming. At any rate, it would be possible for an `ItemListener.onLocationChanged` to update all known matching `WorkflowRun.Owner.job` fields so the next save uses the new `fullName`. That would also have fixed this bug, but I wanted to get rid of the redundant information anyway. Another option would be to define a special `Pickle` for `Owner` that has no state of its own, though the `Pickle` interface does not currently let the build record inject any context.)

> so it was not needed since the beginning, right?

Right, not intrinsically, but the old code expected it to be restored from XML. This was tricky because some parts of the `unmarshal` sequence (XStream custom deserialization) depended on the `storage` field having been initialized, which in turn required `owner` since the flow node storage is in a disk location under the build directory. The fix simplifies `unmarshal` to purely work with in-memory data structures, and defers restoration of actual `FlowNode`s until `onLoad` is called.

amuniz:
Perfect. Thanks for the explanation.
I saw you fixed findbugs warnings (we have some QA now :smile:), so here is my :bee:

aggregator/src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
```diff
@@ -232,4 +230,34 @@ public void failedToStartRun() throws Exception {
         r.assertBuildStatusSuccess(r.waitForCompletion(b));
     }
 
+    @Issue("JENKINS-29571")
+    @Test public void buildRecordAfterRename() throws Exception {
```
KostyaSha: @jglick from your described logic it should be `public void @Test`, isn't it?

aggregator/src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
```diff
@@ -162,6 +161,10 @@ public void iconColor() throws Exception {
         assertFalse(b2.hasntStartedYet());
         assertColor(b2, BallColor.BLUE);
     }
+    private void assertColor(WorkflowRun b, BallColor color) throws IOException {
```
KostyaSha: new line before?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -225,10 +225,14 @@
...
     @GuardedBy("this")
-    /*package*/ final Stack<BlockStartNode> startNodes = new Stack<BlockStartNode>();
+    /*package*/ /* almost final*/ Stack<BlockStartNode> startNodes = new Stack<BlockStartNode>();
+    @SuppressFBWarnings({"IS_FIELD_NOT_GUARDED", "IS2_INCONSISTENT_SYNC"}) // irrelevant here
+    private transient List<String> startNodesSerial; // used only between unmarshal and onLoad
 
     @GuardedBy("this")
-    private final NavigableMap<Integer,FlowHead> heads = new TreeMap();
+    private /* almost final*/ NavigableMap<Integer,FlowHead> heads = new TreeMap<Integer,FlowHead>();
+    @SuppressFBWarnings({"IS_FIELD_NOT_GUARDED", "IS2_INCONSISTENT_SYNC"}) // irrelevant here
```
oleg-nenashev: It would be better to use the <code>justification</code> field. In such case it's possible to plot a report of suppressed warnings in FindBugs. This functionality may be useful IMO

cps/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
```diff
@@ -46,8 +47,9 @@
 
         assertFalse("Expected the execution to be suspended but it has completed", exec.isComplete());
 
+        FlowExecutionOwner owner = exec.getOwner();
         exec = roundtripXStream(exec);    // poor man's simulation of Jenkins restart
```
oleg-nenashev: NIT: now we have a rule allowing to implement a more fair restart

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -419,7 +419,14 @@ private String key() {
             System.err.printf("loading %s @%h%n", this, this);
         }
         if (execution != null) {
-            execution.onLoad();
+            try {
+                execution.onLoad(new Owner(this));
+            } catch (IOException x) {
+                LOGGER.log(Level.WARNING, null, x);
```
oleg-nenashev: Some text would be useful

aggregator/src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
stephenc: @jglick The following is a tad rant. It is a personal opinion of mine, feel free to ignore if you want.

I really wish you could stop with your crazy fetish of putting annotations on the same line as the method declaration.

The annotations target the method, that we have put semantic value on some annotations as somehow applying to the return type is not quite correct in my view.

When you have a `public @CheckForNull Object getSomething()` method the return type is not `@CheckForNull Object`...

I do not see you annotating your local variables

```
  public void doMagic() {
    ...
    @CheckForNull Object s = getSomething();
    ...
    if (s != null) {
      @Nonnull Object sth = s;
      ...
    }
    ...
    if (s != null) {
      @Nonnull Object sth = s;
      ...
    }
    ...
  }
```

A method that is annotated `@CheckForNull` is not semantically the same as a method returning `Optional<T>`.

To my thinking, you are annotating the *method* as potentially returning a `null` the type of the return value doesn't magically morph from a `@CheckForNull T` into a `@Nonnull T` and in fact, as soon as you capture that value into a local variable without a `@CheckForNull` or `@Nonnull` annotation what you have effectively done is "convert" the value into a `@Nullable`... only that findbugs does call path tracing for you allows findbugs to upscale that into either a  `@CheckForNull` or `@Nonnull` based on the properties of the *methods* you have called.

Is this another NetBeans has support for differentiating `@TypeQualifier` annotated annotations and formatting them next to the type while retaining the non `@TypeQualifier` annotations on the preceding lines thingy much like the java 8 for tests thingy?

We will ignore for now that JSR 305 is completely stalled, that there has not been a formal release (even a draft) of the annotations from that JSR and that you are not supposed to populate the `javax.*` namespace outside of the JSR process... and we will pretend that `@TypeQualifier` will actually come out of that JSR is it ever resurrects itself (That JSR and Jenkins having chosen to use it's annotations - incorrectly in my view - is a whole different discussion)

I have checked IntelliJ's code formatting options:
screen shot 2015-07-24 at 10 42 25

No option to put `@TypeQualifier` annotations in between the `(public|protected|private)\s(static\s)?(final\s)?` and the return type. The IDE forced that the type annotations should be before the `(public|protected|private)\s(static\s)?(final\s)?` and my choice is when to line wrap.

Eclipse does not support your code style either:

screen shot 2015-07-24 at 10 45 40

In fact Eclipse is more restrictive. You just get the choice of a new line or not.

After sullying my machine with a temporary Eclipse installation, I said what the heck and installed NetBeans... I could not find any support for your code style:

screen shot 2015-07-24 at 10 52 46

So it would seem that the IDEs have the following options:

* Annotations on the same line *but* before the `(public|protected|private)` (All IDEs)
* Annotations on the same line *but* before the `(public|protected|private)` and wrap if long (IntelliJ & NetBeans)
* Annotations on the same line *but* before the `(public|protected|private)` and chop down if long (IntelliJ)
* Annotations, each on their own line, before the line with `(public|protected|private)` (All IDEs)

So none of the IDEs - from what I can see - support your code style. 

That to me is rather annoying. I like when there is a reasonable code style to keep with. 

Historically the Jenkins code style has tended to be "The default code style that ships with IntelliJ"... largely as a result of KK's preferred IDE. Though I appreciate that specific code style is not very friendly to users of other IDEs. 

As the community has grown, each plugin maintainer is free to define their own code style. 

Issues of code style can affect the ability of others to contribute code.

If a plugin maintainer does not provide code style guidelines, it raises the bar for contribution.

To my way of thinking, the best way to encourage contributions is:

* Provide IDE configurations for code formatting for all three major Java IDEs
* (Optional) provide a configuration file for a CLI formatting tool
* Describe how to handle formatting in pull requests (no harm stating the obvious, e.g. do not mix formatting changes with logic changes. For logic changes sanity proof commits with `git add -p` to ensure only the logic change lines are included)
* Describe what spelling conventions to use for new names (e.g. in your case you would say "prefer synch to sync" ;-) )

I recognise that there is a big ask there and it probably will take time to get there... but right now I do not see your current code style as being one that is compatible with any IDEs formatter. That makes Stephen :sob:

KostyaSha: @stephenc one additional note, @jglick can do any coding style here. But because of workflow changes enforced to all plugins and core, it must be readable for everybody. From reading code from all plugins and core i see only two difficulties for me :crying_cat_face: :
- random mess of spaces (or at all no spaces) by KK that followed by other contributors
- this annotation inlining by @jglick

Plus i don't understand what is the problem to add checkstyle to absolutely new plugin and exclude future coding style mess.

jglick: I never claimed compliance with any particular IDE’s autoformat options. As for NetBeans, I could have [saved you the trouble](https://netbeans.org/bugzilla/show_bug.cgi?id=104757). If you like to use an autoformatter on newly written code, fine for you, but existing code should not be reformatted anyway, at least until we have a next generation of SCM tools which operate at the level of parse trees rather than lines of text.

As to the placement of `@CheckForNull`, _logically_ this is part of the type, which `-source 8` treats properly. Whether today’s annotations can work with that model or not does not really matter; for a human reader, it is a type annotation and should be placed to match so long as a `-source 6` compiler does not complain about it.

`@Test` is an entirely different case. That is indeed a member annotation (not just in the deficient model of `-source 6`, but logically). I keep it on the same line as the method declaration because is a determiner, not an attribute. Same for `@Override`.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
jglick: No, because the method is just a helper used only from the method above.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: Good idea, I will try to remember to use `justification` in the future.

cps/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition2Test.java
jglick: `RestartableJenkinsRule`, yes. I doubt that actually applies in this test, however, since the objects are all mocks not bound to Jenkins disk storage.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Not sure what useful text to put there. An exception was caught; I have no real prediction about what it might be about. If it ever happens, the stack trace is going to be the only interesting thing.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
lanwen: agree with @stephenc :+1:

aggregator/src/test/java/org/jenkinsci/plugins/workflow/WorkflowRunTest.java
lanwen: agree with @stephenc :+1:

PR #160, opened 2015-07-28
[JENKINS-26761] WorkflowRun.checkouts sometimes null

jglick:
[JENKINS-26761](https://issues.jenkins-ci.org/browse/JENKINS-26761)

Supersedes #136.

- [X] recover gracefully from null `checkouts`
- [X] log a warning when this condition is detected
- [X] change the persistence of this field to solve a hypothetical cause
- [X] fix nullness after the edge condition of a build that failed to start (not thought to be the cause of reported issues)

@reviewbybees


oleg-nenashev:
:bee:

PR #163, opened 2015-07-30
Multibranch demo

jglick:
- [X] Switch to multibranch project, renaming `flow.groovy` to `jenkins.groovy`.
- [x] Put parallelized tests into their own `node` blocks (perhaps adding more dummy slaves), each with their own `checkout scm`, so that presentations can emphasize that the revision is guaranteed to match that used for `jenkins.groovy` and the initial `mvn package`.
- [x] Run `git-daemon` in the container with an exposed port with `daemon.receivepack=true` so that someone running the demo can clone and push to the repository from outside.

@reviewbybees (ref. CJP-3047)


demo/repo/jenkins.groovy
```diff
@@ -1,7 +1,7 @@
 def jettyUrl = 'http://localhost:8081/'
 
 node('slave') {
-    git url: '/tmp/files/repo'
+    checkout scm
```
amuniz: What's ```scm``` in this context? Current scm?

jglick: More or less. A magic variable referring to the commit used by this version of `jenkins.groovy`.

amuniz:
:bee: (I even saw it working :smile_cat:)

recena:
:bee: without a shadow of a doubt

PR #161, opened 2015-07-28
[JENKINS-26163] Fixed handling of AbstractStepExecutionImpl.onResume

jglick:
(Same as #158 but from `origin`.)

[JENKINS-26163](https://issues.jenkins-ci.org/browse/JENKINS-26163) appears to be blocking https://github.com/jenkinsci/ssh-agent-plugin/pull/4.

- [X] a test
- [x] extend API to somehow indicate running block-scoped executions
- [x] add test explicitly checking new `getCurrentExecutions` behavior including for block steps and terminated steps
- [x] implement `WaitForConditionStepTest.restartDuringDelay`
- [x] make sure `TimeoutStepTest` checks `onResume`
- ~~check as in #159 whether it is also possible to remove `LOADING_RUNS`~~ (looks too hard still)
- ~~consider solving [JENKINS-26148](https://issues.jenkins-ci.org/browse/JENKINS-26148) while we are at it~~ (eh, another day)

@reviewbybees

PR #165, opened 2015-07-30
[JENKINS-29738] TimeoutStep restarts the timeout value when onResume method is invoked

recena:
https://issues.jenkins-ci.org/browse/JENKINS-29738

In `TimeoutStep`, the timeout value was initialized when `onResume` was invoked.

For example, in this workflow script:

```
node {
  timeout(time: 60, unit: 'SECONDS') {
    sleep 10
    [Jenkins is restarted] <<<
    sleep 50
  }
}
```

The expected result of this execution is `ABORTED`:

1. first sleep step, 10 seg
1. second sleep step, 50 seg
1. restart process (safeRestart), enough to exceed 60 seg.

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
@@ -109,6 +109,36 @@ public class TimeoutStepTest extends Assert {
         });
     }
 
+    @Test
+    public void timeIsConsumed() throws Exception {
+        story.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "timeIsConsumed");
+                p.setDefinition(new CpsFlowDefinition(""
+                        + "node {\n"
+                        + "  timeout(time: 60, unit: 'SECONDS') {\n"
+                        + "    sleep 10\n"
+                        + "    semaphore 'timeIsConsumed'\n"
+                        + "    sleep 50\n"
+                        + "  }\n"
+                        + "}\n"));
+                WorkflowRun b = p.scheduleBuild2(0).getStartCondition().get();
+                SemaphoreStep.waitForStart("timeIsConsumed/1", b);
+            }
+        });
+        story.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.getItemByFullName("timeIsConsumed", WorkflowJob.class);
+                WorkflowRun b = p.getBuildByNumber(1);
+                SemaphoreStep.success("timeIsConsumed/1", null);
+                story.j.assertBuildStatus(Result.ABORTED, story.j.waitForCompletion(b));
+            }
+        });
+    }
+
+
     // TODO: timeout inside parallel
 
 }
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
@@ -138,7 +138,6 @@ public class TimeoutStepTest extends Assert {
         });
     }
 
-
     // TODO: timeout inside parallel
 
 }
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
@@ -1,6 +1,8 @@
 package org.jenkinsci.plugins.workflow.steps;
 
 import hudson.model.Result;
+import jenkins.model.CauseOfInterruption;
+import jenkins.model.InterruptedBuildAction;
 import org.jenkinsci.plugins.workflow.actions.ErrorAction;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.cps.nodes.StepAtomNode;
@@ -14,6 +16,8 @@ import org.junit.runners.model.Statement;
 import org.jvnet.hudson.test.BuildWatcher;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.RestartableJenkinsRule;
+
+import java.util.List;
 
 /**
  * @author Kohsuke Kawaguchi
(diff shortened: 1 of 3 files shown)


recena:
@reviewbybees Could you review these changes? Specially, @jglick

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
```diff
@@ -109,6 +109,35 @@ public void evaluate() throws Throwable {
...
+        story.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "timeIsConsumed");
+                p.setDefinition(new CpsFlowDefinition(""
+                        + "node {\n"
+                        + "  timeout(time: 60, unit: 'SECONDS') {\n"
+                        + "    sleep 10\n"
+                        + "    semaphore 'timeIsConsumed'\n"
+                        + "    sleep 50\n"
```
jglick: Maybe we could make these times shorter, so that test runs are quicker? IIUC running this test under heavy load will not cause it to fail; it would at worst cause it to _not_ fail if the bug fix were regressed. I suspect 10+10=20s would suffice to usually reproduce the bug.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
```diff
@@ -109,6 +109,35 @@ public void evaluate() throws Throwable {
...
+                SemaphoreStep.waitForStart("timeIsConsumed/1", b);
+            }
+        });
+        story.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.getItemByFullName("timeIsConsumed", WorkflowJob.class);
+                WorkflowRun b = p.getBuildByNumber(1);
+                SemaphoreStep.success("timeIsConsumed/1", null);
+                story.j.assertBuildStatus(Result.ABORTED, story.j.waitForCompletion(b));
```
jglick: May as well check for `ExceededTimeout` here.

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -22,26 +29,29 @@ public boolean start() throws Exception {
                 .withCallback(new Callback())
                 .withDisplayName(null)  // hide the body block
                 .start();
-        setupTimer();
+        long now = step.getUnit().convert(System.currentTimeMillis(), TimeUnit.MILLISECONDS);
+        end = now + step.getTime();
```
jglick: I think you got this backwards. The `convert` call should be on `step.time`, not `System.currentTimeMillis`.

jglick: Oh, I see what you are trying to do—convert milliseconds into the step’s defined unit—but it is not a good idea. Consider `timeout(time: 1, unit: 'HOURS')` at 4:58 PM. You will lose a lot of precision. And it is very confusing because `end` will effectively be “hours since the epoch”, which is a measure never used anywhere else in Java.

`now` and `end` should rather be defined to be milliseconds since the epoch, and `step.time` converted to milliseconds in `start`. `Timer.schedule` can then hardcode `TimeUnit.MILLISECONDS`. As a bonus, you can then make `step` `transient`, since it would be used only in `start`.

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -15,6 +20,8 @@
     private BodyExecution body;
     private transient ScheduledFuture<?> killer;
 
+    private long end;
```
jglick: You need to at least consider what happens to an existing build that was inside `timeout` when the upgrade happens: this field will be initialized to `0`. Now due to #161, fixed in this same release (assuming you get this review done soon), such a build would have ignored the timeout. Looks like the behavior now will be that the build gets canceled even if it was within its time (since `0 > now` will be false). Not a big deal—a one-time-only effect—but worth considering.

jglick:
:bug: for the strange time handling.

By the way please file this in JIRA, and the PR should also include a patch to `CHANGES.md` linking to that bug.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
recena: @jglick How I can check this?

recena:
@jglick JIRA is down at this moment. I'll do it when back.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepTest.java
jglick: `b.getAction(InterruptedBuildAction.class).getCauses()` (plus a null check on `getAction`).

recena: I had used this:

```
assertNotNull(run.getCause(TimeoutStepExecution.ExceededTimeout.class));
```

jglick: That should not even compile, much less run. `CauseOfInterruption` != `Cause`

recena: @jglick My mistake.

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/TimeoutStepExecution.java
```diff
@@ -22,26 +29,34 @@ public boolean start() throws Exception {
...
-            public void run() {
-                if (!body.isDone()) {
-                    // TODO use a proper CauseOfInterruption
-                    body.cancel(true);
+    /**
+     * Sets the timer to manage the timeout.
+     *
+     * @param now Current time in milliseconds.
+     */
+    private void setupTimer(@Nonnull final long now) {
```
jglick: :ant: it is meaningless to mark nullness on a primitive-typed parameter.

jglick:
:bee: though so it is not forgotten I recommend you add a `CHANGES.md` entry now (just without a JIRA link).

recena:
@jglick Don't worry. I'll fill Jira ticket and update the `CHANGES.md`

recena:
@jglick IMO, everything is ready to merge :+1:

PR #168, opened 2015-08-03
Introduced BodyExecutionCallback.TailCall

jglick:
Suggested by @kohsuke in a PR some months ago; finally got around to writing it.

@reviewbybees

### DIFF
--- a/step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/BodyExecutionCallback.java
+++ b/step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/BodyExecutionCallback.java
@@ -94,4 +94,38 @@ public abstract class BodyExecutionCallback implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
+    /**
+     * A convenience subclass for the common case that the step expects to run its block just once and return the same value (or throw the same error).
+     */
+    public static abstract class TailCall extends BodyExecutionCallback {
+
+        private static final long serialVersionUID = 1L;
+
+        /**
+         * Called when the body is finished.
+         * @param context the body context as passed to {@link #onSuccess} or {@link #onFailure}
+         * @throws Exception if anything is thrown here, the step fails too
+         */
+        protected abstract void finished(StepContext context) throws Exception;
+
+        @Override public final void onSuccess(StepContext context, Object result) {
+            try {
+                finished(context);
+                context.onSuccess(result);
+            } catch (Exception x) {
+                context.onFailure(x);
+            }
+        }
+
+        @Override public final void onFailure(StepContext context, Throwable t) {
+            try {
+                finished(context);
+                context.onFailure(t);
+            } catch (Exception x) {
+                context.onFailure(x);
+            }
+        }
+
+    }
+
 }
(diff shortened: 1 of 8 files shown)


step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/BodyExecutionCallback.java
```diff
@@ -94,4 +94,38 @@ public void onFailure(StepContext context, Throwable t) {
 
     private static final long serialVersionUID = 1L;
 
+    /**
+     * A convenience subclass for the common case that the step expects to run its block just once and return the same value (or throw the same error).
+     */
+    public static abstract class TailCall extends BodyExecutionCallback {
```
tfennelly: Curious ... why is it called `TailCall` as opposed to `TailCallback`? It extends a class with a `Callback` suffix on the name + all classes extending it seems to be called `Callback`.

tfennelly:
@jglick I think the code looks good insofar as I get the intent. I think a bit of an explanation of the issue being resolved (plug maybe a link to KK's original comments) would be helpful.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/BodyExecutionCallback.java
jglick: Because it represents a [tail call](https://en.wikipedia.org/wiki/Tail_call).

jglick:
Tried to find @kohsuke’s old comment, but seems to be lost in the mists of time. Basically he pointed out that there were more and more `BodyExecutionCallback` implementations which all have the same structure: `onSuccess` and `onFailure` both do some (the same) cleanup, then pass the result/error up the call stack. This PR merely removes some duplicated code and makes these steps a little more readable.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/BodyExecutionCallback.java
tfennelly: Fair enough. Maybe add a comment to that effect (or a link to that Wikipedia page) for dummies like me?

NIT: Still seems like impls of this are all callbacks (`TailCallCallback`).

:bee:

amuniz:
Did you split the test classes for some reason?

:bee: As far as I can understand it, I don't see anything weird.

jglick:
> Did you split the test classes for some reason?

Yes, see the first commit message.

tfennelly:
:bee:

Already gave it to you in a sub comment but giving it again. I forgot where I was when giving the original :)

jglick:
On reflection I guess this is not technically a tail call; that would be limited to `BodyExecutionCallback.wrap` where there is no possibility of a `finished` hook, in which case some hypothetical upgrade of the CPS engine could coalesce virtual stack frames. Not sure I care.

Whether the name is intuitive enough is another matter. Apparently it is not, but there were no alternate suggestions.

tfennelly:
@jglick I thought I suggested `TailCallCallback` as a way of reflecting what it is (a callback) and of being more consistent with what it is extending and what is extending it.

We're good though ... onward (as Sacha would say) !!!

PR #170, opened 2015-08-05
Minor changes to support checkpoints

jglick:
Actually only the patch to `CpsFlowExecution` directly supports checkpoints; the patch to `WorkflowRun` addresses a regression from #149 which I noticed also caused random test failures caught at the same time.

@reviewbybees

### DIFF
--- a/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -184,13 +184,11 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements Q
             newExecution.addListener(new GraphL());
             completed = new AtomicBoolean();
             logsToCopy = new LinkedHashMap<String,Long>();
-            newExecution.start();
-            
-            // It's only okay to initialise the global once the new FlowExecution
-            // has successfully started.
             execution = newExecution;
+            newExecution.start();
             executionPromise.set(newExecution);
         } catch (Throwable x) {
+            execution = null; // ensures isInProgress returns false
             finish(Result.FAILURE, x);
             executionPromise.setException(x);
             return;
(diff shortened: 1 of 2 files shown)


job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -184,13 +184,11 @@ public WorkflowRun(WorkflowJob job, File dir) throws IOException {
             newExecution.addListener(new GraphL());
             completed = new AtomicBoolean();
             logsToCopy = new LinkedHashMap<String,Long>();
-            newExecution.start();
-            
-            // It's only okay to initialise the global once the new FlowExecution
-            // has successfully started.
             execution = newExecution;
+            newExecution.start();
```
tfennelly: Should this not be on the line before so as to prevent `execution` from being initialized when `newExecution.start()` fails?

daniel-beck: If it fails by throwing, this will be reset in the catch block.

recena:
@jglick The source code modifications look fine. I don't have deep knowledge in WF but I didn't see nothing wrong therefore :bee:

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
tfennelly: @daniel-beck fair enough, I missed that. Thanks. I still think it should not be iniitalized until after it is successfully started. If it needs to be initialized before the call to `start`, then that suggests to me that there's a bigger problem somewhere else.

jglick: @tfennelly did you read the summary in 99ca243654430737de5f3be6367c9b24ccfdf5bb?

tfennelly: @jglick I didn't see that, sorry/thanks.

:bee: but it smells to me like there's something fundamentally wrong somewhere in the relationship(s) (`Owner`, `WorkflowRun`, `FlowExecution` etc).

tfennelly:
:bee: but it smells to me like there's something fundamentally wrong somewhere in the relationship(s) (`Owner`, `WorkflowRun`, `FlowExecution` etc).

jglick:
@tfennelly it is just a race condition. Once `start` is called, the build, well, starts. And then some steps can run, and some of them will be given a `FlowExecutionOwner` that they try to resolve, and `Owner.get` needs to look up `WorkflowRun.execution`.

PR #171, opened 2015-08-06
rename 'Running Steps' to 'Workflow Steps'

jtnord:
as the action will display all steps that have both run and are running to
call the action "Running steps" is incorrect and at first glance to a user
is incorrect. This changes the name to "Workflow Steps" which is
marginally better.

@reviewbybees

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 Only noting significant user-visible or major API changes, not internal code cleanups and minor bug fixes.
 
 ## 1.9 (upcoming)
-
+* "Running Steps" action if now called "Workflow Steps" as it will show steps for workflows that have long since completed.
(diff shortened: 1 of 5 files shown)


amuniz:
I've checked there is no acceptance test relying on this text
:bee:

recena:
:bee: details are always welcome

CHANGES.md
```diff
@@ -3,7 +3,7 @@
 Only noting significant user-visible or major API changes, not internal code cleanups and minor bug fixes.
 
 ## 1.9 (upcoming)
-
+* "Running Steps" action if now called "Workflow Steps" as it will show steps for workflows that have long since completed.
```
jglick: typo, but I can fix that up later

jglick:
:bee:

PR #172, opened 2015-08-06
s/jenkins.groovy/Jenkinsfile/g

jglick:
At @kohsuke’s request. @reviewbybees


multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
```diff
@@ -48,7 +48,7 @@
 @SuppressWarnings({"unchecked", "rawtypes"}) // core’s fault
 public class WorkflowMultiBranchProject extends MultiBranchProject<WorkflowJob,WorkflowRun> {
 
-    static final String SCRIPT = "jenkins.groovy";
+    static final String SCRIPT = "Jenkinsfile";
```
stephenc: My preference is that the file name should be configurable by the user but **default** to `Jenkinsfile`. The file is acting as a marker file to identify the branches that should be built. It is rather arrogant to assume that everyone will want to use `Jenkinsfile` as the file name.

OTOH adding the file name being configurable at a future date is not prevented... just more I am stating my opinion - influenced by experience with literate - given that we are touching this behaviour here

Vlatombe: Not sure it is as important as for Literate. The content of this file is really specific to Jenkins-workflow whereas for Literate, the marker file content is descriptive and could be used in another context.

tfennelly:
:bee:

Vlatombe:
:bee:

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
jglick: > the file name should be configurable by the user

I agree, but that is a subject for a future feature (perhaps a form control in `WorkflowBranchProjectFactory`). This is about changing the default filename.

PR #176, opened 2015-08-11
[JENKINS-29875] New API for long lived synchronous steps

amuniz:
[JENKINS-29875](https://issues.jenkins-ci.org/browse/JENKINS-29875)
[JENKINS-25879](https://issues.jenkins-ci.org/browse/JENKINS-25879)

```AbstractSynchronousStepExecution``` blocks the CPS VM thread while running every step. This leads to weird behaviors like the one described in the related JIRA issue when facing with long lived steps, i.e SCM steps (all previous flow nodes are added at the end, once the SCM step finishes).

This PR adds a new API for synchronous long lived steps. It ```run```s the step in a new thread so it does not block the CPS VM thread.

- [X] Add the API class
- [X] Add automated test
- [X] Update javadocs to reflect the new API
- [x] Modify existent steps to use this new API (when required)

### CODE REVIEW

File: aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStep.java [hunk 1]
```diff
@@ -0,0 +1,162 @@
...
+import org.jenkinsci.plugins.workflow.steps.StepContext;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.TestExtension;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import com.google.inject.Inject;
+
+public final class SynchronousNonBlockingStep extends AbstractStepImpl implements Serializable {
+
+    /** State of semaphore steps within one Jenkins home and thus (possibly restarting) test. */
```
jglick: Probably do not need this complexity, since synchronous steps cannot survive a restart anyway.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStep.java
```diff
@@ -0,0 +1,162 @@
...
+            } else if (s.errors.containsKey(k)) {
+                System.err.println("Immediately failing " + k);
+                getContext().onFailure(s.errors.get(k));
+            } else {
+                System.err.println("Blocking " + k);
+                s.contexts.put(k, Jenkins.XSTREAM.toXML(getContext()));
+            }
+
+            // Let's wait 2 seconds to give time to the main test thread to reach the wait point
+            Thread.sleep(2000);
```
jglick: Seems like this should not be necessary. I imagine you are lacking some notification.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStep.java
```diff
@@ -0,0 +1,162 @@
...
+
+            synchronized (s) {
+                s.started.add(k);
+                s.notifyAll();
+            }
+
+            // Let's wait 2 additional seconds after unblocking the test thread
+            // During this 2 seconds it will check for the messages in the build log
+            System.out.println("Sleeping inside the syncnonblocking thread");
+            Thread.sleep(2000);
```
jglick: We should rather be waiting for the test to send a notification that it is OK to proceed.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+        // Stop here until syncnonblocking has printed the message "Test Sync Message"
+        SynchronousNonBlockingStep.waitForStart("wait/1", b);
+
+        // At this point the execution is paused inside the synchronous non-blocking step
+        // Check for FlowNode created
+        FlowGraphWalker walker = new FlowGraphWalker(b.getExecution());
+        boolean found = false;
+        for (FlowNode n = walker.next(); n != null; n = walker.next()) {
+            if (n.getDisplayName().equals("Sync non-blocking Test step")) {
```
jglick: A little cleaner to check for `StepNode`.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStep.java
```diff
@@ -0,0 +1,162 @@
...
+            System.out.println("Continue syncnonblocking");
+
+            return null;
+        }
+
+        private static final long serialVersionUID = 1L;
+    }
+
+    @TestExtension
```
jglick: Read the Javadoc. You cannot use this annotation unless the class is nested inside a test suite. (And preferably specifying the name of the test case, unless it is used across several cases in the suite.)

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,61 @@
...
+                break;
+            }
+        }
+
+        System.out.println("Checking flow node added...");
+        assertTrue("FlowNode has to be added just when the step starts running", found);
+
+        // Check for message the test message sent to context listener
+        System.out.println("Checking build log message present...");
+        j.assertLogContains("Test Sync Message", b);
```
jglick: Use `JenkinsRule.waitForMessage` to avoid a race condition.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,61 @@
...
+        // The last step did not run yet
+        j.assertLogNotContains("Second message", b);
+
+        // No need to call success since it's called in the syncnonblocking thread
+        // SynchronousNonBlockingStep.success("wait/1", null);
+        System.out.println("Waiting until syncnonblocking (and the full flow) finishes");
+        j.waitUntilNoActivity();
```
jglick: `JenkinsRule.waitForCompletion`

jglick:
Another useful test case to add is that if a non-blocking synchronous step is running in one branch of `parallel`, other steps can continue to run in another branch, which is not true today (AFAIK).

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
```diff
@@ -0,0 +1,64 @@
+package org.jenkinsci.plugins.workflow.steps;
+
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+import jenkins.util.Timer;
+
+/**
+ * {@link StepExecution} that always executes synchronously and does not block the CPS VM thread.
```
jglick: `AbstractSynchronousStepExecution` Javadoc should link here.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
```diff
@@ -0,0 +1,64 @@
...
+        return false;
+    }
+
+    /**
+     * If the computation is going synchronously, try to cancel that.
+     */
+    @Override
+    public void stop(Throwable cause) throws Exception {
+        if (task != null) {
+            task.cancel(false);
```
jglick: You would need a test case proving that this actually sends an `InterruptedException` to anything sleeping or blocked on interruptible I/O in `run`.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+        getContext().onFailure(cause);
+    }
+
+    private class StepRunner implements Runnable {
```
jglick: `static`

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
```diff
@@ -0,0 +1,64 @@
...
+     * When this method returns, a step execution is over.
+     */
+    protected abstract T run() throws Exception;
+
+    @Override
+    public final boolean start() throws Exception {
+        task = Timer.get().schedule(new StepRunner(this), 0, TimeUnit.MILLISECONDS);
+        return false;
+    }
+
```
jglick: Should add an `onResume` which, at least for now, fails the step with an explanation that Jenkins restarted in the middle. (Later could be improved to allow the subclass to handle this case explicitly, if for example it is idempotent and wishes to simply resubmit the task.)

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Meat of the execution.
+     *
+     * When this method returns, a step execution is over.
+     */
+    protected abstract T run() throws Exception;
+
+    @Override
+    public final boolean start() throws Exception {
+        task = Timer.get().schedule(new StepRunner(this), 0, TimeUnit.MILLISECONDS);
```
jglick: `Timer` is capped at ten simultaneous threads, which is good for many purposes but imposes an artificial limit here. It might be better to use a custom executor service. (Can use some or all of the various wrapper classes that `Timer` uses, as appropriate.)

jglick: `schedule` is unnecessary here since there is no delay. Simpler to use `submit`.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousStepExecution.java
```diff
@@ -6,7 +6,8 @@
 import static hudson.model.Result.ABORTED;
 
 /**
- * {@link StepExecution} that always executes synchronously.
+ * {@link StepExecution} that always executes synchronously. This API should be used for short-lived tasks that
+ * return almost instantly. For long-lived tasks use {@link AbstractSynchronousNonBlockingStepExecution}.
```
jglick: Should clarify that long-lived tasks which do not need to run within a Java method call should use the more general `AbstractStepExecutionImpl`, and that `AbstractSynchronousNonBlockingStepExecution` is mainly for calling legacy Jenkins APIs which are potentially long-running and interruptible yet offer no asynchronous mode (for example because they block on a remoting call).

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStep.java
```diff
@@ -0,0 +1,162 @@
...
+            String k = step.k();
+            if (s.returnValues.containsKey(k)) {
+                System.err.println("Immediately running " + k);
+                getContext().onSuccess(s.returnValues.get(k));
+            } else if (s.errors.containsKey(k)) {
+                System.err.println("Immediately failing " + k);
+                getContext().onFailure(s.errors.get(k));
+            } else {
+                System.err.println("Blocking " + k);
+                s.contexts.put(k, Jenkins.XSTREAM.toXML(getContext()));
```
jglick: This copied code does not really make sense in this context, though the concepts could be adapted.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStep.java
jglick: …hence the test failure:

    java.lang.NoSuchMethodError: No such DSL method syncnonblocking found among [archive, bat, build, catchError, checkout, dir, echo, error, fileExists, git, injectSubtypesAsContext, input, load, mail, node, parallel, persistenceProblem, pwd, readFile, retry, semaphore, sh, sleep, stage, step, svn, timeout, tmpDir, tool, unarchive, waitUntil, withEnv, wrap, writeFile, ws]

amuniz:
@jglick everything addressed

amuniz:
Build failing with ```Build timed out (after 30 minutes). Marking the build as aborted.```.

amuniz:
@reviewbybees

amuniz:
Failed again with ```Build timed out (after 30 minutes). Marking the build as aborted.```

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,245 @@
...
+        public String getId() {
+            return id;
+        }
+
+        public static void waitForStart(String id, Run<?,?> b) throws IOException, InterruptedException {
+            State s = State.get();
+            synchronized (s) {
+                while (!s.started.contains(id)) {
+                    if (b != null && !b.isBuilding()) {
+                        throw new AssertionError(JenkinsRule.getLog(b));
```
jglick: This will throw an error if the build [has not started yet](https://github.com/jenkinsci/workflow-plugin/blob/5c3cbc2b23853cadd94efacb9616761858b8341f/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L476), a race condition.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,245 @@
...
+                    s.wait(1000);
+                }
+            }
+        }
+
+        public static final void notify(String id) {
+            State s = State.get();
+            synchronized (s) {
+                if (s.started.contains(id)) {
+                    s.started.remove(id);
```
jglick: Tip: can be simplified (if you like) to

```java
if (s.started.remove(id)) {
```

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,245 @@
...
+        public static class StepExecutionImpl extends AbstractSynchronousNonBlockingStepExecution<Void> {
+
+            @Inject(optional=true) 
+            private SynchronousNonBlockingStep step;
+
+            @Override
+            protected Void run() throws Exception {
+                System.out.println("Starting syncnonblocking " + step.getId());
+                // Send a test message to the listener
+                getContext().get(TaskListener.class).getLogger().println("Test Sync Message");
```
jglick:
```java
@StepContextParameter private TaskListener listener;
```

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+                State s = State.get();
+                synchronized (s) {
+                    s.started.add(step.getId());
+                    s.notifyAll();
+                }
+                if (step.getId().equals("waitinterrupt")) {
```
jglick: introduce constant

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
jglick: BTW if you use `BuildWatcher` (see its Javadoc for usage), which you certainly should in this test suite, there is no need to include `getLog` in messages.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,245 @@
...
+        j.waitForCompletion(b);
+    }
+
+    public static final class SynchronousNonBlockingStep extends AbstractStepImpl implements Serializable {
+
+        public static final class State {
+            private static State state;
+            static synchronized State get() {
+                if (state == null) {
+                    state = new State();
```
jglick: This works well enough here, where we are using a `@TestExtension` and so the set of callers is strictly limited and you can ensure that every test uses a distinct `id`, but FYI if you want to make `State.get()` be independent from test run to test run, you would need to tie the `State` instance to the `Jenkins` instance. Most easily by making it a non-`static` field in `DescriptorImpl`.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,245 @@
...
+        System.out.println("Waiting to syncnonblocking to start...");
+        SynchronousNonBlockingStep.waitForStart("wait", b);
+
+        // At this point the execution is paused inside the synchronous non-blocking step
+        // Check for FlowNode created
+        FlowGraphWalker walker = new FlowGraphWalker(b.getExecution());
+        boolean found = false;
+        for (FlowNode n = walker.next(); n != null; n = walker.next()) {
+            if (n instanceof StepNode) {
+                found = true;
```
jglick: :ant: this would pass if only the `echo` step had run. You need to verify the `StepDescriptor`!

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+        // Wait for syncnonblocking to be started
+        System.out.println("Waiting to syncnonblocking to start...");
+        SynchronousNonBlockingStep.waitForStart("wait", b);
+
+        // At this point the execution is paused inside the synchronous non-blocking step
+        // Check for FlowNode created
+        FlowGraphWalker walker = new FlowGraphWalker(b.getExecution());
+        boolean found = false;
+        for (FlowNode n = walker.next(); n != null; n = walker.next()) {
```
jglick: Do feel free to make it `Iterable` (would be helpful elsewhere too)…

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        System.out.println("Checking flow node added...");
+        assertTrue("FlowNode has to be added just when the step starts running", found);
+
+        // Check for message the test message sent to context listener
+        System.out.println("Checking build log message present...");
+        j.waitForMessage("Test Sync Message", b);
+        // The last step did not run yet
+        j.assertLogNotContains("Second message", b);
```
jglick: maybe also

```java
j.assertLogContains("First message", b);
```

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+        // Wait for syncnonblocking to be started
+        System.out.println("Waiting to syncnonblocking to start...");
+        // waitinterrupt is a special key that makes syncnonblocking step to wait inside a while  loop checking
+        // for Thread.interrupted()
+        SynchronousNonBlockingStep.waitForStart("waitinterrupt", b);
+
+        // At this point syncnonblocking is looping and checking for interruption
+
+        CpsFlowExecution e = (CpsFlowExecution) b.getExecutionPromise().get();
```
jglick: No need to cast to `CpsFlowExecution`.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,245 @@
...
+                State s = State.get();
+                synchronized (s) {
+                    s.started.add(step.getId());
+                    s.notifyAll();
+                }
+                if (step.getId().equals("waitinterrupt")) {
+                    while (true) {
+                        // Keep in an interruptible state
+                        if (Thread.interrupted()) {
+                            getContext().get(TaskListener.class).getLogger().println("Run thread interrupted");
```
jglick: Do you not mean to `break` at this point? Otherwise how does the step ever end?

jglick: This all seems weird. You could get rid of the use of `Thread.interrupted` _and_ the special key by simply catching `InterruptedException` in the `wait` below and printing a special message. That is more or less what a normal implementation would do anyway.

jglick: Or even have no special handling in the step at all—just let the `InterruptedException` be thrown out of `run` (again what a normal implementation would do). Then the test script can use a `try`-`catch` block to assert that `InterruptedException` was thrown from the step.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,245 @@
...
+        j.waitForMessage("Run thread interrupted", b);
+        j.waitForCompletion(b);
+    }
+
+    @Test
+    public void parallelTest() throws Exception {
+        WorkflowJob p = j.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("node {\n" +
+            "echo 'First message'\n" +
+            "parallel( a: { syncnonblocking 'wait0'; echo 'a branch'; }, b: { echo 'b branch' } )\n" +
```
jglick: :ant: this does not prove anything, because you are not preventing the second `echo` from running before `syncnonblocking`. You would need to use `semaphore` in the second branch so that you could wait for the `syncnonblocking` step to start but sleep, _then_ run the `echo 'b branch'` and demonstrate that the message was sent to the build log, then resume.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
```diff
@@ -0,0 +1,84 @@
...
+
+import hudson.util.DaemonThreadFactory;
+import hudson.util.NamingThreadFactory;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+/**
+ * Similar to {@link AbstractSynchronousStepExecution} but it always executes synchronously and does not block the CPS VM thread.
```
jglick: > Similar to `AbstractSynchronousStepExecution` but it always executes synchronously

does not make any sense: `AbstractSynchronousStepExecution` always executes synchronously too, of course.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * If the computation is going synchronously, try to cancel that.
+     */
+    @Override
+    public void stop(Throwable cause) throws Exception {
+        if (task != null) {
+            task.cancel(true);
+        }
+        getContext().onFailure(cause);
```
jglick: Debatable whether it is right to call `onFailure` here (when `task != null`). It is more polite to let the `InterruptedException` be processed by `run` however it prefers—perhaps just throwing it up (in which case `StepRunner` will call `onFailure` anyway), but perhaps first doing some cleanup, maybe even deciding to return successfully if it turns out it no longer needed to wait.

See `DurableTaskStep.Execution.stop`. That tries to do a polite interruption the first time, but if you click the stop button a second time, assumes the process is unresponsive, gives up, and calls `onFailure` directly.

#141 currently interprets a _third_ abort attempt as a hard kill: the program is immediately discarded and the result set to `ABORTED` without any processing. Probably I should refine that so that the second abort attempt is handled in the infrastructure by calling `StepContext.onFailure` on the running step(s), removing the special logic in `DurableTaskStep.Execution`; in that case it would be safe to simply delete this line here.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
```diff
@@ -0,0 +1,84 @@
...
+        getContext().onFailure(new Exception("Resume after a restart not supported for non-blocking synchronous steps"));
+    }
+
+    private static class StepRunner implements Runnable {
+
+        private final AbstractSynchronousNonBlockingStepExecution<?> step;
+
+        public StepRunner(AbstractSynchronousNonBlockingStepExecution<?> step) {
+            this.step = step;
+        }
```
jglick: Tip: could simplify code a bit by deleting the `step` field and constructor and just making this a non-`static` (i.e., “inner”) class.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
```diff
@@ -0,0 +1,84 @@
...
+            }
+        }
+    }
+
+    private static class StepRunnerExecutorService {
+
+        private static ExecutorService executorService;
+
+        public static synchronized ExecutorService get() {
+            if (executorService == null) {
```
jglick: :ant: as written, this method could be moved into `AbstractSynchronousNonBlockingStepExecution`, saving a class.

The Java lazy initialization idiom is to make `executorService` a `final` field with an initializer.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
```diff
@@ -0,0 +1,84 @@
...
+                executorService = Executors.newSingleThreadExecutor(new NamingThreadFactory(new DaemonThreadFactory(), "org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution"));
```
jglick: Using `newSingleThreadExecutor` guarantees that at most one `AbstractSynchronousNonBlockingStepExecution` can run in the entire JVM at any given time, which is even worse than `AbstractSynchronousStepExecution` which only blocks a single build’s VM!

I suspect `newCachedThreadPool` is what you meant.

jglick:
:bug: for `newSingleThreadExecutor`

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
amuniz: This would force me to use ```AbstractSynchronousNonBlockingStepExecution.this``` which is a notation not very clear from my point of view.

jglick: Well you can just write `getContext()` and the receiver is implicit. For `run()` it is true you need to disambiguate with `AbstractSynchronousNonBlockingStepExecution.this.`, part of the Java language since 1.1 but infrequently used. Guess it depends on what you are comfortable with.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
amuniz: All this was to avoid an ugly exception is tests logs. But I agree that it would be cleaner to remove all this ```Thread.interrupted``` stuff.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
amuniz: You are the owner of this and I'll follow your rules :smile: But I have to say that I feel better with the current code.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
amuniz: > Probably I should refine that so that the second abort attempt is handled in the infrastructure by calling StepContext.onFailure on the running step(s), removing the special logic in DurableTaskStep.Execution; in that case it would be safe to simply delete this line here.

So, should we wait for this?

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Similar to {@link AbstractSynchronousStepExecution} (it executes synchronously too) but it does not block the CPS VM thread.
+ * @see {@link StepExecution}
+ * @param <T> the type of the return value (may be {@link Void})
+ */
+public abstract class AbstractSynchronousNonBlockingStepExecution<T> extends AbstractStepExecutionImpl {
+
+    private transient volatile Future<?> task;
+
+    private transient static ExecutorService executorService;
```
jglick: BTW there is no purpose in marking a `static` field as `transient`.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,239 @@
...
+        WorkflowJob p = j.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("node {\n" +
+            "echo 'First message'\n" +
+            "parallel( a: { syncnonblocking 'wait0'; echo 'a branch'; }, b: { semaphore 'wait1'; echo 'b branch'; } )\n" +
+            "echo 'Second message'\n" +
+        "}"));
+        WorkflowRun b = p.scheduleBuild2(0).getStartCondition().get();
+
+        SynchronousNonBlockingStep.waitForStart("wait0", b);
+        SemaphoreStep.waitForStart("wait1/1", b);
```
jglick: FYI it is redundant to wait for `semaphore` to start if we are immediately going to let it proceed anyway (and wait for the following `echo`).

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
amuniz: Oh, yeah. Static is implicitly transient.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,239 @@
...
+        j.assertLogContains("Second message", b);
+        j.assertBuildStatusSuccess(b);
+    }
+
+    @Test
+    public void interruptedTest() throws Exception {
+        WorkflowJob p = j.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("node {\n" +
+            "echo 'First message'\n" +
+            "try { syncnonblocking 'wait' } catch(InterruptedException e) { echo 'Interrupted!' }\n" +
```
jglick: As noted [earlier](https://github.com/jenkinsci/workflow-plugin/pull/176/files#r36967923), if you are going with the simple implementation of `State.get` then you must ensure that every `id` is unique. This one duplicates the one from `basicNonBlockingStep`.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
jglick: No strong opinion either way, but I need to make a note in #141 to handle this.

jglick:
:bee: but please also update `CHANGES.md` in the patch.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
jglick: Then stay with the current code.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
amuniz: What would happen if next line (```SynchronousNonBlockingStep.notify```) before the ```semaphore``` is waiting? (IOW test thread reach this line before the CPS thread runs the semaphore step)

jglick: `success` may be called before the step is encountered; `SemaphoreStep` will just save the return value for later. Then we will `waitForMessage` anyway, proving that we got output from the second branch while the first branch was inside the synchronous step.

Nothing really wrong with the test as it is, just noting that there is a reason no other test calls `waitForStart` followed by `success` on the same ID with no intervening work.

amuniz:
@jglick ```CHANGES.md``` updated

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
```diff
@@ -0,0 +1,250 @@
...
+        System.out.println("Waiting to syncnonblocking to start...");
+        SynchronousNonBlockingStep.waitForStart("wait", b);
+
+        // At this point syncnonblocking is waiting for an interruption
+
+        FlowExecution e = b.getExecutionPromise().get();
+        // Let's force a call to stop. This will try to send an interruption to the run Thread
+        e.interrupt(Result.ABORTED);
+        System.out.println("Waiting for interruption");
+        j.waitForCompletion(b);
```
jglick: :ant: you used to be waiting for a message here, but deleted that for some reason. As written you are not really testing much, only that the build ends for some reason. Since there is a `catch` block, I think the expectation is that you want the build to succeed, and to print the `Interrupted!` message.

jglick:
:bee: despite the apparently weakened test.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
amuniz: Fixed.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/SynchronousNonBlockingStepTest.java
amuniz: Call to ```waitForStart``` removed.

jglick:
:bee:

jglick:
BTW confusing that the PR title and `CHANGES.md` entry use different JIRA IDs, which are anagrams of one another. Looks like a typo even though it is not. :-)

amuniz:
Yeah :smile: 

@reviewbybees done

PR #177, opened 2015-08-12
[JENKINS-27039] Option for stage step to cancel older executions

kreinloo:
[JENKINS-27039](https://issues.jenkins-ci.org/browse/JENKINS-27039)

Added `eager` flag to stage step. When a new build enters this stage, it will cancel any running build so that it could proceed on its own. Inspired by @jglick's [comment](https://issues.jenkins-ci.org/browse/JENKINS-27039?focusedCommentId=222237&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-222237).

Example:
```
stage name: 'build', eager: true
input message: 'Proceed?'
```


jglick:
This definitely needs coverage in `StageStepTest`.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/StageStep.java
```diff
@@ -41,6 +41,7 @@
 
     public final String name;
     @DataBoundSetter public @CheckForNull Integer concurrency;
+    @DataBoundSetter public @CheckForNull Boolean eager;
```
jglick: `boolean` suffices (especially as the default remains `false`).

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/StageStepExecution.java
```diff
@@ -147,6 +143,20 @@ private static synchronized void enter(Run<?,?> r, StepContext context, String n
...
         }
+        // Interrupts any build in the holding state
+        if (eager != null && eager) {
+            for (Integer i : stage.holding) {
+                try {
+                    Executor ex = job.getBuildByNumber(i).getExecutor();
+                    if (ex != null) {
+                        ex.interrupt(Result.ABORTED);
+                    }
+                    stage.holding.remove(i);
```
jglick: Need to use `Iterator.remove` to avoid `ConcurrentModificationException` I think.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/StageStepExecution.java
```diff
@@ -147,6 +143,20 @@ private static synchronized void enter(Run<?,?> r, StepContext context, String n
                 throw new IllegalStateException("the same flow is trying to reënter the stage " + name); // see 'e' with two dots, that's Jesse Glick for you! - KK
             }
         }
+        // Interrupts any build in the holding state
+        if (eager != null && eager) {
+            for (Integer i : stage.holding) {
+                try {
+                    Executor ex = job.getBuildByNumber(i).getExecutor();
+                    if (ex != null) {
+                        ex.interrupt(Result.ABORTED);
```
jglick: Better to use `NOT_BUILT` with a `CanceledCause`, as in the `cancel` method.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/StageStepExecution.java
```diff
@@ -147,6 +143,20 @@ private static synchronized void enter(Run<?,?> r, StepContext context, String n
                 throw new IllegalStateException("the same flow is trying to reënter the stage " + name); // see 'e' with two dots, that's Jesse Glick for you! - KK
             }
         }
+        // Interrupts any build in the holding state
+        if (eager != null && eager) {
+            for (Integer i : stage.holding) {
+                try {
+                    Executor ex = job.getBuildByNumber(i).getExecutor();
```
jglick: This probably works well enough, though it is rather indirect; nicer to use `FlowExecution.interrupt` if you can get access to the `FlowExecution`, perhaps by keeping a `FlowExectionOwner` in `Stage` (taking care to deal gracefully with null fields from serialized instances from older plugin versions). I am not sure offhand if there is a concrete downside to going through `Executor` like this; the test would need to verify that the interrupted build’s `result` and `CauseOfInterruption` are being set properly.

jglick:
Looks good so far. Marking WiP pending test coverage (whether written by you or me or someone else).

kreinloo:
As of now, it still cancels jobs with the ABORTED message.
```
Running: End of Workflow
Rejected by SYSTEM
Finished: ABORTED
```
I tried using the FlowExecution.interrupt as well, but the results were the same. Probably did something wrong. I'll try to improve it and write tests as well.

jglick:
`StageTest` IIRC.

Logic for handling aborts is a little subtle. Might need to check with a debugger why it is getting `ABORTED`. I suspect `stop` is called on the `Execution` and this overrides the cancellation cause you tried to set.

witokondoria:
Wouldnt it be nice to merge this feature and handle the exit status on a different issue (IMHO ABORTED fits well)

jglick:
Exit status is a minor issue (mainly consistency with the current behavior), but tests are vital.

jglick:
`support/src/test/java/org/jenkinsci/plugins/workflow/support/steps/StageStepTest.java` only tests configuration. [This](https://github.com/jenkinsci/workflow-plugin/blob/527cbb86ddfbd50eb6aef7823bc23fdec46023f3/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/StageTest.java) tests runtime behavior, the interesting stuff.

witokondoria:
What about another parameter to flag a stage as uncancellable? It could serve to define complex-one level semaphores

amuniz:
Tested locally, it works as advertised.
@kreinloo I'm going to add some tests and will file a new PR based on this one (keeping your commits).

amuniz:
Changed my mind. I'm working on a new step (`milestone`) that supersedes this PR. See [JENKINS-27039 comments](https://issues.jenkins-ci.org/browse/JENKINS-27039?focusedCommentId=249377&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-249377) for more context.

amuniz:
See https://github.com/jenkinsci/workflow-plugin/pull/355

jglick:
I think this has become obsolete; please try out https://github.com/jenkinsci/pipeline-milestone-step-plugin/pull/1 and let us know if it works for you.

PR #178, opened 2015-08-13
Iterable FlowGraphWalker

amuniz:
Add ```Iterable``` capabilities to ```FlowGraphWalker```

### DIFF
--- a/api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowGraphWalker.java
+++ b/api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowGraphWalker.java
@@ -2,7 +2,9 @@ package org.jenkinsci.plugins.workflow.graph;
 
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 
+import java.util.ConcurrentModificationException;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Set;
@@ -13,8 +15,7 @@ import java.util.Stack;
  * @author Kohsuke Kawaguchi
  */
 // TODO guarantee DFS order (is it not already in DFS order?)
-// TODO implement Iterable<FlowNode>
-public class FlowGraphWalker {
+public class FlowGraphWalker implements Iterable<FlowNode> {
     // queue of nodes to visit.
     // it's a stack and not queue to visit nodes in DFS
     Stack<FlowNode> q = new Stack<FlowNode>();
@@ -41,7 +42,9 @@ public class FlowGraphWalker {
     /**
      * Each time this method is called, it returns a new node until all the nodes are visited,
      * in which case this method returns null.
+     * @deprecated This class is {@link Iterable} now. Use {@link #iterator()} instead.
      */
+    @Deprecated
     public FlowNode next() {
         while (!q.isEmpty()) {
             FlowNode n = q.pop();
@@ -52,5 +55,60 @@ public class FlowGraphWalker {
             return n;
         }
         return null;
+    }
+
+    @Override
+    public Iterator<FlowNode> iterator() {
+        return new FlowGraphWalkerIterator();
+    }
+
+    private class FlowGraphWalkerIterator implements Iterator<FlowNode> {
+
+        private FlowNode next;
+        private int expectedCount;
+
+        public FlowGraphWalkerIterator() {
+            expectedCount = q.size();
+            next = getNext();
+        }
+
+        @Override
+        public boolean hasNext() {
+            return next != null;
+        }
+
+        @Override
+        public FlowNode next() {
+            FlowNode temp = next;
+            next = getNext();
+            return temp;
+        }
+
+        @Override
+        public void remove() {
+            getNext();
+        }
+
+        private FlowNode getNext() {
+            checkForComodification();
+            while (!q.isEmpty()) {
+                FlowNode n = q.pop();
+                if (!visited.add(n)) {
+                    continue;
+                }
+
+                List<FlowNode> parents = n.getParents();
+                addHeads(parents);
+                expectedCount = q.size();
+                return n;
+            }
+            return null;
+        }
+
+        private final void checkForComodification() {
+            if (q.size() != expectedCount) {
+                throw new ConcurrentModificationException();
+            }
+        }
     }
 }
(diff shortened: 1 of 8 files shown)


amuniz:
@reviewbybees

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowGraphWalker.java
```diff
@@ -53,4 +56,59 @@ public FlowNode next() {
         }
         return null;
     }
+
+    @Override
+    public Iterator<FlowNode> iterator() {
```
jglick: :ant: document that, unlike `Iterable.iterator` generally, this may be iterated only once.

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowGraphWalker.java
```diff
@@ -53,4 +56,59 @@ public FlowNode next() {
...
+        @Override
+        public FlowNode next() {
+            FlowNode temp = next;
+            next = getNext();
+            return temp;
+        }
+
+        @Override
+        public void remove() {
+            getNext();
```
jglick: :ant:

```java
throw new UnsupportedOperationException();
```

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowGraphWalker.java
```diff
@@ -53,4 +56,59 @@ public FlowNode next() {
...
+                addHeads(parents);
+                expectedCount = q.size();
+                return n;
+            }
+            return null;
+        }
+
+        private final void checkForComodification() {
+            if (q.size() != expectedCount) {
+                throw new ConcurrentModificationException();
```
jglick: I guess this is just for defense against deprecated calls to `FlowGraphWalker.next()`, not for actual changes to the flow graph, right?

job/src/main/java/org/jenkinsci/plugins/workflow/job/views/FlowGraphAction.java
```diff
@@ -72,8 +72,7 @@ private FlowGraphAction(WorkflowRun run) {
         }
         List<FlowNode> nodes = new ArrayList<FlowNode>();
         FlowGraphWalker walker = new FlowGraphWalker(exec);
-        FlowNode n;
-        while ((n = walker.next()) != null) {
+        for (FlowNode n : walker) {
             nodes.add(n);
         }
         Collections.reverse(nodes); // TODO would be more intuitive for FlowGraphWalker to do this to begin with
```
jglick: I think this comment can be removed.

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowGraphWalker.java
amuniz: Right

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowGraphWalker.java
```diff
@@ -13,8 +15,7 @@
  * @author Kohsuke Kawaguchi
  */
 // TODO guarantee DFS order (is it not already in DFS order?)
```
jglick: Would be nice to actually test the order that is returned. For example, from

```groovy
stage '1'
parallel a: {
  stage '2'
  stage '3'
}, b: {
  node {
    stage '4'
  }
}
stage '5'
```

a DFS walk would I think produce the order 5, 3, 2, `parallel`, 1, 4, `node`, plus some non-`StepNode`s.

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowGraphWalker.java
```diff
@@ -13,8 +15,7 @@
...
-// TODO implement Iterable<FlowNode>
-public class FlowGraphWalker {
+public class FlowGraphWalker implements Iterable<FlowNode> {
     // queue of nodes to visit.
     // it's a stack and not queue to visit nodes in DFS
     Stack<FlowNode> q = new Stack<FlowNode>();
```
jglick: BTW confusion between two comment lines and name of variable. I guess at some point it used to be a `Queue<FlowNode>`.

jglick:
:bee:

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowGraphWalker.java
amuniz: I think all this new code is already covered by existent tests. Let's leave this test for later.

amuniz:
@reviewbybees done

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowGraphWalker.java
tfennelly: How would this iterator work if used while the flow is still executing i.e. the graph still changing after the `FlowGraphWalkerIterator` was created and `expectedCount ` was initialized? That would be possible, right? Seems like `ConcurrentModificationException` would be getting thrown, which seems kinda wrong to me.

amuniz: I've not found any call to deprecated ```next()``` in the code which could produce that issue. Anyway, if we want to comply with the ```Iterable``` contract we have to check for concurrent modifications. The only way to modify a collection while iterating is by using the ```Iterator.remove()``` method (which is not supported here neither).

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowGraphWalker.java
jglick: As you like.

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowGraphWalker.java
jglick: @tfennelly it ought to be tested, but I believe this is not an issue. The flow graph is append-only: you can create new heads, but not modify the structure of existing nodes. Since `FlowGraphWalker` takes a snapshot of the current heads in its constructor, if new steps run while you are iterating they simply will not be seen. Akin to a `CopyOnWriteWhatever` collection.

jglick:
:bee:

PR #179, opened 2015-08-14
Error in dead branch cleanup

jglick:
We deny `DELETE` permission to regular users, but during garbage collection of branches the system does do a `Job.delete` which must be allowed.

@reviewbybees


recena:
:bee:

amuniz:
:bee:

jglick:
Unfortunately there is no test coverage for `DeadBranchStrategy` usage here yet. Some needs to be added.

PR #181, opened 2015-08-17
[JENKINS-29653] Added a little vertical space between  button and the textarea

recena:
[JENKINS-29653](https://issues.jenkins-ci.org/browse/JENKINS-29653)

workflow

@reviewbybees

### DIFF
--- a/cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
+++ b/cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
@@ -61,7 +61,7 @@ THE SOFTWARE.
         <j:set var="id" value="${h.generateId()}"/>
         <f:block>
             <input type="button" value="Generate Groovy" onclick="handlePrototype_${id}(); return false" class="submit-button primary"/>
-            <f:textarea id="prototypeText_${id}" readonly="true"/>
+            <f:textarea id="prototypeText_${id}" readonly="true" style="margin-top: 10px" />
         </f:block>
         <script>
             function handlePrototype_${id}() {


jglick:
:bee:

jtnord:
:bee:

recena:
@reviewbybees done

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
```diff
@@ -61,7 +61,7 @@ THE SOFTWARE.
         <j:set var="id" value="${h.generateId()}"/>
         <f:block>
             <input type="button" value="Generate Groovy" onclick="handlePrototype_${id}(); return false" class="submit-button primary"/>
-            <f:textarea id="prototypeText_${id}" readonly="true"/>
+            <f:textarea id="prototypeText_${id}" readonly="true" style="margin: 10px 0px;" />
```
KostyaSha: just add `<br/>` before? Style for textarea may be controlled and changed in core.

recena:
@jglick @jtnord I've review the CSS style because only `margin-top` necessary.

KostyaSha:
@recena maybe check with @tfennelly why such issue exists?

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
recena: @KostyaSha http://www.w3.org/TR/html-markup/br.html, specially here this section **Permitted parent elements**.

recena:
@KostyaSha Any opinion is welcome, specially from @tfennelly

recena:
@KostyaSha When Jenkins Core provides a UI toolkit more flexible, semantized, non table-based layout, etc... then, I'll review this PR again.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
KostyaSha: maybe try wrap with `f.entry() { <f:textarea ... }` that is common pattern in jenkins jelly? AFAIR in comparison to block in adds small indentation vertically

KostyaSha:
@recena btw, for any UI changes always attach 2 screenshots: before and after

KostyaSha:
@recena they can do table div migration for generated jelly codes, when you hacking styles you may end with issues as soon as somebody changed styles in core/jelly. I already had such fun with layous in build-pipeline plugin. Though for buttons it shouldn't be so much...

jglick:
:bee: (previous bees reset by apparent rebasing)

recena:
@jglick Yes, I reviewed the CSS style.

jtnord:
:bee: 

@KostyaSha whilst there is likely an issue in the Jenkins style sheet somewhere, getting that fixed will require an updated Jenkins and a bump in the workflow dependency which will delay getting this usability issue out to users.  (at present it looks like a tab header rather than a button).
A tweak would likely be needed when the root cause is fixed in Jenkins core - but at worst it will produce an ugly layout in the interim (which is exactly what this is attempting to to a quick fix of)  but that is better than locking users out by requiring a latest and greatest version of core.


Pre fix screenshot

recena:
@reviewbybees done

PR #183, opened 2015-08-20
Workflow script is a proud first class citizen - own it

michaelneale:
Groovy CPS DSL means not a lot to most, and to those who do know that TMA means, they already know it, so they don't need to be told it. flow.groovy should be "workflow script"

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -606,7 +606,7 @@ Therefore it makes sense to load the program from another source, one that you c
 
 ## Entire script from SCM
 
-The easiest way to do this is to select _Groovy CPS DSL from SCM_ when defining the workflow.
+The easiest way to do this is to select _Workflow script from SCM_ when defining the workflow.
 In that case you do not enter any Groovy code in the Jenkins UI; you just indicate where in source code you want to retrieve the program.
 (If you update this repository, a new build will be triggered, so long as your job is configured with an SCM polling trigger.)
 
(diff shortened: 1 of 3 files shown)


apemberton:
+1

Perhaps we should take a pass at UX tweak like this to workflow?

michaelneale:
@recena I will fix the java doc too and squash.

jglick:
:bee:, been meaning to do this for a long time, just never got around to it!

jglick:
> replace also other occurrences in Javadoc

The only place I see _Groovy CPS_ in Javadoc is in `CpsFlowExecution`, and if you are reading that you had better know what _Groovy CPS_ means. :-) So I think this is complete as is.

recena:
@jglick Yes, I was thinking about this file:

https://github.com/jenkinsci/workflow-plugin/blob/master/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java#L112

https://github.com/jenkinsci/workflow-plugin/blob/master/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java#L808

recena:
:bee:

svanoort:
:bee:  and :beer:  for the Austin Powers macro use!

michaelneale:
Thanks @jglick, agree, the javadocs are probably the place to own up to it being groovy. For most people its very high level and I even could describe it as "groovy or javascript like". I don't expect there is another non groovy implementation of the script for some time but if/when there is, can revisit how these choices are presented. 

VERY HAPPY ABOUT THIS 

\o/

PR #186, opened 2015-08-20
[JENKINS-26942] Stash files

jglick:
[JENKINS-26942](https://issues.jenkins-ci.org/browse/JENKINS-26942)

A fairly simple solution to the common problem in workflows of needing to save some files temporarily that were created in one node/workspace and get them back later. You could previously do this with `archive`/`unarchive` but the syntax was a little clumsier, and there were unwanted side effects to saving artifacts. (Now `archive` should be used only for deliberately saving artifacts.)

@reviewbybees esp. @jtnord

### DIFF
--- a/api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
+++ b/api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
@@ -1,0 +1,200 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.plugins.workflow.flow;
+
+import hudson.AbortException;
+import hudson.ExtensionList;
+import hudson.ExtensionPoint;
+import hudson.FilePath;
+import hudson.Util;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import hudson.org.apache.tools.tar.TarInputStream;
+import hudson.util.DirScanner;
+import hudson.util.io.ArchiverFactory;
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Map;
+import java.util.TreeMap;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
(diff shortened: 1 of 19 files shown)


api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -0,0 +1,200 @@
...
+     * Saves a stash of some files from a build.
+     * @param build a build to use as storage
+     * @param name a simple name to assign to the stash (must follow {@link Jenkins#checkGoodName} constraints)
+     * @param workspace a directory to use as a base
+     * @param includes a set of Ant-style file includes, separated by commas; null/blank is allowed as a synonym for {@code **} (i.e., everything)
+     * @param excludes an optional set of Ant-style file excludes
+     */
+    public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull TaskListener listener, @CheckForNull String includes, @CheckForNull String excludes) throws IOException, InterruptedException {
+        Jenkins.checkGoodName(name);
+        File storage = storage(build, name);
```
jtnord: should this complain if it already exists? it looks like now it silently overwrites (which may be correct - but it doesn't mention this on the listener so could cause confusion if a user cuts and pastes or uses some library function multiple times without thinking?

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -449,6 +450,11 @@ private void finish(@Nonnull Result r, @CheckForNull Throwable t) {
             }
         }
         FlowExecutionList.get().unregister(new Owner(this));
+        try {
+            StashManager.maybeClearAll(this);
+        } catch (IOException x) {
+            LOGGER.log(Level.WARNING, null, x);
```
jtnord: a message for a Jenkins admin would be nice.   "Failed to remove workflow stashes" ?

jtnord:
sweet :smile:   couple of :ant:s but otherwise time for :cake:

api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -0,0 +1,203 @@
...
+     * @param build a build to use as storage
+     * @param name a simple name to assign to the stash (must follow {@link Jenkins#checkGoodName} constraints)
+     * @param workspace a directory to use as a base
+     * @param includes a set of Ant-style file includes, separated by commas; null/blank is allowed as a synonym for {@code **} (i.e., everything)
+     * @param excludes an optional set of Ant-style file excludes
+     */
+    @SuppressFBWarnings(value="RV_RETURN_VALUE_IGNORED_BAD_PRACTICE", justification="fine if mkdirs returns false")
+    public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull TaskListener listener, @CheckForNull String includes, @CheckForNull String excludes) throws IOException, InterruptedException {
+        Jenkins.checkGoodName(name);
+        File storage = storage(build, name);
```
jglick: @jtnord wrote:
> should this complain if it already exists? it looks like now it silently overwrites (which may be correct - but it doesn't mention this on the listener so could cause confusion if a user cuts and pastes or uses some library function multiple times without thinking?

Yes, it should probably issue a warning if overwriting.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Well, I think this falls into the category of internal errors that get sent to the Jenkins log.

jtnord: yes but these internal errors should always say what the error is about so the poor admin knows that he can probably ignore this as it's not the cause of the issue he is looking for.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /**
+     * Restores a stash of some files from a build.
+     * @param build a build used as storage
+     * @param name a name passed previously to {@link #stash}
+     * @param workspace a directory to copy into
+     */
+    public static void unstash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull TaskListener listener) throws IOException, InterruptedException {
+        InputStream is = new FileInputStream(storage(build, name));
```
jtnord: directory traversal would allow any tar.gz to be read from the master.

jtnord: ideally a check and informative warning rather than just throwing a FileNotFoundException.
(the user won;t really know that /opt/jenkins/jobs/bob/builds/stashes/wibb1e.tar.gz was nescessarily a typos in the stash archive name?

api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
jglick: (resolved)

jtnord:
:bee:

PR #188, opened 2015-08-24
Additional warning from StepDescriptor.advanced

jglick:
Off the top of my head @cyrille-leclerc, @apemberton, and @gregsymons have gotten tripped up on the steps added by `docker-workflow`, which are not intended to be used from your script. Trying harder to guide users away from these.

(Really want a different UI, at least a form validation warning, maybe a checkbox guarding access to these steps, etc. All of those options seem to require more work than I have time for, due to lack of options in `/lib/form`. This is a quick fix in the interim. CC @tfennelly.)

@reviewbybees esp. @jtnord


jtnord:
:bee:

amuniz:
:bee:

PR #190, opened 2015-08-25
[JENKINS-29739] Snippet generation broken for single-parameter jobs

jglick:
[JENKINS-29739](https://issues.jenkins-ci.org/browse/JENKINS-29739)

@reviewbybees

### DIFF
--- a/cps/src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
+++ b/cps/src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
@@ -29,6 +29,7 @@ import com.gargoylesoftware.htmlunit.WebRequestSettings;
 import com.gargoylesoftware.htmlunit.WebResponse;
 import groovy.lang.GroovyObjectSupport;
 import groovy.lang.GroovyShell;
+import hudson.model.BooleanParameterDefinition;
 import hudson.model.BooleanParameterValue;
 import hudson.model.FreeStyleProject;
 import hudson.model.ParametersDefinitionProperty;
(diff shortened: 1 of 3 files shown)


cps/src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
```diff
@@ -149,8 +150,22 @@ private static void assertRoundTrip(Step step, String expected) throws Exception
...
+    }
+
+    @Issue("JENKINS-29739")
+    @Test public void generateSnippetForBuildTriggerSingle() throws Exception {
+        FreeStyleProject ds = r.jenkins.createProject(FreeStyleProject.class, "ds1");
+        FreeStyleProject us = r.jenkins.createProject(FreeStyleProject.class, "us1");
         ds.addProperty(new ParametersDefinitionProperty(new StringParameterDefinition("key", "")));
-        assertGenerateSnippet("{'stapler-class':'" + BuildTriggerStep.class.getName() + "', 'job':'../d1/ds', 'parameter':[{'name':'key', 'value':'stuff'}]}", "build job: '../d1/ds', parameters: [[$class: 'StringParameterValue', name: 'key', value: 'stuff']]", us.getAbsoluteUrl() + "configure");
+        assertGenerateSnippet("{'stapler-class':'" + BuildTriggerStep.class.getName() + "', 'job':'ds1', 'parameter': {'name':'key', 'value':'stuff'}}", "build job: 'ds1', parameters: [[$class: 'StringParameterValue', name: 'key', value: 'stuff']]", us.getAbsoluteUrl() + "configure");
```
jglick: Note the type of `parameter` in this case: `JSONObject` not `JSONArray`.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
```diff
@@ -85,7 +85,7 @@ public DescriptorImpl() {
         @Override public Step newInstance(StaplerRequest req, JSONObject formData) throws FormException {
             BuildTriggerStep step = (BuildTriggerStep) super.newInstance(req, formData);
             // Cf. ParametersDefinitionProperty._doBuild:
-            JSONArray params = formData.optJSONArray("parameter");
+            JSONArray params = JSONArray.fromObject(formData.get("parameter"));
```
amuniz: Is it granted that `formData.get("parameter")`is not null?

jglick: No, it may be null, which is handled on the next line and tested in `generateSnippetForBuildTriggerNone`.

amuniz: Ok. I thought `JSONArray.fromObject` would [throw an exception](https://github.com/jenkinsci/json-lib/blob/master/src/main/jdk15/net/sf/json/JSONArray.java#L189) if the parameter is `null` (java null not JS null).

If the test is actually checking that case, everything should be fine.

amuniz:
:bee:

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep.java
jglick: > I thought `JSONArray.fromObject` would throw an exception if the parameter is `null`

It does not, but it does return an array of null which was not wanted here, so I improved that.

amuniz:
A belated :bee: (after the last commit)

PR #191, opened 2015-08-25
Downstream flow build variables

jglick:
[JENKINS-29169](https://issues.jenkins-ci.org/browse/JENKINS-29169)

@reviewbybees


support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -160,7 +161,12 @@ public String getId() throws AbortException {
         if (build instanceof AbstractBuild) {
             return Collections.unmodifiableMap(((AbstractBuild<?,?>) build).getBuildVariables());
         } else { // not applicable
```
amuniz: Seems it is applicable now. Remove this comment?

amuniz:
:bee:

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
jglick: Well, `AbstractBuild.getBuildVariables` is inapplicable, but I agree the comment could be confusing now.

amuniz:
:bee:

PR #192, opened 2015-08-25
[JENKINS-30122] AbstractSynchronousNonBlockingStepExecution neglected to pick up the build’s authentication

jglick:
[JENKINS-30122](https://issues.jenkins-ci.org/browse/JENKINS-30122)

@reviewbybees esp. @amuniz

### DIFF
--- a/step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
+++ b/step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
@@ -34,7 +34,15 @@ public abstract class AbstractSynchronousNonBlockingStepExecution<T> extends Abs
 
     @Override
     public final boolean start() throws Exception {
-        task = getExecutorService().submit(new StepRunner(this));
+        task = getExecutorService().submit(new Runnable() {
+            @Override public void run() {
+                try {
+                    getContext().onSuccess(AbstractSynchronousNonBlockingStepExecution.this.run());
+                } catch (Exception e) {
+                    getContext().onFailure(e);
+                }
+            }
+        });
         return false;
     }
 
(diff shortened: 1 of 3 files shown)


step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
```diff
@@ -61,21 +78,4 @@ private static synchronized ExecutorService getExecutorService() {
         return executorService;
     }
 
-    private static class StepRunner implements Runnable {
```
amuniz: Was it needed? I see it clearer by having the run logic inside this private inner class. It would suffice to pass the `auth` object in the constructor and store it in a local field.

amuniz:
:bee:

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
jglick: Seemed easier and clearer to me to use a `final` local variable used close to its declaration, and let `javac` take care of the rest.

amuniz: There's no accounting for taste... but I'll try to follow your preferred notation in the future in order to have a uniform style across the project.

amuniz:
I must wait until the build finishes to set the bee... reverting to a :bug: until the findbugs issue is solved.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/AbstractSynchronousNonBlockingStepExecution.java
jglick: Do not worry about it, I was just editing this and found it clearer here.

amuniz:
:bee:

PR #193, opened 2015-08-25
[JENKINS-29656] Diagnostics

jglick:
[JENKINS-29656](https://issues.jenkins-ci.org/browse/JENKINS-29656)

Hoping that Java 7 methods will do a better job of reporting the problem.

@reviewbybees


oleg-nenashev:
:bee: Hopefully the atomic operation does not cause severe overheads

recena:
:bee: Always learning from you. Nice!

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -40,6 +40,8 @@
 import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
+import java.nio.file.Files;
```
amuniz: Wouldn't this throw `NoClassDefFoundException` just on `CpsThreadGroup` instantiation on Java 6?

jglick: `import` statements are not executable code.

As to the timing of a `ClassNotFoundException`/`NoClassDefFoundError`, all JREs I know of (well, OK, Oracle’s) wait to throw this until you actually try to run a method using the linked class. Thus so long as `rename` is not called until we have determined that we are on 7+ it should be safe to load `CpsThreadGroup`. I am not sure if the JVM specification permits an implementation to throw `NoClassDefFoundError` eagerly upon resolving `CpsThreadGroup`.

At any rate lots of Jenkins code uses the idiom here. The only way to be sure that any conceivable JVM could not be trying too aggressively to link in `Files` would be

```java
interface Link {
  void rename(File from, File to) throws IOException;
}
@IgnoreJRERequirement
public static class Link7 implements Link {
    @Override public void rename(File from, File to) throws IOException {
        Files.move(from.toPath(), to.toPath(), StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING);
    }
}
private static void rename(File from, File to) throws IOException {
    try {
        Class.forName("java.nio.file.Files");
        // Java 7
        String implName = Link.class.getName() + "7";
        Link impl;
        try {
            URL[] here = {Link.class.getProtectionDomain().getCodeSource().getLocation()};
            impl = new URLClassLoader(new MaskingClassLoader(Link.class.getClassLoader(), implName), here).loadClass(implName).asSubclass(Link.class).newInstance();
        } catch (Exception x) {
            throw new AssertionError(x);
        }
        impl.rename(from, to);
    } catch (ClassNotFoundException x) { // Java 6
        // fallback impl…
    }
}
```

The above is sure to be safe because there is no static reference to `Link7` and thus it could not be part of the class linking process for the caller. But at that point it is arguably easier to just use `java.lang.reflect` to find and call `Files.move`, `File.toPath`, and `StandardCopyOption` constants.

amuniz:
Thanks for the (rich) explanation!
I think I've seen `NoClassDefFoundException` just at class loading time, perhaps in OpenJDK, but not sure. Anyway :bee:

PR #194, opened 2015-08-26
load docs are not correct

jtnord:
@jglick my attempts to follow the original docs by getting a `echo "hello world"` to be called on load failed misserably - which leads me to believe that they are wrong.

I could only get the echo to be called if I did
```
def mylib = load 'mylib`
mylib()
```

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -616,7 +616,6 @@ For some cases you may prefer to explicitly load Groovy script text from some so
 The standard Groovy `evaluate` function can be used, but most likely you will want to load a flow definition from a workspace.
 For this purpose you can use the `load` step, which takes a filename in the workspace and runs it as Groovy source text.
 
-The loaded file can contain statements at top level, which are run immediately.
 That is fine if you only want to use a single executor and workspace, and do not mind hard-coding the slave label in the Jenkins job.
 For more complex cases, though, you want to leave the external script in full control of slave allocation.
 In that case the main script defined in the job can just load and run a closure (block of code to be run later):
@@ -645,9 +644,9 @@ def hello(whom) {
 }
 ```
 
-Note that while it can contain helper functions, the only code at top level is a Groovy `Closure`, which is the return value of the script, and thus of the main script’s `load` step.
+Note that while it can contain helper functions, the only code at top level is a Groovy `Closure`, which can be invoked with a call to `()` just like any other method defined in the function `lib.foo`
 
-The helper script can alternately define functions and return `this`, in which case the result of the `load` step can be used to invoke those functions like object methods.
+The helper script can also define functions and return `this`, in which case the result of the `load` step can be used to invoke those functions like object methods.
 An older version of the [Docker demo](demo/README.md) showed this technique in practice:
 
 ```groovy


jglick:
Can you be more explicit about what you tried and did not work? The loaded script and the main script in full, and the output?

[example](https://github.com/jenkinsci/workflow-plugin/blob/6d0982796508206da3fe33fc1839e56c9cb4beeb/aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/steps/RestartingLoadStepTest.java#L37-L50)

jtnord:
@jglick 

function.groovy
```groovy
{ ->
  echo "loaded functions.groovy v1"
}
```

flow.groovy
```groovy
node('debian') {
        checkout([$class: 'GitSCM', ....)
        load 'cjp-champagne-lib/lib/functions.groovy'
}
```

based on the original documentation I expected the `echo` step should have been invoked immediately.

but what I saw is (excluding the git fluff)

```
Running: Loaded script: cjp-champagne-lib/lib/functions.groovy
Running: Evaluate a Groovy source file into the workflow script : Body : Start
Running: Evaluate a Groovy source file into the workflow script : Body : End
Running: General Build Wrapper : Body : End
Running: General Build Wrapper : End
Running: Allocate node : Body : End
Running: Allocate node : End
Running: End of Workflow
Finished: SUCCESS
```

What I managed to get working was

function.groovy
```groovy
{ ->
  echo "loaded functions.groovy v1"
  return this;
}
```

flow.groovy
```groovy
node('debian') {
        checkout([$class: 'GitSCM', ....)
        def mylib = load 'cjp-champagne-lib/lib/functions.groovy'
        mylib()
}
```
at which point you have access to the functions defined as well as the closure and produces the following output - and gives you access to all functions defined in the groovy file.

```
Running: Loaded script: cjp-champagne-lib/lib/functions.groovy
Running: Evaluate a Groovy source file into the workflow script : Body : Start
Running: Evaluate a Groovy source file into the workflow script : Body : End
Running: Print Message
loaded functions.groovy v1
Running: General Build Wrapper : Body : End
Running: General Build Wrapper : End
Running: Allocate node : Body : End
Running: Allocate node : End
Running: End of Workflow
```

jglick:
Your original example does not work, because you are running a script which defines a `Closure` and then discards it without ever running anything.

What is wrong with `function.groovy`

```groovy
echo 'loaded functions.groovy'
```

and `flow.groovy`

```groovy
load 'functions.groovy'
```

?

jtnord:
> Your original example does not work, because you are running a script which defines a Closure and then discards it without ever running anything.

right but this is why I was getting confused in the docs.   

> The loaded file can contain statements at top level, which are run immediately.

I hope I'm not classed as a noddy developer (just new to groovy) - but if I've made these mistakes - what chance does the average joe have here?

Obviously the changes I have made to the docs are wrong, hopefully @estheryouhana can rework into something more user friendly?

estheryouhana:
Good for pointing that out --  i'm working on that content at the moment.

jglick:
Yes the docs need to be improved, just have not figured out exactly how yet.

PR #195, opened 2015-08-26
[JENKINS-25736] Adding the pause functionality

kohsuke:
[JENKINS-25736](https://issues.jenkins-ci.org/browse/JENKINS-25736)

@uaarkoti told me that he talks to lots of people who wants to be able
to arbitrarily pause workflows while they are executing.

This seems like a relatively easy addition as far as the engine is
concerned, so this is a prototype of that functionality.

Given the way the things are split into plugins I'm not sure how to
properly expose this to UI --- should the pausability be a base contract
of FlowExecution, or should I define TransientActionFactory impl that
depends on both WorkflowRun and CpsFlowExecution that adds an Action
from outside?

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/CpsFlowExecutionTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/CpsFlowExecutionTest.java
@@ -34,6 +34,7 @@ import java.util.List;
 import java.util.ListIterator;
 import org.codehaus.groovy.transform.ASTTransformationVisitor;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
+import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
 import org.jenkinsci.plugins.workflow.cps.CpsStepContext;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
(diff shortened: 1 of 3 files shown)


aggregator/src/test/java/org/jenkinsci/plugins/workflow/CpsFlowExecutionTest.java
```diff
@@ -189,5 +190,30 @@ private static void assertStepExecutions(FlowExecution e, String... steps) throw
...
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                p.setDefinition(new CpsFlowDefinition(
+                        "echo 'before'; semaphore 'one';  echo 'after';"));
+                WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+                SemaphoreStep.waitForStart("one/1", b);
+                CpsFlowExecution e = (CpsFlowExecution) b.getExecution();
+                e.pause(true);
+                e.waitForSuspension();
+                story.j.assertLogContains("before", b);
```
jglick: `waitForSuspension` is basically deprecated. Use `JenkinsRule.waitForMessage` instead.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java
```diff
@@ -119,6 +135,8 @@ private Object readResolve() {
 
     private void setupTransients() {
         runner = new CpsVmExecutorService(this);
+        if (paused==null) // earlier versions did not have this field.
+            paused = new AtomicBoolean();
```
jglick: You meant for this to be in the serial form as an `AtomicBoolean`?

jglick:
> should the pausability be a base contract of `FlowExecution`

Seems right, so long as there is some sensible fallback behavior.

jglick:
Needs some polish, but probably close to a usable implementation. Needs to be refiled in `workflow-cps-plugin`.

PR #196, opened 2015-08-27
Demonstrating that it is possible to build the same downstream project several times concurrently

jglick:
[Question](http://stackoverflow.com/questions/32228590/start-a-job-multiple-times-concurrently-with-workflow-plugin)

Note that it is hard for the `build` step to differentiate between `ScheduleResult.Created` and `.Existing`, which is the first step in diagnosing the mistake, since both return non-null from `getItem()` and `ParameterizedJobMixIn.scheduleBuild2` does not give access to the underlying `ScheduleResult`. Unfortunately I neglected to correct this mistake in https://github.com/jenkinsci/jenkins/pull/1771 so it is still not possible with a 1.623 core dependency (without copying & pasting a bunch of code from `ParameterizedJobMixIn`). Probably need yet another core patch to get it right.

@reviewbybees


rsandell:
:bee:

PR #197, opened 2015-08-27
Temporarily declare Jenkins-Version: 1.609.1

jglick:
Amends #184. Still building & testing against 1.609.2, but declaring a dependency only on 1.609.1, so the plugins will still be offered in the update center for 1.609.1 users. Desirable until more people have 1.609.2 installed.

(Would be nice to build against 1.609.1 but test against 1.609.2, but I could find no way to do that with Maven: you cannot select different artifact versions for `compile` vs. `test` scope. Deleting some entries from the Surefire classpath and then readding them with a different version might work, but this would cause havoc when running tests in a debugger and source-stepping, etc.)

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>1.609.2</version>
+        <version>1.609.2</version> <!-- but see Jenkins-Version override below -->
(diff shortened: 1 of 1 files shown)


svanoort:
:bee:

oleg-nenashev:
:bee: (nice to know about such trick for backported fixes)

tfennelly:
Ah, clever !!

PR #203, opened 2015-09-08
Add more detailed examples of usage for the input step.

bwalding:


bwalding:
@reviewbybees

recena:
@bwalding Great contribution. Examples and documentation always are welcome :bee:

tfennelly:
:bee:

jglick:
For technical reasons the `input` step is in `workflow-support`, not `workflow-basic-steps`.

The content is good (from a quick pass) but I think it belongs rather in `help.html` where people can find it when actually writing a flow. I do not want detailed step reference documentation to be on the website in general.

(CC @estheryouhana)

bwalding:
@jglick Happy to move it to support (support/input.md or some other location)

Though I would be concerned about only having this iteration of the document in the Jenkins inline help.html

1) it's a rather large chunk of text - overwhelming for inline help
2) it's not reachable except inside Jenkins (IIUC it's not published to another location)
3) when you say "reference documentation on the website" - it was only visible as part of the github browser of the .md file - of course I have no idea where it goes after that.  To be fair though - I wasn't really *trying* to write reference documentation - more just jotting down some usage examples.
4) to access the input-step help.html you need to use the snippet generator - but I'm not interested in generating snippets - I just wanted to find the step documentation.  Perhaps the snippet generator is poorly named in the general case if it is supplying step information.  Is it actually a "workflow step assistant"? or just "workflow assistant"?

jglick:
> it's a rather large chunk of text

That is my main concern as well. I may try to pare this down somewhat and see if it fits reasonably in step help.

> it's not reachable except inside Jenkins

This we expect to change in a separate system, related to [JENKINS-26126](https://issues.jenkins-ci.org/browse/JENKINS-26126).

> Is it actually a "workflow step assistant"?

Effectively yes.

I think we _do_ need to cover `input` better in `TUTORIAL.md`, though in a tutorial format rather than reference documentation.

apemberton:
@jglick just FYI I respectfully disagree and learn toward @bwalding's feelings  on: "I do not want detailed step reference documentation to be on the website in general." I think a better way forward is to "source of truth" the technical help content somewhere we all feel good about it then pull it into different tools. I dpo think Workflow needs more documentation on the web (in addition to, of course, the inline help always-up-to-date step reference per environment).

recena:
@apemberton I totally agree with *I dpo think Workflow needs more documentation on the web*, **but**  we should think in a different tool. Some `*.md` are not enough to manage something so important like Workflow. IMHO.

estheryouhana:
@ Manuel, what do you mean by another tool exactly?

jglick:
> a better way forward is to "source of truth" the technical help content somewhere […] then pull it into different tools

That is exactly the plan for `help*.html`.

jglick:
We will try to get your content fit into the style used in #211.

PR #204, opened 2015-09-08
[FIXED JENKINS-30346] Added a deleteDirStep to wipe out a directory

jtnord:
Added a new step (deleteDir) which will wipe out the current directory in a workspace.

[JENKINS-30346](https://issues.jenkins-ci.org/browse/JENKINS-30346)

@reviewbybees

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 Only noting significant user changes, not internal code cleanups and minor bug fixes.
 
+## 1.11 (not yet released)
+* [JENKINS-30346](https://issues.jenkins-ci.org/browse/JENKINS-30346): added a cross platform `deleteDir` step to recursivley delete a directory and its contents.
+
 ## 1.10 (Aug 31 2015)
 
 * [JENKINS-30122](https://issues.jenkins-ci.org/browse/JENKINS-30122): regression in usage of the Authorize Project plugin in 1.10-beta-1.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 Only noting significant user changes, not internal code cleanups and minor bug fixes.
 
 ## 1.11 (not yet released)
-* [JENKINS-30346](https://issues.jenkins-ci.org/browse/JENKINS-30346): added a cross platform `deleteDir` step to recursivley delete a directory and its contents.
+* [JENKINS-30346](https://issues.jenkins-ci.org/browse/JENKINS-30346): added a cross platform `deleteDir` step to recursively delete a directory and its contents.
 
 ## 1.10 (Aug 31 2015)
 
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStepTest.java
@@ -1,0 +1,128 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 CLoudbees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.plugins.workflow.steps;
+
+import java.io.File;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import hudson.model.queue.QueueTaskFuture;
+
+public class DeleteDirStepTest {
+
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
+
+    @Test
+    public void testDeleteEmptyWorkslapce() throws Exception {
+        String workspace = runAndGetWorkspaceDir(
+                "node {\n" +
+                "  deleteDir()\n" + 
+                "}");
+        File f = new File(workspace);
+        Assert.assertFalse("Workspace driectory should no longer exist", f.exists());
+    }
+
+    @Test
+    public void testDeleteTopLevelDir() throws Exception {
+        String workspace = runAndGetWorkspaceDir(
+                "node {\n" +
+                "  writeFile file: 'f1', text: 'some text'\n" +
+                "  writeFile file: 'f2', text: 'some text'\n" +
+                "  writeFile file: '.hidden', text: 'some text'\n" +
+                "  writeFile file: 'sub1/f1', text: 'some text'\n" +
+                "  writeFile file: '.sub2/f2', text: 'some text'\n" +
+                "  writeFile file: '.sub3/.hidden', text: 'some text'\n" +
+                "  echo 'workspace is ---' + pwd() + '---'\n" + 
+                "  deleteDir()\n" + 
+                "}");
+        File f = new File(workspace);
+        Assert.assertFalse("Workspace driectory should no longer exist", f.exists());
+    }
+
+    @Test
+    public void testDeleteSubFolder() throws Exception {
+        String workspace = runAndGetWorkspaceDir(
+                "node {\n" +
+                "  writeFile file: 'f1', text: 'some text'\n" +
+                "  writeFile file: 'f2', text: 'some text'\n" +
+                "  writeFile file: '.hidden', text: 'some text'\n" +
+                "  writeFile file: 'sub1/f1', text: 'some text'\n" +
+                "  writeFile file: '.sub2/f2', text: 'some text'\n" +
+                "  writeFile file: '.sub3/.hidden', text: 'some text'\n" +
+                "  echo 'workspace is ---' + pwd() + '---'\n" + 
+                "  dir ('sub1') {\n" +
+                "    deleteDir()\n" +
+                "  }" + 
+                "}");
+
+        File f = new File(workspace);
+        Assert.assertTrue("Workspace directory should still exist", f.exists());
+        Assert.assertTrue("f1 should still exist", new File(f, "f1").exists());
+        Assert.assertTrue("f1 should still exist", new File(f, "f2").exists());
+        Assert.assertFalse("sub1 should not exist", new File(f, "sub1").exists());
+        Assert.assertTrue(".sub2/f2 should still exist", new File(f, ".sub2/f2").exists());
+        Assert.assertTrue(".sub3/.hidden should still exist", new File(f, ".sub3/.hidden").exists());
+    }
+
+
+    /**
+     * Runs the given flow and returns the workspace used.
+     * @param flow a flow definition.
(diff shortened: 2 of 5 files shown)


CHANGES.md
```diff
@@ -2,6 +2,9 @@
 
 Only noting significant user changes, not internal code cleanups and minor bug fixes.
 
+## 1.11 (not yet released)
+* [JENKINS-30346](https://issues.jenkins-ci.org/browse/JENKINS-30346): added a cross platform `deleteDir` step to recursivley delete a directory and its contents.
```
jglick: typo

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStepTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+import hudson.model.queue.QueueTaskFuture;
+
+public class DeleteDirStepTest {
+
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
+
+    @Test
+    public void testDeleteEmptyWorkslapce() throws Exception {
```
jglick: typo

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStepTest.java
```diff
@@ -0,0 +1,128 @@
...
+                "node {\n" +
+                "  writeFile file: 'f1', text: 'some text'\n" +
+                "  writeFile file: 'f2', text: 'some text'\n" +
+                "  writeFile file: '.hidden', text: 'some text'\n" +
+                "  writeFile file: 'sub1/f1', text: 'some text'\n" +
+                "  writeFile file: '.sub2/f2', text: 'some text'\n" +
+                "  writeFile file: '.sub3/.hidden', text: 'some text'\n" +
+                "  echo 'workspace is ---' + pwd() + '---'\n" + 
+                "  dir ('sub1') {\n" +
+                "    deleteDir()\n" +
```
amuniz: IMO `deleteDir('sub1')` would be more user-friendly

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStepTest.java
```diff
@@ -0,0 +1,128 @@
...
+     * @param flow a flow definition.
+     * @return the workspace used.
+     */
+    private String runAndGetWorkspaceDir(String flow) throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+
+        p.setDefinition(new CpsFlowDefinition(flow));
+        QueueTaskFuture<WorkflowRun> future = p.scheduleBuild2(0);
+        r.assertBuildStatusSuccess(future);
+        WorkflowRun run = future.get();
```
jglick: More concisely:

```java
WorkflowRun run = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
```

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStepTest.java
```diff
@@ -0,0 +1,128 @@
...
+        
+        String workspace = null;
+        // what would be really nice is the equivalent of a head rather than a tail...
+        List<String> logs = run.getLog(50);
```
jglick: `JenkinsRule.getLog` is the nondeprecated way to get the complete log from a test.

Anyway it is much easier to call:

```java
FilePath ws = r.jenkins.getWorkspaceFor(p);
```

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStepTest.java
```diff
@@ -0,0 +1,128 @@
...
+    public JenkinsRule r = new JenkinsRule();
+
+    @Test
+    public void testDeleteEmptyWorkslapce() throws Exception {
+        String workspace = runAndGetWorkspaceDir(
+                "node {\n" +
+                "  deleteDir()\n" + 
+                "}");
+        File f = new File(workspace);
+        Assert.assertFalse("Workspace driectory should no longer exist", f.exists());
```
jglick: typo

jglick: Does this actually fail even if you omit `deleteDir`? IIRC `node` does not force creation of the workspace, so I suspect not. A better test would use `writeFile` first to ensure that the directory is nonempty.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStepTest.java
jglick: We decided to go with this for now. Other steps all default to using current directory as a context, with `dir` to change it.

jglick:
:bee:

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStepTest.java
amuniz: It's ok like it is, just a matter of usability.
I think it should use the context as it does now (`@StepContextParameter private FilePath cwd`) and accept a parameter to append to `cwd`, so the `dir ('sub1') {...}` stuff could be omitted.

It can be done in a future PR, of course.

amuniz:
:bee:

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStep.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+package org.jenkinsci.plugins.workflow.steps;
+
+import hudson.Extension;
+import hudson.FilePath;
+
+import org.kohsuke.stapler.DataBoundConstructor;
+
+public final class DeleteDirStep extends AbstractStepImpl {
```
tfennelly: :ant: no docs?

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStep.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        @Override
+        public String getFunctionName() {
+            return "deleteDir";
+        }
+
+        @Override
+        public String getDisplayName() {
+            return "Recursively delete the current directory from the workspace";
```
tfennelly: :ant: i18n?

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStep.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public static final class Execution extends AbstractSynchronousNonBlockingStepExecution<Void> {
+
+        @StepContextParameter
+        private transient FilePath cwd;
+
+        @Override
+        protected Void run() throws Exception {
+            cwd.deleteRecursive();
+            return null;
```
tfennelly: :ant: How about a check on the delete op and a log if it fails?

tfennelly: I'd say there'd be a future need for a `failOnError` option too Vs just ignoring it.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStepTest.java
jtnord: no it doesn't fail - but I wanted to make sure that the deleteDir step did not cause the build to fail in this case.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStepTest.java
jtnord: future enhancement.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStepTest.java
jtnord: OT: how does that work with workflow where you have multiple workspaces?

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStep.java
jtnord: I am following the current conventions of the plugin :-)

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStep.java
jtnord: following current conventions of the plugin :-)

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStep.java
jtnord: the return of the delete is void.  it throws an `IOException` if it fails - which I believe will translate the step into a failure.

if you don;t want to fail on error - then you have `try { deleteDir} catch { /* ignore /* }`  syntax in workflow.

jtnord:
@tfennelly I think I have addressed the :ant:s

basic-steps/src/main/java/org/jenkinsci/plugins/workflow/steps/DeleteDirStep.java
jglick: Right, nothing special need be done here. You are thinking of crappy `File.delete`. :-)

jglick:
> how does that work with workflow where you have multiple workspaces?

`myjob@2` etc. as usual.

jglick:
> no docs?

:ant: there ought to be a `help.html` for newly added steps (yes I know many existing steps are missing it).

jtnord:
>> how does that work with workflow where you have multiple workspaces?

> myjob@2 etc. as usual.

Well the workflow can have multiple workspaces - so this is different to a normal build that has concurrent builds.

```groovy
def branches [:]
 for (int i=1; i< 254; i++) {
  branches "branch$i" = {
   node ("192.168.1.${i}") {
     writeFile name: 'someFile', text: 'wibble'
     ws () {
       writeFile name: 'someFile', text: 'wibble' // should be a different file
     }  
   }
 }
 parallel branches
}

which workspace would be returned (there is by my count 508 workspaces?)

**Comment by jglick** [comment]
:bee:

**Comment by jglick** [comment]
> which workspace would be returned

Just the first one, with no `@` suffix.

> there is by my count 508 workspaces

Right, up to `myjob@508`.

**Comment by jtnord** [comment]
@reviewbybees done

File: basic-steps/src/main/resources/org/jenkinsci/plugins/workflow/steps/DeleteDirStep/help.html [hunk 10]
```diff
@@ -0,0 +1,6 @@
+<div>
+    Recursively deletes the current directory and its contents.
+    Symbolic links and junctions will not be followed but will be removed.
+    To delete a specific directory of a workspace wrap the <code>deleteDir</code>
```
oleg-nenashev: NIT: An optional <code>path</code> parameter would be more handy

oleg-nenashev:
:+1:  for the fix, but :bug: for wiping workspaces in such way (as it requested in JENKINS-30346).

Workspaces may contain SCM repositories, which straightforward deletion may bork the SCM configuration for future builds (e.g. Perforce). Jenkins has listeners and API calls, which allow to processes such cases (e.g. see <code>WorkspaceCleanupThread</code>). In order to support the workspace cleanup, there should be a more generic workflow command

jtnord:
Oleg an scm must be able to cope with a clean workspace. If not an initial checkout would always fail, and there is no gaurantee due to concurrency/ slave availability that you would not get a clean workspace anyway.
Whilst you can use this command to wipe out a workspace, you can use it to delete a specific directory.
The lack of callback is no different to running 'sh "rm *"' which is the current ugly hack.
As such it perfectly fixes the jira issue.  If you wan t the workspace cleanup thread to work better woth workflow that is orthogonal to this issue.

oleg-nenashev:
> Oleg an scm must be able to cope with a clean workspace.

The problem is not with the workspace. For example, Perforce stores the info about the client workspace statuses... on the server. If you wipe the workspace, it will still think the data has been already checked out unless you call a force pull. If you workaround the issue by creating new workspace registry entries, you may end up with overflow on the server side. Perforce plugin uses Jenkins extension points to mitigate such situations during the workspace cleanup.

I agree with merging this command, but IMO it just replaces "rm -rf *" as you stated above. It's definitely useful, but it should not be advertised as "workspace cleanup" in Jenkins terms.

oleg-nenashev:
I replaced the bug by :ant: . BTW it would be useful to mention that the command is not designed for the workspace cleanup in the command docs.

basic-steps/src/main/resources/org/jenkinsci/plugins/workflow/steps/DeleteDirStep/help.html
jtnord: this was discussed [further up](https://github.com/jenkinsci/workflow-plugin/pull/204/files#r38955418)

jglick:
@oleg-nenashev your concerns about P4 are valid; I recorded them in [JENKINS-26138](https://issues.jenkins-ci.org/browse/JENKINS-26138).

rsandell:
A good to have alternative to this could be using a "clean" property of ```ws``` something like

```groovy
ws(name: 'dir', clean: true) {
    // some block
}
```

jglick:
@rsandell would not work for the most common use case, of cleaning the original workspace, since it cannot be relocked.

PR #206, opened 2015-09-10
[JENKINS-27421] Investigating ArrayList.Itr problem

jglick:
[JENKINS-27421](https://issues.jenkins-ci.org/browse/JENKINS-27421)

Simply trying

```diff
diff --git a/support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverWriter.java b/support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverWriter.java
index 1d18130..ba22f52 100644
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverWriter.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverWriter.java
@@ -100,7 +101,11 @@ public class RiverWriter implements Closeable {
         dout.writeInt(0);     // we'll back-fill this address with the pointer to the ephemerals stream
 
         MarshallingConfiguration config = new MarshallingConfiguration();
-        //config.setSerializabilityChecker(new SerializabilityCheckerImpl());
+        config.setSerializabilityChecker(new SerializabilityChecker() {
+            @Override public boolean isSerializable(Class<?> clazz) {
+                return true;
+            }
+        });
         config.setObjectResolver(new ObjectResolver() {
             public Object readResolve(Object o) {
                 throw new IllegalStateException();
```

does not work: apparently `SerializabilityChecker` cannot be used to treat something as `Serializable`:

```
java.lang.NullPointerException
	at org.jboss.marshalling.reflect.SerializableClass.invokeConstructorNoException(SerializableClass.java:510)
	at org.jboss.marshalling.reflect.SerializableClass.callNonInitConstructor(SerializableClass.java:484)
	at org.jboss.marshalling.river.RiverUnmarshaller.doReadNewObject(RiverUnmarshaller.java:1283)
	at …
```

Anyway I doubt this simple approach would work even if it did try to serialize all non-`transient` fields, since `Itr.expectedModCount` would be positive after a restart, while `AbstractList.modCount` would be reset to zero, so the next loop call would just get a `ConcurrentModificationException`. Would need to force `expectedModCount` to be treated as `transient` too.

In the meantime, it is useful to have the problem demonstrated even without a fix, so I do not consider this a work in progress.

@reviewbybees


oleg-nenashev:
:bee: for the refactoring and :bee: for the new test even though it required a per-commit review. I would prefer to see it in separate PRs, but it's only an :ant: in this case.

tfennelly:
:bee: 

NIT: I think it would be useful to add a comment or two in the test where it's currently failing (if not ignored). Also, having flow script all on one line makes it a bit painful to read.

jglick:
> I would prefer to see it in separate PRs

True, the first two commits could have been in an upstream “refactoring” PR.

> it would be useful to add a comment or two in the test where it's currently failing

In this case where in the JUnit test code the failure occurs is actually not very interesting; the build fails, that is the bug. The interesting bit is the stack trace in the build log, the detail message of which is noted.

> having flow script all on one line makes it a bit painful to read

Which script? The new one is split across three lines.

PR #207, opened 2015-09-10
Added help for BuildTriggerStep.job

jglick:
@reviewbybees

### DIFF
--- a/support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help-job.html
+++ b/support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help-job.html
@@ -1,0 +1,7 @@
+<div>
+    Name of a downstream job to build.
+    May be another Workflow job, but more commonly a freestyle or other project.
+    Use a simple name if the job is in the same folder as this upstream workflow job;
+    otherwise can use relative paths like <code>../sister-folder/downstream</code>
+    or absolute paths like <code>/top-level-folder/nested-folder/downstream</code>.
+</div>


support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStep/help-job.html
```diff
@@ -0,0 +1,7 @@
+<div>
+    Name of a downstream job to build.
+    May be another Workflow job, but more commonly a freestyle or other project.
+    Use a simple name if the job is in the same folder as this upstream workflow job;
+    otherwise can use relative paths like <code>../sister-folder/downstream</code>
+    or absolute paths like <code>/top-level-folder/nested-folder/downstream</code>.
```
KostyaSha: afair as the last step plain name fallbacks to root job

jglick: Yes…but I do not want to advertise that, because it is only for compatibility with historical settings, which does not apply here.

jglick:
Test failures are the infamous [JENKINS-30395](https://issues.jenkins-ci.org/browse/JENKINS-30395).

KostyaSha:
:+1:

recena:
:bee: I prefer to use `<p>` to mark different paragraphs

oleg-nenashev:
:bee:

oleg-nenashev:
+1 for decoupling the summary and the description into two paragraphs BTW

jglick:
> I prefer to use `<p>` to mark different paragraphs

The convention for `help*.html` is to use `<div>` at top level. It is possible to use `<p>` or even `<br>` within the file to distinguish paragraphs (in this case there is only one). Wrapping the entire file in `<p>` is undesirable because it introduces extra space which looks wrong in this context (possibly fixable in core using better CSS).

recena:
@jglick To use `<br>` to distinguish paragraphs is wrong. We should use `<p>` and next step, review the CSS in core.

PR #208, opened 2015-09-15
Compatibility cleanup

estheryouhana:
@reviewbybees (1 of 2 PRs)


COMPATIBILITY.md
```diff
@@ -1,13 +1,17 @@
-# Plugin Compatibility with Workflow
+# Workflow Compatibility Plugins
```
jglick: Seems more confusing than the original. These are not plugins that offer compatibility with Workflow. They are plugins that existed independently, and here we are listing whether and how they are compatible with Workflow.

COMPATIBILITY.md
```diff
@@ -1,13 +1,17 @@
...
 
-For architectural reasons, plugins providing various extensions of interest to builds cannot be made automatically compatible with Workflow.
-Typically they require use of some newer APIs, large or small (see the bottom of this document for details).
-This document captures the ongoing status of plugins known to be compatible or incompatible.
+For architectural reasons, plugins that provide various extensions of interest to builds are not automatically compatible with Workflow.
+Typically, they require use of newer APIs - both large or small.
```
jglick: :ant: use m-dashes please (—)

COMPATIBILITY.md
```diff
@@ -1,13 +1,17 @@
...
-Entries list the class name serving as the entry point to the relevant functionality of the plugin (generally an `@Extension`), the plugin short name, and implementation status.
+Entries list:
+* the class name serving as the entry point to the relevant functionality of the plugin (generally an `@Extension`)
+* the plugin short name  
+* implementation status
 
-## SCMs
+## Compatible SCMs
 
+The following SCMs are compatible. See [this guide](scm-step/README.md#supporting-workflow-from-an-scm-plugin) for further details on making `SCM`s compatible.
```
jglick: Looks like your Git merge failed, because this is reintroducing a line I intentionally deleted some time ago. Always review the effective diff after a merge for sanity.

oleg-nenashev:
:bug: for not specifying "Workflow  Plugin" in the page title. Workflow term is unique in Jenkins. E.g. https://wiki.jenkins-ci.org/display/JENKINS/jBPM+workflow+plugin and https://wiki.jenkins-ci.org/display/JENKINS/Jenkow+Plugin

COMPATIBILITY.md
```diff
@@ -67,7 +71,7 @@ Entries list the class name serving as the entry point to the relevant functiona
 - [X] `NpmPackagesBuildWrapper` (`nodejs`): scheduled to be supported as of 0.3
 
 ## Triggers
```
oleg-nenashev: Rename to "Build Triggers"?

COMPATIBILITY.md
```diff
@@ -77,7 +81,7 @@ Entries list the class name serving as the entry point to the relevant functiona
 - [X] `bitbucket`: supported as of 1.1.2
 
 ## Clouds
-
+The following Clouds are delivered:
```
oleg-nenashev: The sentence is vague a bit. "Workflow supports the on-demand nodes provisioning from the following Cloud providers" is better IMHO

COMPATIBILITY.md
```diff
@@ -77,7 +81,7 @@ Entries list the class name serving as the entry point to the relevant functiona
...
 - [ ] `elasticbox`: [JENKINS-25978](https://issues.jenkins-ci.org/browse/JENKINS-25978) (could also include build wrapper integration)
 - [ ] `mansion-cloud`: [JENKINS-24815](https://issues.jenkins-ci.org/browse/JENKINS-24815)
 - [X] `mock-slave` (for prototyping): supported as of 1.7
```
oleg-nenashev: Reference plugin Wiki pages where it is possible

COMPATIBILITY.md
```diff
@@ -87,7 +91,7 @@ Entries list the class name serving as the entry point to the relevant functiona
 - [X] `ec2`: known to work as is
```
oleg-nenashev: The version specifications above do not provide enough information to me. Does it mean that only one version is supported? Or do we support all further versions?

COMPATIBILITY.md
jglick: Agreed.

COMPATIBILITY.md
jglick: Probably would be useful.

COMPATIBILITY.md
jglick: Could say something like “known to work at least in version 1.23”, though that could be misread as implying that older or newer versions do not work, when really what we want to say is that we (of course) tested a specific version, which worked, and there is no reason to believe that other versions do not work as well.

COMPATIBILITY.md
```diff
@@ -114,12 +118,12 @@ For cases when a first-class Workflow step (rather than an adaptation of functio
...
 
-If you are maintaining (or creating) a plugin and wish its features to work smoothly with Workflow, there are a number of special considerations.
+If you are maintaining (or creating) a plugin and want its features to work smoothly with Workflow, there are a number of special considerations.
 
-## Extension points accessible via metastep
+## Extension Points Accessible via Metastep
 
 Several common types of plugin features (`@Extension`s) can be invoked from a Workflow script without any special plugin dependencies so long as you use newer Jenkins core APIs.
-Then there is “metastep” in Workflow (`step`, `checkout`, `wrap`) which loads the extension by class name and calls it.
+ The “metastep” in Workflow (`step`, `checkout`, and `wrap`). It loads the extensions by class name and calls it.
```
jglick: Even less grammatical than the original.

COMPATIBILITY.md
estheryouhana: fixed.

COMPATIBILITY.md
estheryouhana: fixed.

COMPATIBILITY.md
estheryouhana: What if we used: `1.7+ of Workflow Plugin` instead. This implies support for this plugin was added in 1.7 and beyond of the workflow plugin.

COMPATIBILITY.md
estheryouhana: Please expound.

COMPATIBILITY.md
jglick: Well that is not the case. We just tested some version of the EC2 plugin with some version of Workflow at some point and it just worked without changes.

COMPATIBILITY.md
jglick: Fixed how? There was no subsequent change here.

BTW if there is a line comment in a GH PR, and you subsequently add a commit changing that line (or one nearby), GH will automatically hide the thread from the _Files changed_ tab (and show the comment as being _on an outdated diff_ in the _Conversation_ tab). That is the usual way of tracking which comments are outstanding.

COMPATIBILITY.md
estheryouhana: Done. I fixed conflicts.

jglick:
I think effd4f07c25ed85eade951c49795f67eb48b3b61 was a bad merge; somehow this PR was botched. Have to start from scratch I am afraid.

PR #209, opened 2015-09-17
JobPropertyStep

jglick:
[JENKINS-30519](https://issues.jenkins-ci.org/browse/JENKINS-30519)

- [X] create `JobPropertyStep`
- [x] fix or work around [JENKINS-26535](https://issues.jenkins-ci.org/browse/JENKINS-26535) problem in `DescribableHelper`
- [x] work around lack of `@DataBoundConstructor` on `ParametersDefinitionProperty`
- [x] figure out how to deal with the nightmare that is the design of `JobProperty/config.jelly`
- [ ] ~~fix [JENKINS-29711](https://issues.jenkins-ci.org/browse/JENKINS-29711) problem in `Snippetizer`~~ (leave it for another day)
- [X] delete `WorkflowParameterDefinitionBranchProperty`
- [x] `buildDiscarder`
- [ ] ~~`triggers`~~ (not now)
- [ ] ~~`quietPeriod`~~ (not now)
- [ ] ~~`concurrentBuild`~~ (not now)
- [ ] ~~`authToken`~~ (not now)
- [x] suppress `RateLimitBranchProperty`
- [ ] ~~test against https://github.com/jenkinsci/github-plugin/pull/84 when released~~ (have tested against snapshots)
- [X] move `OptionalJobProperty` upstream: #228
- [X] move `BuildDiscarderProperty` upstream: #229

@reviewbybees esp. @kohsuke & @stephenc


multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
```diff
@@ -0,0 +1,177 @@
...
+            }
+            return result;
+        }
+
+    }
+
+    @Extension public static class HideSuperfluousBranchProperties extends DescriptorVisibilityFilter {
+
+        @Override public boolean filter(Object context, Descriptor descriptor) {
+            if (context instanceof WorkflowMultiBranchProject && (descriptor.clazz == RateLimitBranchProperty.class || descriptor.clazz == BuildRetentionBranchProperty.class)) {
```
amuniz: It seems hacky? If another property like RateLimitBranchProperty is added for some reason, it should be excluded here too. Is not there another way to exclude them?

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -0,0 +1,143 @@
...
+        assertEquals(-1, lr.getArtifactDaysToKeep());
+        assertEquals(3, lr.getArtifactNumToKeep());
+    }
+
+    @Issue("JENKINS-30206")
+    @Test public void useParameter() throws Exception {
+        sampleRepo.init();
+        ScriptApproval.get().approveSignature("method groovy.lang.Binding hasVariable java.lang.String"); // TODO add to generic whitelist
+        sampleRepo.write("Jenkinsfile",
+                "properties([[$class: 'ParametersDefinitionProperty', parameterDefinitions: [[$class: 'StringParameterDefinition', name: 'myparam', defaultValue: 'default value']]]])\n" +
```
amuniz: I know it's what we have for now, but I have to say this is the longest way to define a key=value pair I have ever seen in my life.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
jglick: We want to include unknown branch properties by default. There is no way that I can think of to mechanically determine whether they are simply wrappers for job properties.

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
jglick: [JENKINS-29922](https://issues.jenkins-ci.org/browse/JENKINS-29922) should help somewhat. Beyond that, we are stuck with what Jenkins has defined historically (long before code-as-config was a consideration). Creating a special “snowflake” syntax for each known job property will not scale.

PR #210, opened 2015-09-21
Better document and test @NonCPS

jglick:
@reviewbybees

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/SerializationTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/SerializationTest.java
@@ -185,4 +185,22 @@ public class SerializationTest extends SingleJobTestBase {
         });
     }
 
+    @Test public void nonCps() {
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                p = jenkins().createProject(WorkflowJob.class, "demo");
+                p.setDefinition(new CpsFlowDefinition(
+                    "echo \"first parse: ${parse('foo <version>1.0</version> bar')}\"\n" +
+                    "echo \"second parse: ${parse('foo bar')}\"\n" +
+                    "@NonCPS def parse(text) {\n" +
+                    "  def matcher = text =~ '<version>(.+)</version>'\n" +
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
:bee:

amuniz:
:bee:

aggregator/src/test/java/org/jenkinsci/plugins/workflow/SerializationTest.java
```diff
@@ -185,4 +185,22 @@ public void stop(Throwable cause) throws Exception {
...
+                    "echo \"first parse: ${parse('foo <version>1.0</version> bar')}\"\n" +
+                    "echo \"second parse: ${parse('foo bar')}\"\n" +
+                    "@NonCPS def parse(text) {\n" +
+                    "  def matcher = text =~ '<version>(.+)</version>'\n" +
+                    "  matcher ? matcher[0][1] : null\n" +
+                    "}\n", true));
+                b = story.j.assertBuildStatusSuccess(p.scheduleBuild2(0));
+                story.j.assertLogContains("first parse: 1.0", b);
+                story.j.assertLogContains("second parse: null", b);
+            }
```
tfennelly: Stupid question... how is this test testing serialization (and `@NonCPS`) ?

jglick: Not a stupid question at all. It is testing that (a) `@NonCPS` can be used, (b) there is _not_ a serialization error as in some variants shown in the tutorial. The trick is that the test will also pass without `@NonCPS`, because nothing happens to ask for the program state to be saved _in the middle of this method running_. Simulating the request for save would be great, but I do not know how to do that, except by running a step—which you cannot do in a `@NonCPS` method.

Possibly there is some way to simulate the error in the middle of a CPS-transformed method, perhaps by setting up two `parallel` branches and having one run a step while the other is paused inside the transformed method, in which case you could prove that adding `@NonCPS` fixes the problem, but I doubt it: in that case if you added `@NonCPS` it would simply block the whole CPS VM thread until the method finishes.

Could also test that `@NonCPS` can be used to work around bugs such as JENKINS-27421 (above in the test file), but I am not sure how interesting this is.

PR #213, opened 2015-09-25
[JENKINS-30055] Handle a large number of steps in quick succession

jglick:
[JENKINS-30055](https://issues.jenkins-ci.org/browse/JENKINS-30055)

Performance is still not great (about a minute to run on my laptop); mostly it just seems to be waiting in disk I/O to write lots of little log files, which is unavoidable so long as we maintain separate logs per step ([JENKINS-30896](https://issues.jenkins-ci.org/browse/JENKINS-30896)).

@reviewbybees


jglick:
The build 1697 failure was actually against 815afe3. 1698 is building 6140881. The PR builder is just confused.

jglick:
1698 hung; cannot get a thread dump to tell if it was stuck in shutdown of a particular test or what.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/ScalabilityTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Rule public JenkinsRule r = new JenkinsRule();
+
+    @Issue("JENKINS-30055")
+    @Test public void manySteps() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "demo");
+        p.setDefinition(new CpsFlowDefinition("for (int i = 0; i < 10000; i++) {echo \"iteration #${i}\"}", true));
+        WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
+        r.assertLogContains("iteration #9876", b);
+    }
```
tfennelly: The problem, as I remember it, only became visible after a restart. In that case, wouldn't the test try reloading the persisted form of the `Run` (`WorkflowRun.onLoad etc`)?

aggregator/src/test/java/org/jenkinsci/plugins/workflow/ScalabilityTest.java
```diff
@@ -0,0 +1,47 @@
...
+public class ScalabilityTest {
+
+    @Rule public JenkinsRule r = new JenkinsRule();
+
+    @Issue("JENKINS-30055")
+    @Test public void manySteps() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "demo");
+        p.setDefinition(new CpsFlowDefinition("for (int i = 0; i < 10000; i++) {echo \"iteration #${i}\"}", true));
+        WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
+        r.assertLogContains("iteration #9876", b);
```
jtnord: can you add a comment for this seemingly random number (i.e. I presume it is not 9999 for some specifc reason that eludes me when looking at the code)

api/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
```diff
@@ -158,7 +158,25 @@ public String getUrl() throws IOException {
...
+    public void addListener(GraphListener listener) {
+        addListener(listener, false);
+    }
+
+    /**
+     * Add a listener to changes in the flow graph structure.
+     * @param listener a listener to add
+     * @param synchronous true to receive events immediately as they occur (you must be very careful not to acquire locks or block); false to receive notification as soon as possible
+     */
+    public /*abstract*/ void addListener(GraphListener listener, boolean synchronous) {
```
tfennelly: `synchronous` seems like something that should be configured on the `GraphListener` impl.

Yes, I see it's an `interface` => backward compatibility issues etc., but setting is as an arg on `addListener` smells bad to me.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -767,10 +786,31 @@ synchronized FlowHead getFirstHead() {
...
+        List<GraphListener> _listeners = synchronous ? synchronousListeners : listeners;
+        if (_listeners != null) {
+            Saveable s = Saveable.NOOP;
+            try {
+                Queue.Executable exec = owner.getExecutable();
+                if (exec instanceof Saveable) {
+                    s = (Saveable) exec;
+                }
+            } catch (IOException x) {
+                LOGGER.log(Level.WARNING, null, x);
```
tfennelly: Why not provide some info here in the log message?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -767,10 +786,31 @@ synchronized FlowHead getFirstHead() {
...
+                    for (GraphListener listener : _listeners) {
+                        listener.onNewHead(node);
+                    }
+                }
+            } finally {
+                try {
+                    bc.commit();
+                } catch (IOException x) {
+                    LOGGER.log(Level.WARNING, null, x);
+                }
```
tfennelly: So how are the 2 lists any different (`synchronousListeners` Vs `listeners`)? It doesn't see like they're being treated any differently here at least, no?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/FlowHead.java
```diff
@@ -119,7 +120,9 @@ void setNewHead(FlowNode v) {
             } else {
                 // in recovering from error and such situation, we sometimes need to grow the graph
                 // without running the program.
-                execution.notifyListeners(v);
+                // TODO can CpsThreadGroup.notifyNewHead be used instead?
+                execution.notifyListeners(Collections.singletonList(v), true);
+                execution.notifyListeners(Collections.singletonList(v), false);
```
tfennelly: Maybe add a comment here explaining what this is about coz it seems odd. Notifying the synchronous and "non-synchronous" listeners.

Something really does smell strange about all of this. I'm sure it's just that I don't fully understand the issue you're trying to work around.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/ScalabilityTest.java
jglick: Yeah just something close to 10000 but not a fencepost.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
jglick: I also considered adding a

```java
public interface GraphListener {
    // …
    interface Synchronous extends GraphListener {}
}
```

and wound up going up with a flag on the listener instead, but if you think this is more intuitive as a marker on the listener then it should be easy to change it that way.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: The stack trace will be displayed. There is not much more to say. No particular reason to expect this to be thrown.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: They are not being treated differently here. One or the other is selected according to the method parameter.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/FlowHead.java
jglick: This caller is rarely used, only for error handling. We need to notify _all_ listeners.

jglick:
> The problem, as I remember it, only became visible after a restart. In that case, shouldn't the test try reloading the persisted form of the `Run` (`WorkflowRun.onLoad` etc)?

Indeed I just noticed this aspect of your JENKINS-30055. Perhaps not related to JENKINS-30651 after all. Looking into it.

jtnord:
:bee: on wip as my comment is addressed.

jglick:
`SimpleXStreamFlowNodeStorage` leads to `StackOverflowError`s due to its craziness.

jglick:
Currently `InputStepRestartTest.oldFlow` is failing:

```
"Running CpsFlowExecution[Owner[p/2015-11-17_20-22-49:p #1]]":
SUREFIRE-859:   waiting to lock monitor 0x00007fa470009598 (object 0x00000007a0c5a440, a org.jenkinsci.plugins.workflow.job.WorkflowRun),
  which is held by "Handling POST /jenkins/job/p/1/input/D13c3f7baed576768b11a714ef4d90e2/submit from 127.0.0.1 : Jetty Thread Pool"
"Handling POST /jenkins/job/p/1/input/D13c3f7baed576768b11a714ef4d90e2/submit from 127.0.0.1 : Jetty Thread Pool":
SUREFIRE-859:   waiting to lock monitor 0x00007fa47c030018 (object 0x00000007a0c5caf8, a org.jenkinsci.plugins.workflow.cps.CpsFlowExecution),
  which is held by "Running CpsFlowExecution[Owner[p/2015-11-17_20-22-49:p #1]]"
"Running CpsFlowExecution[Owner[p/2015-11-17_20-22-49:p #1]]":
	at hudson.model.Run.save(Run.java:1897)
	- waiting to lock <0x00000007a0c5a440> (a org.jenkinsci.plugins.workflow.job.WorkflowRun)
	at hudson.BulkChange.commit(BulkChange.java:112)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.notifyListeners(CpsFlowExecution.java:827)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.notifyNewHead(CpsThreadGroup.java:323)
	at org.jenkinsci.plugins.workflow.cps.FlowHead.setNewHead(FlowHead.java:119)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.onProgramEnd(CpsFlowExecution.java:789)
	- locked <0x00000007a0c5caf8> (a org.jenkinsci.plugins.workflow.cps.CpsFlowExecution)
	at org.jenkinsci.plugins.workflow.cps.CpsThreadGroup.run(CpsThreadGroup.java:297)
	at …
"Handling POST /jenkins/job/p/1/input/D13c3f7baed576768b11a714ef4d90e2/submit from 127.0.0.1 : Jetty Thread Pool":
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution$ConverterImpl.marshal(CpsFlowExecution.java:892)
	- waiting to lock <0x00000007a0c5caf8> (a org.jenkinsci.plugins.workflow.cps.CpsFlowExecution)
	at hudson.util.XStream2$AssociatedConverterImpl.marshal(XStream2.java:358)
	at …
	at hudson.XmlFile.write(XmlFile.java:178)
	at hudson.model.Run.save(Run.java:1898)
	- locked <0x00000007a0c5a440> (a org.jenkinsci.plugins.workflow.job.WorkflowRun)
	at org.jenkinsci.plugins.workflow.support.steps.input.InputStepExecution.postSettlement(InputStepExecution.java:223)
	at org.jenkinsci.plugins.workflow.support.steps.input.InputStepExecution.proceed(InputStepExecution.java:163)
	at org.jenkinsci.plugins.workflow.support.steps.input.InputStepExecution.doProceed(InputStepExecution.java:142)
	at org.jenkinsci.plugins.workflow.support.steps.input.InputStepExecution.doSubmit(InputStepExecution.java:126)
	at …
```

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -391,24 +394,48 @@ private void copyLogs() {
         }
     }
 
+    // CacheBuilder would be more convenient, but apparently does not support null values. Memoizer does not support weak keys.
```
jglick: Could use [`Optional`](http://stackoverflow.com/a/8298578/12916) but seems like overkill.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
```diff
@@ -158,8 +158,14 @@ public String getUrl() throws IOException {
      */
     public abstract void interrupt(Result r, CauseOfInterruption... causes) throws IOException, InterruptedException;
 
+    /**
+     * Add a listener to changes in the flow graph structure.
+     * @param listener a listener to add
+     */
     public abstract void addListener(GraphListener listener);
 
+    public /*abstract*/ void removeListener(GraphListener listener) {}
```
oleg-nenashev: Needs Javadoc and "TODO: make abstract" for the next backward-incompatible version

api/src/main/java/org/jenkinsci/plugins/workflow/flow/GraphListener.java
```diff
@@ -38,4 +38,12 @@
      * One of the use cases of this listener is to persist the state of {@link FlowExecution}.
      */
     void onNewHead(FlowNode node);
+
+    /**
+     * Listener which should be notified of events immediately as they occur.
+     * You must be very careful not to acquire locks or block.
+     * If you do not implement this marker interface, you will receive notifications in batched deliveries.
+     */
+    interface Synchronous extends GraphListener {}
```
oleg-nenashev: Such update would make https://github.com/jenkinsci/workflow-plugin/pull/285 unreliable if I use this interface

oleg-nenashev: #285 also must be reworked if you want to make the behavior async by default

api/src/main/java/org/jenkinsci/plugins/workflow/graph/BlockEndNode.java
```diff
@@ -34,19 +37,29 @@
...
         this.start = start;
+        startId = start.getId();
     }
 
     public START getStartNode() {
+        if (start == null) {
+            try {
+                start = (START) getExecution().getNode(startId);
+            } catch (IOException x) {
+                Logger.getLogger(BlockEndNode.class.getName()).log(Level.WARNING, null, x);
```
oleg-nenashev: If <code>getExecution().getNode(startId)</code> fails, we may be spammed by warnings here if a code above is recurrent. Maybe makes sense to add a flag, which blocks logs after first failure(?)

api/src/main/java/org/jenkinsci/plugins/workflow/graph/BlockEndNode.java
```diff
@@ -34,19 +37,29 @@
...
+        startId = start.getId();
     }
 
     public BlockEndNode(FlowExecution exec, String id, START start, List<FlowNode> parents) {
         super(exec, id, parents);
         this.start = start;
+        startId = start.getId();
     }
 
     public START getStartNode() {
```
oleg-nenashev: CheckForNull

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -103,6 +115,16 @@ public final boolean isRunning() {
      * Returns a read-only view of parents.
      */
     public List<FlowNode> getParents() {
+        if (parents == null) {
+            parents = new ArrayList<FlowNode>(parentIds.size());
+            for (String id : parentIds) {
```
oleg-nenashev: This code is thread-unsafe. It's a part of the  public API, hence :bug: . Use a local variable or synchronize the method

api/src/main/java/org/jenkinsci/plugins/workflow/graph/ForkNode.java
```diff
@@ -27,8 +27,7 @@
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 
 /**
- * @author Kohsuke Kawaguchi
- * @author Jesse Glick
+ * @deprecated unused
  */
 public class ForkNode extends BlockStartNode {
```
oleg-nenashev: Also needs <code>@Deprecated</code> annotation

api/src/main/java/org/jenkinsci/plugins/workflow/graph/JoinNode.java
```diff
@@ -29,8 +29,7 @@
 import java.util.List;
 
 /**
- * @author Kohsuke Kawaguchi
- * @author Jesse Glick
+ * @deprecated unused
  */
 public class JoinNode extends BlockEndNode<ForkNode> {
```
oleg-nenashev: same as above

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
oleg-nenashev: What blocks from using a FillerObject instead of null?

api/src/main/java/org/jenkinsci/plugins/workflow/flow/GraphListener.java
jglick: Do not follow. The existing behavior was asynchronous listeners. I just added an option for synchronous listeners. From what I know of #285 it should continue to use asynchronous listeners so this is irrelevant.

api/src/main/java/org/jenkinsci/plugins/workflow/graph/BlockEndNode.java
jglick: If and when I see such a problem.

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
jglick: Should only be used during deserialization, not randomly during runtime. Synchronizing this seems more dangerous than not. Unsure if there is some practical way to make this safer.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
jglick: Hence my comment about `Optional`.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
jglick: There is no planned incompatible version.

amuniz: backward-what? There is no such word in Jenkins vocabulary :smiley_cat:

amuniz:
:bee:

api/src/main/java/org/jenkinsci/plugins/workflow/flow/GraphListener.java
oleg-nenashev: Those listener must be synchronous according to our use-case (node capturing), but it's another story. Does not block this PR in any case

oleg-nenashev:
LGTM :bee:

PR #214, opened 2015-09-28
Noting JENKINS-30680

harniman:


oleg-nenashev:
Something strange is happening in the commits, but :bee:

PR #211, opened 2015-09-23
[JENKINS-30086] Steps help and description clean up

amuniz:
Both related:
[JENKINS-30086](https://issues.jenkins-ci.org/browse/JENKINS-30086)

Work in progress:
- [X] Review display names.
- [X] General inline help.
- [X] Fields inline help (if not clear from the field name).
- [X] Include the step function name in the dropdown list (format: `stepname: Step Display name`).

@reviewbybees (esp. @estheryouhana)

PR #215, opened 2015-09-28
[JENKINS-30088] Clean up step display in logs

amuniz:
[JENKINS-30088](https://issues.jenkins-ci.org/browse/JENKINS-30088)
Note that the first point described in the JIRA issue is fixed in https://github.com/jenkinsci/workflow-plugin/pull/211 which is more focused on the Snippet Generator clean up.

A screenshot showing how the logs are printed given the following script:

```
node {
  echo 'Starting!'
  stage 'Checkout'
  git 'https://github.com/amuniz/maven-helloworld.git'
  stage 'Building'
  sh 'uname -a'
  ws('other-ws') {
    echo 'This is another workspace'
  }
  stage 'Release'
  echo 'Releasing...'
}
```

workflow-logs

I'm not an UX expert, so any comment on how _metadata_ log lines and the _rest of the build log_ could live better together will be welcome!

@reviewbybees

### DIFF
--- a/api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
+++ b/api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
@@ -141,6 +141,20 @@ public abstract class FlowNode extends Actionable implements Saveable {
         else            return getTypeDisplayName();
     }
 
+    public String getDisplayFunctionName() {
+        String functionName = getTypeFunctionName();
+        if (functionName == null) {
+            return getDisplayName();
+        } else {
+            LabelAction a = getAction(LabelAction.class);
+            if (a != null) {
+                return functionName + ": " + a.getDisplayName();
+            } else {
+                return functionName;
+            }
+        }
+    }
+
     /**
      * Returns colored orb that represents the current state of this node.
      *
@@ -161,6 +175,19 @@ public abstract class FlowNode extends Actionable implements Saveable {
      * This is used to implement {@link #getDisplayName()} as a fallback in case {@link LabelAction} doesnt exist.
      */
     protected abstract String getTypeDisplayName();
+
+    /**
+     * Gets the function name for this type of node.
+     *
+     * Note that this method should be abstract (suppossed to be implemented in all subclasses), but keeping 
+     * it non-abstract to avoid binary incompatibilities.
+     *
+     * @return the function name or null
+     */
+    @CheckForNull
+    protected /* abstract */ String getTypeFunctionName() {
+        return null;
+    }
 
     /**
      * Returns the URL of this {@link FlowNode}, relative to the context root of Jenkins.
--- a/api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
+++ b/api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
@@ -42,6 +42,7 @@ import javax.annotation.Nonnull;
 import org.jenkinsci.plugins.workflow.actions.ErrorAction;
 import org.jenkinsci.plugins.workflow.actions.LabelAction;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
+import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
 import org.kohsuke.stapler.export.Exported;
@@ -177,16 +178,19 @@ public abstract class FlowNode extends Actionable implements Saveable {
     protected abstract String getTypeDisplayName();
 
     /**
-     * Gets the function name for this type of node.
+     * Gets a human readable text that may include a {@link StepDescriptor#getFunctionName()}.
+     * It would return "echo" for a flow node linked to an EchoStep or "ws {" for WorkspaceStep.
      *
-     * Note that this method should be abstract (suppossed to be implemented in all subclasses), but keeping 
+     * For StepEndNode it would return "} // step.getFunctionName()".
+     *
+     * Note that this method should be abstract (supposed to be implemented in all subclasses), but keeping
      * it non-abstract to avoid binary incompatibilities.
      *
-     * @return the function name or null
+     * @return the text human-readable representation of the step function name
+     *      or {@link FlowNode#getDisplayName()} by default (if not overriden in subclasses)
      */
-    @CheckForNull
     protected /* abstract */ String getTypeFunctionName() {
-        return null;
+        return getDisplayName();
     }
 
     /**
--- a/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepEndNode.java
+++ b/cps/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepEndNode.java
@@ -38,6 +38,17 @@ public class StepEndNode extends BlockEndNode<StepStartNode> implements StepNode
     }
 
     @Override
+    protected String getTypeFunctionName() {
+        StepDescriptor d = getDescriptor();
+        boolean isBody = getStartNode().isBody();
+        if (isBody) {
+            return "} //" + (d != null ? d.getFunctionName() : getStartNode().getStepName());
+        } else {
+            return getStartNode().getStepName() + " : End";
+        }
+    }
+
+    @Override
     public StepDescriptor getDescriptor() {
         return getStartNode().getDescriptor();
     }
--- a/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -88,6 +88,7 @@ import org.jenkinsci.plugins.workflow.flow.StashManager;
 import org.jenkinsci.plugins.workflow.graph.BlockEndNode;
 import org.jenkinsci.plugins.workflow.graph.FlowEndNode;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
+import org.jenkinsci.plugins.workflow.job.console.WorkflowConsoleLogger;
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 import org.kohsuke.stapler.HttpResponse;
 import org.kohsuke.stapler.HttpResponses;
@@ -656,7 +657,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements Q
             }
             node.addAction(new TimingAction());
 
-            logNodeMessage(node, "Running: " + node.getDisplayFunctionName());
+            logNodeMessage(node);
             if (node instanceof FlowEndNode) {
                 finish(((FlowEndNode) node).getResult(), execution.getCauseOfFailure());
             } else {
@@ -669,19 +670,19 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements Q
         }
     }
 
-    private void logNodeMessage(FlowNode node, String message) {
-        PrintStream logger = listener.getLogger();
+    private void logNodeMessage(FlowNode node) {
+        WorkflowConsoleLogger wfLogger = new WorkflowConsoleLogger(listener);
         String prefix = getLogPrefix(node);
         if (prefix != null) {
-            logger.printf("[%s] %s%n", prefix, message);
+            wfLogger.log(String.format("[%s] %s", prefix, node.getDisplayFunctionName()));
         } else {
-            logger.println(message);
+            wfLogger.log(node.getDisplayFunctionName());
         }
         // Flushing to keep logs printed in order as much as possible. The copyLogs method uses
         // LargeText and possibly LogLinePrefixOutputFilter. Both of these buffer and flush, causing strange
         // out of sequence writes to the underlying log stream (and => things being printed out of sequence)
         // if we don't flush the logger here.
-        logger.flush();
+        wfLogger.getLogger().flush();
     }
 
     static void alias() {
--- a/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -656,7 +656,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements Q
             }
             node.addAction(new TimingAction());
 
-            logNodeMessage(node, "Running: " + node.getDisplayName());
+            logNodeMessage(node, "Running: " + node.getDisplayFunctionName());
             if (node instanceof FlowEndNode) {
                 finish(((FlowEndNode) node).getResult(), execution.getCauseOfFailure());
             } else {
--- a/job/src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowMetadataConsoleFilter.java
+++ b/job/src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowMetadataConsoleFilter.java
@@ -24,39 +24,32 @@
 
 package org.jenkinsci.plugins.workflow.job.console;
 
-import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import java.io.IOException;
+import java.io.OutputStream;
 
-import hudson.MarkupText;
-import hudson.console.ConsoleAnnotator;
-import hudson.console.ConsoleNote;
-import hudson.model.Run;
+import hudson.console.LineTransformationOutputStream;
 
 /**
- * Console note for Workflow metadata specific messages.
- * See {@link WorkflowConsoleLogger} for more information.
+ * It transforms workflow metadata log messages through {@link WorkflowRunConsoleNote}.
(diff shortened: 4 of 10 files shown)


job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -656,7 +657,7 @@ public int hashCode() {
             }
             node.addAction(new TimingAction());
 
-            logNodeMessage(node, "Running: " + node.getDisplayName());
+            logNodeMessage(node);
```
oleg-nenashev: Hm? Seems it deletes the "Running" text prefix. +1 from me for annotated logs, but I'm not sure how good it will appear in plain text

recena:
@amuniz Have you considered something like that?

```
[Workflow::node]
[Workflow::echo] 
[Workflow::sh] 
```

amuniz:
@recena I followed the proposal described in [JENKINS-30088](https://issues.jenkins-ci.org/browse/JENKINS-30088). I personally like the current format, since there is a fixed length prefix that allows to easily identify metadata related log lines.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
amuniz: Yeah, I see the "Running:" prefix at the beginning of each metadata line as a way to easily identify metadata logs, but it is redundant now since we have the "[Workflow]" prefix, and it is quite obvious that a step is "Running" when you see something like `[Workflow] node {` or `[Workflow] echo`.

The plain text version would be something like this (where is still easy to identify the metadata log lines IMO):

```
Started by user anonymous
[Workflow] Allocate node : Start
Running on master in /home/amuniz/workspace-cloudbees/workflow-plugin/aggregator/work/jobs/test/workspace
[Workflow] node {
[Workflow] echo
Hi!
[Workflow] git
 > git rev-parse --is-inside-work-tree # timeout=10
Fetching changes from the remote Git repository
 > git config remote.origin.url https://github.com/amuniz/maven-helloworld.git # timeout=10
Fetching upstream changes from https://github.com/amuniz/maven-helloworld.git
 > git --version # timeout=10
 > git fetch --tags --progress https://github.com/amuniz/maven-helloworld.git +refs/heads/*:refs/remotes/origin/*
 > git rev-parse refs/remotes/origin/master^{commit} # timeout=10
 > git rev-parse refs/remotes/origin/origin/master^{commit} # timeout=10
Checking out Revision 197d621fea6a797cbb4a158df573580a83fd8b4f (refs/remotes/origin/master)
 > git config core.sparsecheckout # timeout=10
 > git checkout -f 197d621fea6a797cbb4a158df573580a83fd8b4f
 > git rev-list 197d621fea6a797cbb4a158df573580a83fd8b4f # timeout=10
[Workflow] stage: After Git Clone
Entering stage After Git Clone
Proceeding
[Workflow] echo
Bye!
[Workflow] sh
[workspace] Running shell script
 mvn clean install
[INFO] Scanning for projects...
[INFO]                                                                         
[INFO] ------------------------------------------------------------------------
[INFO] Building example 1.0-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[INFO] 
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ example ---
...
```

jtnord:
**disclaimer**: I have not read the dull PR just skimmed though a few comments and the unit tests.

So a big thing is that the logs are cluttered and contain redundant (to me) information and miss some vital information.

```
[Workflow] Allocate node : Start
Running on master in /home/amuniz/workspace-cloudbees/workflow-plugin/aggregator/work/jobs/test/workspace
[Workflow] node {
```

So if you have `node {` you do not need `[Workflow] Allocate node : Start`  you only need one (clutter)

```
[Workflow] Retry the body up to N times : Start",
[Workflow] retry {
/// some more logs
[Workflow] } //retry
[Workflow] retry {
/// some more logs
[Workflow] } //retry
[Workflow] retry {
/// some more logs
[Workflow] } //retry
```

`retry {` is not really valid syntax for the inner loop.  It is missing the important information which is Which loop is this - the first or the 12th or the 100th?  How many tries are remaining?  (incorrect information and missing information)

amuniz:
> So if you have `node {` you do not need `[Workflow] Allocate node : Start` you only need one (clutter)

Both log lines were there before, I just replaced `Allocate node : Start : Body` with `node {`. If we want to remove the first message, it's ok for me, but it's not what was defined in the JIRA issue.

> `retry {` is not really valid syntax for the inner loop.

Right. This a general log representation where every node with an execution body is represented as `node_name {` and `} // node_name`. It's not intended to be syntactically correct but useful when reading logs.

> Which loop is this - the first or the 12th or the 100th?

Then the logging inside `RetryStepExecution` must be improved. The retry counter is not visible outside the execution class (and it shouldn't be), where this _metadata_ logs are managed.

jtnord:
So i think then i am saying the jira is off the mark.

To focus on the dsl the cruft should be removed so you have dsly stuff, stuff that is important in the context of the step and the outputs if any of the step.  If you leave the clutter in then the dslish does not have as much use.  If you use the dsl when it is not the dsl (retry loops) then this makes thing far worse for understandability.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
oleg-nenashev: Yes, it's much better

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepEndNode.java
```diff
@@ -38,6 +38,17 @@ protected String getTypeDisplayName() {
     }
 
     @Override
+    protected String getTypeFunctionName() {
+        StepDescriptor d = getDescriptor();
+        boolean isBody = getStartNode().isBody();
+        if (isBody) {
+            return "} //" + (d != null ? d.getFunctionName() : getStartNode().getStepName());
```
tfennelly: Is there an opening curly brace somewhere else?

tfennelly: Ah, I see it below ... ignore :)

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepEndNode.java
```diff
@@ -38,6 +38,17 @@ protected String getTypeDisplayName() {
...
+        } else {
+            return getStartNode().getStepName() + " : End";
```
tfennelly: I assume there's no overhead to `getStartNode`? It's called 3 times here in this method. I only ask because this flow node stuff is mad and it would not surprise me if `getStartNode` involved some sort of convoluted graph traversal.

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
tfennelly: Looks great ... a lot less noise.

job/src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowMetadataConsoleFilter.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private final OutputStream out;
+
+    public WorkflowMetadataConsoleFilter(OutputStream out) {
+        this.out = out;
+    }
+
+    @Override
+    protected void eol(byte[] b, int len) throws IOException {
+        new WorkflowRunConsoleNote().encodeTo(out);
```
tfennelly: Do you need to create a new `WorkflowRunConsoleNote` for each line? Would one per `WorkflowMetadataConsoleFilter` not work?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/nodes/StepEndNode.java
amuniz: No, it's [returning a class field directly](https://github.com/jenkinsci/workflow-plugin/blob/c2afa61d28b2922df955bf7f6b45e456e898c10c/api/src/main/java/org/jenkinsci/plugins/workflow/graph/BlockEndNode.java#L50).

job/src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowRunConsoleNote.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * CSS color selector.
+     */
+    private static final String TEXT_COLOR = "9A9999";
+
+    @Override
+    public ConsoleAnnotator<Run<?,?>> annotate(Run<?, ?> context, MarkupText text, int charPos) {
+        if (context instanceof WorkflowRun) {
+            if (text.getText().startsWith(CONSOLE_NOTE_PREFIX)) {
+                text.addMarkup(0, text.length(), "<span style=\"color:#"+ TEXT_COLOR +"\">", "</span>");
```
tfennelly: Maybe I'm not seeing it correctly but it looks to me like the 3rd arg in the call to `addMarkup` could be statically defined, saving some string manipulation for every call to `annotate`. I'm assuming these log related methods might get called a lot .... if not then ignore :)

tfennelly:
:bee: insofar as I get it .... kinda skimmed through it.

job/src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowMetadataConsoleFilter.java
amuniz: Well, the note object is directly serialized in the log so I guess the must be different objects, but I'm not sure, perhaps someone with a deeper knowledge of the `ConsoleNote` API could tell us?

BTW this is the pattern I've seen everywhere when using console notes (for example [here](https://github.com/jenkinsci/openJDK-native-plugin/blob/4ef6b93b56b3d7ca896c787aad04671fd6d1c87d/src/main/java/org/jenkinsci/plugins/openjdk_native/OpenJDKConsoleAnnotator.java#L23), [here](https://github.com/jenkinsci/grails-plugin/blob/8e1aafa0482272d981f681e7eec8d82221ecfac0/src/main/java/com/g2one/hudson/grails/GrailsBuilder.java#L264) or [here](https://github.com/jenkinsci/docker-build-step-plugin/blob/3c2295f3a54c6fecda4e705bd0b87c108f015993/src/main/java/org/jenkinsci/plugins/dockerbuildstep/log/DockerConsoleAnnotator.java#L31)).

job/src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowRunConsoleNote.java
amuniz: Fixed Tom!

tfennelly:
:bee:

jglick:
> If we want to remove the first message

I think we do, if possible, but at least for the case of `retry` this probably needs to wait for a fix of [JENKINS-26156](https://issues.jenkins-ci.org/browse/JENKINS-26156).

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -163,6 +177,19 @@ public BallColor getIconColor() {
     protected abstract String getTypeDisplayName();
 
     /**
+     * Gets the function name for this type of node.
+     *
+     * Note that this method should be abstract (suppossed to be implemented in all subclasses), but keeping
```
jglick: typo

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -163,6 +177,19 @@ public BallColor getIconColor() {
     protected abstract String getTypeDisplayName();
 
     /**
+     * Gets the function name for this type of node.
```
jglick: Pretty vague Javadoc. I would have some idea what you meant, but be wrong, unless I were actually reading the implementations. This is in general not a “function name” at all, it is just a block of human-readable text that may or may not include a `StepDescriptor.functionName`.

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -141,6 +141,20 @@ public String getDisplayName() {
         else            return getTypeDisplayName();
     }
 
+    public String getDisplayFunctionName() {
+        String functionName = getTypeFunctionName();
+        if (functionName == null) {
+            return getDisplayName();
```
jglick: Why not make `getTypeFunctionName` return `@Nonnull` and default to `getDisplayName`?

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -141,6 +141,20 @@ public String getDisplayName() {
         else            return getTypeDisplayName();
```
jglick: Can we deprecate `getDisplayName`? If so, should the `@Exported` annotation be moved to another method (with `name="displayName"` added)?

job/src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowRunConsoleNote.java
```diff
@@ -0,0 +1,65 @@
...
+     * Prefix used in metadata lines.
+     */
+    public static final String CONSOLE_NOTE_PREFIX = "[Workflow] ";
+
+    /**
+     * CSS color selector.
+     */
+    private static final String TEXT_COLOR = "9A9999";
+
+    private static final String START_NOTE = "<span style=\"color:#"+ TEXT_COLOR +"\">";
```
jglick: Hmm, I wonder if there is no way to specify inline CSS when the first such note appears in a given build log.

jglick:
Minor comments but :bee: if you want to stop here.

As noted earlier, we can improve the display of block steps later.

jglick:
BTW please include a note in `CHANGES.md`.

job/src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowRunConsoleNote.java
recena: Indeed, this kind of things should not be hardcoded here.

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
amuniz: Well, `getDisplayName()` is overriding `ModelObject.getDisplayName()`, it's hard to say that it's not actually being used somewhere when accessing the flow node as a model object. Do you think it is safe to deprecate it?

job/src/main/java/org/jenkinsci/plugins/workflow/job/console/WorkflowRunConsoleNote.java
amuniz: Not sure about your proposal, could you elaborate please?

recena: @amuniz I need to check one thing before. Give me some time. Obviously, go forward with this PR.

amuniz:
@jglick I fixed some things. I can improve the use console note CSS later (and once I know how).

tfennelly:
:bee: go for it @amuniz

amuniz:
@reviewbybees done

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
jglick: Half-addressed. You made it return `getDisplayName` as a default, but did not take advantage of that to simplify the code here.

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
jglick: If it is an override then we cannot deprecate it. We *could* consider moving the REST-exported property elsewhere, or just leave it, although it is kind of useless.

api/src/main/java/org/jenkinsci/plugins/workflow/graph/FlowNode.java
```diff
@@ -163,6 +178,22 @@ public BallColor getIconColor() {
...
+     *
+     * For StepEndNode it would return "} // step.getFunctionName()".
+     *
+     * Note that this method should be abstract (supposed to be implemented in all subclasses), but keeping
+     * it non-abstract to avoid binary incompatibilities.
+     *
+     * @return the text human-readable representation of the step function name
+     *      or {@link FlowNode#getDisplayName()} by default (if not overriden in subclasses)
+     */
+    protected /* abstract */ String getTypeFunctionName() {
```
jglick: Still seems like a poor name choice, since it is only a function name in the case of `StepAtomNode`s.

PR #218, opened 2015-10-01
Add build number link info in BuildTriggerStepExecution

wutingbupt:
Hi Jenkins team,

This commit will help improve log message for Build Trigger Step.

Current log: Starting building project: test
This makes us very hard to know which build is been triggered, we have to go to the job page check it carefully.

After change: there will have a link point out the build page, just like current build flow plugin.
demo

Right now, this patch will fail the jenkins unit test, the root cause is one of test case:
BuildTriggerStepRestartTest.
Inside this testcase, it will set executor to 0, and start up a free style job inside a build trigger step, in my fix, I add a sentence like this to get Run instance,          
Run r = (Run) f.waitForStart();

Due to there is no executor, sub job inside build step will hanging there, however, in my fix, I want to wait for it start, so that I can get the build number, this situation cause unit test hanging there until Jenkins abort it.

Let me know if you have better idea to fix this issue.

Great thanks for your help.

Br,
Tim


AnalogJ:
Yeah, I would love to see this PR merged. This is definitely a small but common annoyance with the workflow plugin vs the buildflow plugin.

jglick:
The request is valid but the patch is unacceptable, because it makes a blocking call during `StepExecution.start`. In general, it could take days for the submitted build to start, or it might never start. Rather the existing `start` method should be left alone (or it can hyperlink to the _job_), with an additional message printed if and when the downstream build actually begins.

wutingbupt:
Hi Jenkins team,

Based on my limited knowledge about Jenkins, I should use a listener to monitor the job, I try to add a listener like the following code inside step execution class, however, I found it is very hard to get the TaskListener instance to print out the log. 

I try use this way to create a new one,                         
StreamTaskListener listener2 = new StreamTaskListener(upStreamRun.getLogFile());

However, during test, I find it may lost some logs if we trigger a job without wait process.

Let me know if there is better idea for this case.

    @Extension
    public static final class Listener extends RunListener<Run<?,?>> {
        @Override public void onStarted(Run<?,?> r, TaskListener listener) {
            Cause.UpstreamCause upstreamCause = r.getCause(Cause.UpstreamCause.class);

            if (upstreamCause != null){
                Run upStreamRun = upstreamCause.getUpstreamRun();
                if (upStreamRun.isBuilding()){ //TODO Also need to check if upStream job is a work flow job.
                    try {
                        
                        StreamTaskListener listener2 = new StreamTaskListener(upStreamRun.getLogFile());
                        listener2.getLogger().println("!#####!!!Starting building project: " + HyperlinkNote.encodeTo('/' + r.getUrl(), r.getFullDisplayName()));
                        listener2.close();
                    } catch (IOException e) {
                        e.printStackTrace();
                    }


                }
            }
        }
    }

jglick:
Add it to `BuildTriggerListener`, look for `BuildTriggerAction`, use `action.getStepContext().get(TaskListener.class)`.

wutingbupt:
Hi,

Thanks for your suggestion, good to know the Jenkins magic world.

You can find how it looks like from this picture.

test_result

jglick:
Looks right from a five-second glance, let me get back to it soon.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
```diff
@@ -29,4 +35,24 @@ public void onDeleted(Run<?,?> run) {
...
+            Cause.UpstreamCause upstreamCause = run.getCause(Cause.UpstreamCause.class);
+            if (upstreamCause != null) {
+                Run upStreamRun = upstreamCause.getUpstreamRun();
+                BuildTriggerAction buildTriggerAction = run.getAction(BuildTriggerAction.class);
+                //Only works when upstream is still running
+                if (upStreamRun != null && upStreamRun.isBuilding() && buildTriggerAction != null) {
+                    try {
+                        TaskListener taskListener = buildTriggerAction.getStepContext().get(TaskListener.class);
+                        taskListener.getLogger().println("Starting building project: " + HyperlinkNote.encodeTo('/' + run.getUrl(), run.getFullDisplayName()));
+                    } catch (Exception e) {
```
KostyaSha: Why `onStarted` doesn't allow throw exceptions? `fireStarted` has
```
                try {
                    l.onStarted(r,listener);
                } catch (Throwable e) {
                    report(e);
                }

    private static void report(Throwable e) {
        LOGGER.log(Level.WARNING, "RunListener failed",e);
    }
```

wutingbupt: @KostyaSha, based on the javadoc, there is no big diff actually, however, current way will print error log in the main work flow page,  if we directly throw it out, it will be just in somewhere depends on log4j configuration not in current log page.

KostyaSha: Good point, then `report(e)` should also print to listener

wutingbupt:
Hm... you are right, forget why I put a logger there, let's see what is the reply from @jglick

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
jglick: @KostyaSha `onStarted` is not allowed to throw any exceptions. The `catch` block is there to recover gracefully from buggy plugins which throw `RuntimeException`s (or even `Error`s)—we want the event to still be delivered to unrelated plugins.

jglick:
Please do not `chmod a+x` files in the source tree you edit. Revert the mode change.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
```diff
@@ -29,4 +35,24 @@ public void onDeleted(Run<?,?> run) {
         }
     }
 
+    @Extension
+    public static final class Listener extends RunListener<Run<?, ?>> {
```
jglick: Why a separate listener? You can just put `onStarted` directly in `BuildTriggerListener`.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
```diff
@@ -29,4 +35,24 @@ public void onDeleted(Run<?,?> run) {
...
+        @Override
+        public void onStarted(Run<?, ?> run, TaskListener listener) {
+            Cause.UpstreamCause upstreamCause = run.getCause(Cause.UpstreamCause.class);
+            if (upstreamCause != null) {
+                Run upStreamRun = upstreamCause.getUpstreamRun();
+                BuildTriggerAction buildTriggerAction = run.getAction(BuildTriggerAction.class);
+                //Only works when upstream is still running
+                if (upStreamRun != null && upStreamRun.isBuilding() && buildTriggerAction != null) {
```
jglick: I do not think you need to bother checking for `upstreamCause`/`upStreamRun` at all. If the upstream build has already finished, the `TaskListener` from its `StepContext` should just be ignored. You can call `StepContext.isReady` I suppose.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerListener.java
KostyaSha: It not allowed but catches, any operation on listener may end with IOException, so handling it in implementation doesn't look like a standard case. My idea that onStarted should allow throwing exceptions and handle it outside implementation by providing normal logging/description to listener/logger.

KostyaSha: Adding `throws Exception` will be backward compatible and will allow exclude such handlings in implementations. That will provide easier code in Listener implementations (+ add error print to listener). WDYT?

wutingbupt:
@jglick 
Thanks for your comments, I will check it and back to you.

Br,
Tim

wutingbupt:
Hi,

Let me know if there is any problem for the new commit.

Thanks.

Br,
Tim

PR #219, opened 2015-10-06
Improve getBuildVariables method inside RunWrapper so that we can get…

wutingbupt:
Hi Jenkins team,

As a user, I would like to get current work flow job parameters, however, it is not supported right now. You have to do some tracks if you really want to get the job parameter.

My fix will improve this situation.

Br,
Tim


jglick:
Parameters can [just be accessed as variables](https://github.com/jenkinsci/workflow-plugin/blob/master/TUTORIAL.md#build-parameters). Not sure what is missing.

wutingbupt:
Hi,

Thanks for the info, sorry for missing important wiki!

Br,
Tim

I try to trigger sub-work-flow inside main-flow, after that, I want to get the parameter back

PR #220, opened 2015-10-08
WorkflowMultiBranchProjectFactory

jglick:
Downstream of https://github.com/jenkinsci/branch-api-plugin/pull/8.

- [X] introduce `WorkflowMultiBranchProjectFactory`
- [x] deduplicate a few things

@reviewbybees esp. @stephenc

### DIFF
--- a/multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectFactory.java
+++ b/multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectFactory.java
@@ -34,7 +34,7 @@ import jenkins.branch.MultiBranchProjectFactory;
 import jenkins.branch.MultiBranchProjectFactoryDescriptor;
 import jenkins.scm.api.SCMSource;
 import jenkins.scm.api.SCMSourceCriteria;
-import org.jenkinsci.plugins.workflow.multibranch.WorkflowMultiBranchProject;
+import static org.jenkinsci.plugins.workflow.multibranch.WorkflowMultiBranchProject.CRITERIA;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 public class WorkflowMultiBranchProjectFactory extends MultiBranchProjectFactory.BySCMSourceCriteria {
@@ -44,12 +44,7 @@ public class WorkflowMultiBranchProjectFactory extends MultiBranchProjectFactory
     @DataBoundConstructor public WorkflowMultiBranchProjectFactory() {}
 
     @Override protected SCMSourceCriteria getSCMSourceCriteria(SCMSource source) {
-        // TODO share implementation with WorkflowMultiBranchProject
-        return new SCMSourceCriteria() {
-            @Override public boolean isHead(SCMSourceCriteria.Probe probe, TaskListener listener) throws IOException {
-                return probe.exists(JENKINSFILE);
-            }
-        };
+        return CRITERIA;
     }
 
     @Override protected MultiBranchProject<?,?> doCreateProject(ItemGroup<?> parent, String name, Map<String,Object> attributes) {
--- a/multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectFactory.java
+++ b/multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectFactory.java
@@ -1,0 +1,70 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.plugins.workflow.multibranch;
+
+import hudson.Extension;
+import hudson.model.ItemGroup;
+import hudson.model.TaskListener;
+import java.io.IOException;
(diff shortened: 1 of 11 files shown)


jglick:
BTW PR builds are useless here since `multibranch` is not listed in the parent’s `<modules>`, so the CI job does not try to build it. It cannot simply be added without then participating in regular WF releases, which I am trying to avoid since it cannot use the same version numbers are regular releases, because it has `-beta` dependencies, and the UC does not handle that well. Could be in a separate repo, but it seems pretty closely aligned with core WF. IIUC you can add `<modules>` in a `<profile>`, but what I want is a profile which is activated by default _except_ when releasing, and not sure how to do that, except via [this hack](http://stackoverflow.com/a/31495604/12916) which seems inappropriate since it pretends to be creating releases which never actually get uploaded (and I _do_ want to create `workflow-multibranch` beta releases, just on a different timeline). So a bit of a mess at the moment.

amuniz:
What about using two profiles? `default` which would be active by default (`<activeByDefault>true</activeByDefault>`) and `release` active when running release plugin (making use of [releaseProfiles](http://maven.apache.org/maven-release/maven-release-plugin/perform-mojo.html#releaseProfiles)).

amuniz:
Or better, activate [useReleaseProfile](http://maven.apache.org/maven-release/maven-release-plugin/perform-mojo.html#useReleaseProfile) and configure `release-profile` profile locally.

jglick:
> What about using two profiles?

The trouble is that then _both_ will be active during releases, and I think modules cannot be _subtracted_ by a profile.

amuniz:
>  both will be active during releases

Hmm, not sure, I think that any `activeByDefault` is deactivated when another profile is explicitly activated.

jglick:
Does sound like that is true. Would need to play with it a bit.

stephenc:
@jglick have a look at oc-all, we have modules there that are active for a full build but excluded from a release (mind you we are using additional scripting to help, but the principle is sound)

stephenc:
:bee: (and I like the test harness, seems like it might be useful in other multi-branch factories too)

jglick:
> seems like it might be useful in other multi-branch factories too

BTW `yaml-project` has a lot of tests, but it seems like they [stop short of actually testing multibranch indexing & builds](https://github.com/jenkinsci/yaml-project-plugin/blob/83a0d05bc7eb145190c4d25e4284adfaf365e334/src/test/java/com/google/jenkins/plugins/dsl/YamlMultiBranchProjectTest.java#L80-L81).

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/GitDirectorySCMNavigator.java
```diff
@@ -0,0 +1,109 @@
...
+        if (kids == null) {
+            listener.error(directory + " does not exist");
+            return;
+        }
+        for (File kid : kids) {
+            if (!new File(kid, ".git").isDirectory()) {
+                listener.getLogger().format("Ignoring %s since it does not look like a Git checkout%n", kid);
+                continue;
+            }
+            listener.getLogger().format("Considering: %s%n", kid);
```
oleg-nenashev: The code above could be replaced by https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/FileUtils.html#iterateFiles(java.io.File,%20org.apache.commons.io.filefilter.IOFileFilter,%20org.apache.commons.io.filefilter.IOFileFilter)

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/GitDirectorySCMNavigator.java
```diff
@@ -0,0 +1,109 @@
...
+                continue;
+            }
+            String origin = kid.getAbsolutePath(); // fallback
+            for (String line : IOUtils.readLines(new ByteArrayInputStream(baos.toByteArray()))) {
+                Matcher m = ORIGIN.matcher(line);
+                if (m.matches()) {
+                    origin = m.group(1);
+                    listener.getLogger().format("Found origin URL: %s%n", origin);
+                    if (origin.startsWith("git@")) {
+                        listener.getLogger().format("Warning: SSH private keys not yet supported here%n");
```
oleg-nenashev: :bug:  You may need to check other users. Otherwise the code may blow up on private instances

oleg-nenashev:
:bug: for handling a "git@" prefix to check the connection type. It's not reliable approach. See https://www.kernel.org/pub/software/scm/git/docs/git-push.html#URLS

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/GitDirectorySCMNavigator.java
jglick: This is in test code, not production code.

oleg-nenashev:
Didn't notice the "test" entry in the path. Tests are just tests => the issue is NIT
:bee:

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/GitDirectorySCMNavigator.java
jglick: Without lambdas, that just seems like it would be more verbose.

amuniz:
:bee:

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectFactoryTest.java
```diff
@@ -0,0 +1,116 @@
...
+        // Verify permissions:
+        Authentication admin = User.get("admin").impersonate();
+        ACL acl = one.getACL();
+        assertTrue(acl.hasPermission(ACL.SYSTEM, Item.CONFIGURE));
+        assertTrue(acl.hasPermission(ACL.SYSTEM, Item.DELETE));
+        assertFalse(acl.hasPermission(admin, Item.CONFIGURE));
+        assertFalse(acl.hasPermission(admin, Item.DELETE));
+        assertTrue(acl.hasPermission(admin, Item.EXTENDED_READ));
+        assertTrue(acl.hasPermission(admin, Item.READ));
+        r.waitUntilNoActivity();
```
amuniz: Why is this needed?

amuniz: Nothing is running asynchronously in background, right?

tfennelly:
:bee:

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectFactory.java
```diff
@@ -0,0 +1,66 @@
...
+import jenkins.branch.MultiBranchProjectFactory;
+import jenkins.branch.MultiBranchProjectFactoryDescriptor;
+import jenkins.scm.api.SCMSource;
+import jenkins.scm.api.SCMSourceCriteria;
+import static org.jenkinsci.plugins.workflow.multibranch.WorkflowMultiBranchProject.CRITERIA;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+public class WorkflowMultiBranchProjectFactory extends MultiBranchProjectFactory.BySCMSourceCriteria {
+
+    public static final String JENKINSFILE = "Jenkinsfile";
```
stephenc: Unused?

stephenc:
:bee:

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectFactory.java
jglick: Yes, good catch.

stephenc:
:bee:

amuniz:
:bee:

amuniz:
:bee:

oleg-nenashev:
:bee: and :beers:

PR #221, opened 2015-10-14
[JENKINS-29542] Escaping '$' to be not parsed in jelly help page

rodrigozrusso:


cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/EnvActionImpl/Binder/help.jelly
```diff
@@ -41,7 +41,7 @@ node {
     </p>
     <p>
         <pre>mail to: 'devops@acme.com',
-    subject: "Job '${env.JOB_NAME}' (${env.BUILD_NUMBER}) is waiting for input",
+    subject: "Job '$${env.JOB_NAME}' ($${env.BUILD_NUMBER}) is waiting for input",
     body: "Please go to ${env.BUILD_URL} and verify the build"</pre>
```
jglick: What about this line?

mrroohian:
Without this change ${env.JOB_NAME} and ${env.BUILD_NUMBER} results in empty string in the snippet generator help section:

mail to: 'devops@acme.com',
    subject: "Job '' () is waiting for input",
    body: "Please go to  and verify the build"

mroohian:
Without this change ${env.JOB_NAME} and ${env.BUILD_NUMBER} results in empty string in the snippet generator help section:

mail to: 'devops@acme.com',
subject: "Job '' () is waiting for input",
body: "Please go to and verify the build"

jglick:
> body: "Please go to and verify the build"

So line 45 indeed needs to be patched too.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/EnvActionImpl/Binder/help.jelly
KostyaSha: seems jelly as xml escapes this special char.

PR #225, opened 2015-10-20
Added absoluteUrl property to RunWrapper

rsandell:
@reviewbybees

### DIFF
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -36,7 +36,6 @@ import java.util.Map;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 
-import jenkins.model.Jenkins;
 import org.acegisecurity.context.SecurityContext;
 import org.acegisecurity.context.SecurityContextHolder;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.Whitelisted;
@@ -172,6 +171,7 @@ public final class RunWrapper implements Serializable {
         }
     }
 
+    @SuppressWarnings("deprecation")
     @Whitelisted
     public @Nonnull String getAbsoluteUrl() throws AbortException {
         return build().getAbsoluteUrl();
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -35,6 +35,8 @@ import java.util.Collections;
 import java.util.Map;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+
+import jenkins.model.Jenkins;
 import org.acegisecurity.context.SecurityContext;
 import org.acegisecurity.context.SecurityContextHolder;
 import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.Whitelisted;
@@ -167,6 +169,16 @@ public final class RunWrapper implements Serializable {
             } else { // who knows
                 return Collections.emptyMap();
             }
+        }
+    }
+
+    @Whitelisted
+    public @Nonnull String getAbsoluteUrl() throws AbortException {
+        String root = Jenkins.getActiveInstance().getRootUrl();
+        if (root == null) {
+            throw new AbortException("Jenkins root url is not configured!");
+        } else {
+            return root + build().getUrl();
         }
     }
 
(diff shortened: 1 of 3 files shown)


support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -170,4 +172,14 @@ public String getId() throws AbortException {
         }
     }
 
+    @Whitelisted
+    public @Nonnull String getAbsoluteUrl() throws AbortException {
+        String root = Jenkins.getActiveInstance().getRootUrl();
+        if (root == null) {
+            throw new AbortException("Jenkins root url is not configured!");
+        } else {
+            return root + build().getUrl();
```
jglick: :ant: this is slightly different from `Run.getAbsoluteUrl`. The handling of `Stapler.getCurrentRequest()` probably does not matter, because that would be null from the Workflow CPS VM thread, but the failure to call `Util.encode` might matter. Better to just use the standard method here.

amuniz:
Is not this already accesible through `getRawBuild().getAbsoluteUrl()`?

jglick:
@amuniz yes but you cannot use `.rawBuild.absoluteUrl` from a sandboxed script.

jglick:
:bee: despite the previous note.

amuniz:
@jglick ok, thanks.

:bee:

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
rsandell: ```Run.getAbsoluteUrl()``` is marked as deprecated from what I can see.

rsandell:
@reviewbybees done

amuniz:
@rsandell It's a _special_ deprecation just to avoid using it in jelly. Anyway, the restriction on sandboxed scripts is totally justifying this new API.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
amuniz: @rsandell It's a special deprecation just to avoid using it in jelly AFAIK. Anyway, the restriction on sandboxed scripts is totally justifying this new API.

jglick: Right, it is only deprecated as a way of getting people’s attention so they do not use it in HTML pages, which is a different use case.

jglick:
Again a :bee: if you wish this to be merged as is and either disagree with the :ant: or do not have time to work on it further.

rsandell:
Well I don't like using deprecated methods but if you insist I'm fine with changing it, so I'll do it now.

amuniz:
:bee:

amuniz:
I didn't insist, I was only suggesting :)

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -35,6 +35,8 @@
 import java.util.Map;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+
+import jenkins.model.Jenkins;
```
jglick: :ant: unused

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -170,4 +172,9 @@ public String getId() throws AbortException {
         }
     }
 
+    @Whitelisted
+    public @Nonnull String getAbsoluteUrl() throws AbortException {
```
jglick: :ant: actually it does not throw `AbortException` any longer

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -170,4 +172,9 @@ public String getId() throws AbortException {
...
+        return build().getAbsoluteUrl();
```
jglick: :ant:

```java
@SuppressWarnings("deprecated")
```

would be appropriate.

jglick:
:bee:

jglick:
Actually :bug: forgot to update `BuildTriggerStep/help-wait.html`.

BTW you can add an entry to `CHANGES.md`, or I will when merging.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
rsandell: the call to ```build()``` can throw ```AbortException```

jglick:
:bee:

tfennelly:
:bee:

oleg-nenashev:
:bee:

PR #227, opened 2015-10-23
Fix demo image assembly

tfennelly:
- [x] Update Jetty version. Original no longer available for download.
- [x] Set ownership on the jetty logs dir for user jenkins. Otherwise, results in write failures when demo is running.
- [x] Set ownership on the jetty/webapps dir for user jenkins.

I also see `hudson.model.Computer$TerminationRequest` exceptions while the build is running ([see gist](https://gist.github.com/tfennelly/252efe7f9d34b90e14b5) ). These same exceptions happen when I pull and run the latest published image (so is not a regression). The demo runs fine.

I did not update the demo to use v1.10.1 of the workflow plugins, or to use the 1.625.1 jenkins docker image, but I did change and test it locally and it worked fine.

### DIFF
--- a/demo/Dockerfile
+++ b/demo/Dockerfile
@@ -10,10 +10,11 @@ RUN ln -sv /usr/local/apache-maven-$MAVEN_VERSION /usr/local/maven
 # install Jetty
 WORKDIR /opt
 # jetty package is still 8
-ENV JETTY_VERSION 9.2.11.v20150529
+ENV JETTY_VERSION 9.2.12.v20150709
 RUN wget -O - "http://mirrors.ibiblio.org/eclipse/jetty/$JETTY_VERSION/dist/jetty-distribution-$JETTY_VERSION.tar.gz" | tar xvfz -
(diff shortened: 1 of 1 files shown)


tfennelly:
@reviewbybees

oleg-nenashev:
:bee:

jglick:
The exceptions are a bug in core, I think. I drew @kohsuke’s attention to them a while back but it is probably on @oleg-nenashev to follow up now. Easily reproduced just by running the functional tests in `mock-slave`.

jglick:
Can you explain more about the `chown`? What error did you see? Is it related to the Jetty version? It worked before as far as I know, so I am curious why this is coming up now.

Anyway, the `Dockerfile` does specify that `jetty.sh` be run as `jenkins`, so it seems right. :bee:

tfennelly:
@jglick You get write permission exceptions without those `chown` changes. Right, these errors do not happen with the older version of Jetty, so something must have changed there ... I didn't bother digging in to find out.

PR #228, opened 2015-10-27
JobPropertyStep with OptionalJobProperty & fixed ParametersDefinitionProperty

jglick:
Downstream of #209 and https://github.com/jenkinsci/jenkins/pull/1888.

[Incremental diff](https://github.com/jglick/workflow-plugin/compare/JobPropertyStep...jglick:JobPropertyStep%2BOptionalJobProperty)

@reviewbybees


multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
```diff
@@ -0,0 +1,177 @@
...
+            }
+            return result;
+        }
+
+    }
+
+    @Extension public static class HideSuperfluousBranchProperties extends DescriptorVisibilityFilter {
+
+        @Override public boolean filter(Object context, Descriptor descriptor) {
+            if (context instanceof WorkflowMultiBranchProject && (descriptor.clazz == RateLimitBranchProperty.class || descriptor.clazz == BuildRetentionBranchProperty.class)) {
```
amuniz: It seems hacky? If another property like `RateLimitBranchProperty` is added for some reason, it should be excluded here too. Is not there another way to exclude them?

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
```diff
@@ -0,0 +1,140 @@
...
+        assertEquals(-1, lr.getArtifactDaysToKeep());
+        assertEquals(3, lr.getArtifactNumToKeep());
+    }
+
+    @Issue("JENKINS-30206")
+    @Test public void useParameter() throws Exception {
+        sampleRepo.init();
+        ScriptApproval.get().approveSignature("method groovy.lang.Binding hasVariable java.lang.String"); // TODO add to generic whitelist
+        sampleRepo.write("Jenkinsfile",
+                "properties([[$class: 'ParametersDefinitionProperty', parameterDefinitions: [[$class: 'StringParameterDefinition', name: 'myparam', defaultValue: 'default value']]]])\n" +
```
amuniz: I know it's what we have for now, but I have to say this is the longest way to define a `key=value` pair I have ever seen in my life.

amuniz:
The compilation errors must be due to the upstream dependency, but there are some weird test failures that I'm not sure if they are related to upstream or not. Holding the bee until you confirm they are expected (and accepted).

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
jglick: @amuniz please comment on the upstream PR except for things specifically added in this PR.

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java
jglick: @amuniz ditto

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java
amuniz: Sorry. Done.

jglick:
> The compilation errors must be due to the upstream dependency, but there are some weird test failures

I have not bothered to publish snapshots of upstream dependencies since currently `multibranch` is excluded from the reactor build of this repository anyway. Running tests manually.

tfennelly:
:bee:

jglick:
Covered by #229.

PR #231, opened 2015-10-28
Wrong invocation method

recena:
if it uses `fetch(observer, listener)`, actually it is invoking `retrieve(observer, listener)`. For example, in its implementation in `AbstractGitSCMSource` this method is designed to *discover* branches.

But in [this point](https://github.com/jenkinsci/workflow-plugin/blob/master/multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java#L87) it does not want to *discover* but executes our `WorkflowJob`. For this reason, when it executes a `WorkflowJob` it sees this in the console output:

```
Started by user anonymous
Connecting to GitHub Pull Requests using recena/****** (ec28dda8d2f4d4f59b8856ef56ed83d3e5106356)
Looking up seville-jam/website

  Getting remote pull requests

    Checking pull request #PR-4
      Jenkinsfile exists in this pull request
    Met criteria

Done examining seville-jam/website

 > git rev-parse --is-inside-work-tree # timeout=10
Fetching changes from the remote Git repository
 > git config remote.origin.url https://github.com/seville-jam/website.git # timeout=10
Fetching upstream changes from https://github.com/seville-jam/website.git
 > git --version # timeout=10
using .gitcredentials to set credentials
 > git config --local credential.helper store --file=/tmp/git3668002782393825619.credentials # timeout=10
 > git fetch --tags --progress https://github.com/seville-jam/website.git +refs/pull/*/merge:refs/remotes/origin/pr/*
 > git config --local --remove-section credential # timeout=10
Checking out Revision 5c1b15276ab71c027000e1653466945a160d66f5 (PR-4)
 > git config core.sparsecheckout # timeout=10
 > git checkout -f 5c1b15276ab71c027000e1653466945a160d66f5
 > git rev-list 6919e8f59a23fa6e05f8188e77bf7442bf04eaf9 # timeout=10
[Workflow] Allocate node : Start
Running on master in /home/recena/projects/cloudbees-github-pull-requests-plugin/work/jobs/CJP-3204-4/branches/PR-4/workspace
[Workflow] node {
[Workflow] echo
Hello World 10
[Workflow] } //node
[Workflow] Allocate node : End
[Workflow] End of Workflow

GitHub has been notified of this commit’s build result

Finished: SUCCESS
```
Don't make sense scanning again our branches.

There is a specific `fetch` method to use it when the job is going to be executed:
```
SCMRevision fetch(@NonNull SCMHead head, @CheckForNull TaskListener listener)
```

@reviewbybees, specially @jglick


jglick:
This is probably the correct fix of [JENKINS-30798](https://issues.jenkins-ci.org/browse/JENKINS-30798), but I want a test—so far I have not been able to reproduce it.

recena:
@jglick I can reproduce it.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
```diff
@@ -84,10 +84,8 @@
             throw new IllegalStateException(branch.getSourceId() + " not found");
         }
         SCMHead head = branch.getHead();
-        SCMRevision tip = scmSource.fetch(SCMHeadObserver.select(head), listener).result();
+        SCMRevision tip = scmSource.fetch(head, listener);
         if (tip == null) {
```
recena: I'm not sure if we need this `if`.

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
```diff
@@ -229,4 +231,27 @@
         });
     }
 
+    @Issue("JENKINS-30798")
+    @Test public void invalidFetchMethod() throws Exception {
```
recena: This test is also valid to verify that `CRITERIA` is being executed.

tfennelly:
:bee: 

Though I'm not sure why the fix actually fixes this issue since I'm not familiar with this multi-branch code. I'm happy enough since there's a test and I'm assuming it's good (lgtm).

amuniz:
:bee:

tfennelly:
Btw ... did you ask that user on the JIRA to retest your fix?

recena:
@tfennelly I've decided to rename the PR title because these changes are related with [JENKINS-30798](https://issues.jenkins-ci.org/browse/JENKINS-30798), but strictly, don't solve the problem.

jglick:
> these changes are related with JENKINS-30798, but strictly, don't solve the problem

Can you be more specific about what it does and does not solve then? Explain yourself in the PR description.

recena:
@jglick I've tried to explain the problem.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
jglick: Well, if the `tip` is null, something is wrong, which we would like to diagnose somehow.

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
```diff
@@ -229,4 +231,56 @@
...
+    @Issue("JENKINS-30798")
+    @Test public void branchName() throws Exception {
+        story.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                sampleGitRepo.init();
+                sampleGitRepo.write("Jenkinsfile", "node { echo 'Hello World 1' }");
+                sampleGitRepo.git("add", "Jenkinsfile");
+                sampleGitRepo.git("commit", "--all", "--message=flow");
+                sampleGitRepo.git("checkout", "-b", "dev/main");
```
jglick: Branch names containing `/` were not _expected_ to work without a proper fix of [JENKINS-30744](https://issues.jenkins-ci.org/browse/JENKINS-30744), which you are not really testing here. Claiming this tests JENKINS-30798 is misleading because that exception has been observed without weird branch names. I would just delete this test for now.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
recena: @jglick It is ok to check it but `fetch()` should never return `null`

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
```diff
@@ -229,4 +231,56 @@
...
+                mp.getSourcesList().add(new BranchSource(new GitSCMSource(null, sampleGitRepo.toString(), "", "*", "", false), new DefaultBranchPropertyStrategy(new BranchProperty[0])));
+                WorkflowJob p = WorkflowMultiBranchProjectTest.scheduleAndFindBranchProject(mp, "master");
+                assertEquals(1, mp.getItems().size());
+                story.j.waitUntilNoActivity();
+                WorkflowRun b1 = p.getLastBuild();
+                assertEquals(1, b1.getNumber());
+                sampleGitRepo.git("rm", "Jenkinsfile");
+                sampleGitRepo.git("commit", "--all", "--message=remove");
+                WorkflowRun b = story.j.assertBuildStatus(Result.FAILURE, p.scheduleBuild2(0).get());
+                story.j.assertLogContains("java.io.FileNotFoundException", b);
```
jglick: We do not want to _expect_ the build to fail with a stack trace in this case.

jglick:
The patch is probably right in some sense but the tests are potentially misleading. I need to spend some time with this, evaluating the behavior before and after the main source patch, to verify what is really being fixed.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinder.java
jglick: Sure it may. It is marked `@CheckForNull`.

> `@return the revision hash (may be non-deterministic) or {@code null} if the head no longer exists.`

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
```diff
@@ -229,4 +231,56 @@
         });
     }
 
+    /**
+     * This test aims to verify that {@link WorkflowMultiBranchProject#CRITERIA} is being applied.
```
jglick: This is exactly backwards. The criteria were already being applied _before_ the main patch. With that patch, `AbstractGitSCMSource.retrieve` is called, which does _not_ check branch criteria. Probably it was not meant to; `SCMSourceOwner.getSCMSourceCriteria` talks about a “candidate” head, whereas `SCMSource.retrieve(SCMHead, TaskListener)` is allowed to return null but only if it “no longer exists” (no mention of criteria).

This test _does_ reproduce one instance of JENKINS-30798, though not the one I have observed, nor the one reported in JIRA; there could be multiple root causes.

jglick:
Can close if #235 is accepted.

PR #243, opened 2015-11-06
[JENKINS-30744] Updated to use Branch.encodedName

jglick:
Downstream of https://github.com/jenkinsci/branch-api-plugin/pull/18. Interactively tested using a [reverse proxy](https://hub.docker.com/r/jglick/jenkins-demo-reverse-proxy/) and [this repo](https://github.com/jglick/multibranch-demo) to make sure the encoded slashes were probably passed around.

Also checked that functional test passes on Windows.

@reviewbybees esp. @stephenc

### DIFF
--- a/multibranch/pom.xml
+++ b/multibranch/pom.xml
@@ -58,7 +58,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>branch-api</artifactId>
-            <version>0.2-beta-6</version>
+            <version>0.2-beta-7-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
@@ -107,6 +107,12 @@ THE SOFTWARE.
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>mercurial</artifactId>
             <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.modules</groupId>
+            <artifactId>sshd</artifactId>
+            <version>1.6</version>
+            <scope>test</scope><!-- from git-server via workflow-cps-global-lib and otherwise unavailable during tests -->
         </dependency>
     </dependencies>
     <build>
--- a/multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactoryTest.java
+++ b/multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactoryTest.java
@@ -1,0 +1,100 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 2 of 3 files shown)


multibranch/pom.xml
```diff
@@ -108,6 +108,12 @@ THE SOFTWARE.
             <artifactId>mercurial</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.modules</groupId>
+            <artifactId>sshd</artifactId>
+            <version>1.6</version>
+            <scope>test</scope><!-- from git-server via workflow-cps-global-lib and otherwise unavailable during tests -->
+        </dependency>
```
jglick: Unrelated, but stack traces had been irritating me for a while.

recena: :+1:

recena:
:bee:

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactoryTest.java
```diff
@@ -0,0 +1,100 @@
...
+                assertNotNull(p);
+                WorkflowRun b2 = p.scheduleBuild2(0).get();
+                assertEquals(2, b2.getNumber());
+                story.j.assertLogContains("branch=dev/main", b2);
+                story.j.assertLogContains("workspace=dev%2Fmain", b2);
+                verifyProject(p);
+            }
+        });
+    }
+    private void verifyProject(WorkflowJob p) throws Exception {
```
recena: I will be happy with a `newline` here

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactoryTest.java
```diff
@@ -0,0 +1,100 @@
...
+                WorkflowJob p = story.j.jenkins.getItemByFullName("p/dev%2Fmain", WorkflowJob.class);
+                assertNotNull(p);
+                WorkflowRun b2 = p.scheduleBuild2(0).get();
+                assertEquals(2, b2.getNumber());
+                story.j.assertLogContains("branch=dev/main", b2);
+                story.j.assertLogContains("workspace=dev%2Fmain", b2);
+                verifyProject(p);
+            }
+        });
+    }
```
recena: I will be happy with a newline here

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactoryTest.java
```diff
@@ -0,0 +1,100 @@
...
+    private void verifyProject(WorkflowJob p) throws Exception {
+        assertEquals("dev%2Fmain", p.getName());
+        assertEquals("dev/main", p.getDisplayName());
+        assertEquals("p/dev%2Fmain", p.getFullName());
+        assertEquals("p » dev/main", p.getFullDisplayName());
```
recena: Did you try with special characters? or `dev\main`?

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactoryTest.java
jglick: Lack of newline intentional, since the method is just a helper only used in the method above.

If Java allowed local functions I would use them. If we were in `-source 8` for `src/test/java/` I would perhaps use lambdas, though it is awkward since Java still lacks anonymous SAMs, and the standard `Consumer` does not take checked exceptions, and Guava has no apparent replacement, so you are stuck with a local SAM definition plus lambdas:

```java
@Test public void whatever() {
    interface Verifier {void verifyProject(WorkflowJob p) throws Exception;}
    final Verifier verifier = (p) -> {
        // …
    };
    // …
    verifier.verifyProject(p);
}
```

which seems more trouble than it is worth.

amuniz:
:bee:

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactoryTest.java
jglick: As mentioned in the PR description I ran this test on Windows (XP, if it matters) and it passed, so there is not a filesystem problem. `/` is the only special character I have heard of from the field (used in Git Flow IIUC). `\` is handled by `Util.rawEncode`. If there are other characters which are _not_ handled by `rawEncode` but which pose a problem on some filesystems, that could be a follow-up fix localized to `Branch.getEncodedName` plus `MultiBranchProject.rawDecode`.

recena: Thanks for your comments

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactoryTest.java
recena: IMO, are different things but you explanation was nice.

PR #244, opened 2015-11-06
[JENKINS-33256] Calling SCMSource.getTrustedRevision

jglick:
Downstream of https://github.com/jenkinsci/scm-api-plugin/pull/5.

@reviewbybees esp. @stephenc


stephenc:
:bee:

recena:
:bee:

KostyaSha:
Status?

jglick:
I ran out of time to try using the new API and confirm that it solves the problem. On my list.

KostyaSha:
@jglick i see no examples or description of supporting multibranches

jglick:
Because the proposed usage is in a plugin which is currently not OSS.

kzantow:
:bee:

jglick:
@reviewbybees done

(but pointless to merge unless and until https://github.com/jenkinsci/github-branch-source-plugin/pull/29 is)

jglick:
Actually I take that back—Pipeline has a more heavyweight release process so there might be a timing advantage in having this released earlier. Also I know of at least one other `SCMSource` which might be able to take advantage of this (@amuniz). The patch should be safe enough since `getTrustedRevision` defaults to just returning its argument, so the worst that can happen is that the API turns out to be not as useful as originally thought and must be deprecated.

An `scm-api` 1.1 release is a prerequisite, though.

PR #249, opened 2015-11-16
[JENKINS-29705] Added thread dump support

kohsuke:
[JENKINS-29705](https://issues.jenkins-ci.org/browse/JENKINS-29705)

This change adds a virtual thread dump of workflow CPS threads.

- [X] basic code contribution from @kohsuke
- [X] add a sidepanel
- [X] fix API for allowing `workflow-cps` to attach a UI to something like `WorkflowRun` with a `CpsFlowExecution`
- [X] improve display of steps in stack trace
- [ ] fix up file names to reflect actual files the user created
- [x] fix random test failures
- [x] investigate whether #252 fix can be simplified with new `FlowExecutionOwner.Executable`
- [ ] include native portion of thread when blocked in it
- [ ] add a `support-core` `Component` showing thread dumps for all running flow builds
- [x] hide action when flow completes
- [ ] print stack trace to build log before offering `term` link

@reviewbybees


kohsuke:
And what's my next airplane project, @jglick ?

job/src/main/java/org/jenkinsci/plugins/workflow/job/ActionFromFlowExecution.java
```diff
@@ -0,0 +1,67 @@
...
+
+        @Override
+        public String getDisplayName() {
+            return base.getDisplayName();
+        }
+
+        @Override
+        public String getUrlName() {
+            String u = base.getUrlName();
+            if (u!=null) {
```
recena: `if (u != null) {`

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -35,6 +35,7 @@
 import hudson.XmlFile;
 import hudson.console.AnnotatedLargeText;
 import hudson.console.LineTransformationOutputStream;
+import hudson.model.Action;
```
recena: required?

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -179,6 +180,7 @@ public WorkflowRun(WorkflowJob job, File dir) throws IOException {
         return getRunMixIn().getNextBuild();
     }
 
+
```
recena: Unrelated change.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsThreadDump/index.jelly
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?jelly escape-by-default='true'?>
+<!--
+  ~ The MIT License
+  ~
+  ~ Copyright (c) 2013-2014, CloudBees, Inc.
```
recena: Wrong copyright

jglick:
Some FindBugs failures so I suppose I will need to take this over.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDumpTest.java
```diff
@@ -0,0 +1,108 @@
...
+            "WorkflowScript.foo(WorkflowScript:1)",
+            "WorkflowScript.zot(WorkflowScript:7)",
+            "org.jenkinsci.plugins.workflow.cps.steps.ParallelStepExecution.<>(Native Method)",
+            "WorkflowScript.zot(WorkflowScript:6)",
+            "WorkflowScript.run(WorkflowScript:10)");
+
+        assertStackTrace( td.getThreads().get(1),
+            "org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep$Execution.<>(Native Method)",
+            "WorkflowScript.bar(WorkflowScript:3)",
+            "WorkflowScript.zot(WorkflowScript:8)");
```
jglick: A little odd that this does not show the rest of the stack that thread 0 shows, but OK—I guess that is considered redundant information?

api/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
```diff
@@ -211,4 +214,10 @@ public boolean blocksRestart() {
      */
     public abstract @Nonnull Authentication getAuthentication();
 
+    /**
+     * Create {@link Action}s to be attached to {@linkplain #getOwner() owner} when it has UI.
+     */
+    public @Nonnull Collection<? extends Action> createActions() {
```
jglick: Or `TransientActionFactory`?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -637,6 +641,31 @@ public void onFailure(Throwable t) {
         return r;
     }
 
+    /**
+     * Synchronously obtain the current state of the workflow program.
+     *
+     * <p>
+     * The workflow can be already completed, or it can be in the process of
```
jglick: …?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -637,6 +641,31 @@ public void onFailure(Throwable t) {
...
+     */
+    public CpsThreadDump getThreadDump() {
+        if (programPromise == null || isComplete()) {
+            return CpsThreadDump.EMPTY;
+        }
+        if (!programPromise.isDone()) {
+            // CpsThreadGroup state isn't ready yet, but this is probably one of the common cases
+            // when one wants to obtain the stack trace. Can we do anything better?
```
jglick: I would hope so: [JENKINS-26130](https://issues.jenkins-ci.org/browse/JENKINS-26130)

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDump.java
```diff
@@ -0,0 +1,139 @@
...
+            CpsThread head = e.get(e.size() - 1);
+            headline = head.toString();
+
+            ListIterator<CpsThread> itr = e.listIterator(e.size());
+            while (itr.hasPrevious()) {
+                CpsThread t = itr.previous();
+
+                StepExecution s = t.getStep();
+                if (s !=null) {
+                    stack.add(new StackTraceElement(s.getClass().getName(), "<>", null, -2));
```
jglick: I think the `functionName` would be more appropriate here. People do not generally care about the implementation class.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDump.java
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Constant that indicates everything is done and no thread is alive.
+     */
+    public static CpsThreadDump EMPTY = new CpsThreadDump();
+
+    /**
+     * Constant that indicates the state of {@link CpsThreadGroup} is unknown and so it is not possible
+     * to produe a thread dump.
+     */
+    public static CpsThreadDump UNKNOWN = from(new Exception("Program state is not yet known"));
```
jglick: We do not actually want to capture the stack trace here.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/CpsThreadDump/index.jelly
```diff
@@ -0,0 +1,42 @@
...
+  ~ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  ~ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  ~ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  ~ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+  ~ THE SOFTWARE.
+  -->
+
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:i="jelly:fmt">
+  <l:layout title="${%Thread Dump}">
+    <l:main-panel>
```
jglick: Probably want to add a sidepanel here.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowExecution.java
jglick: I think you could get rid of this API and `ActionFromFlowExecution` if there were simply a supported way to go from `Run` to `FlowExecution` without linking to `WorkflowRun.getExecutionPromise`, which would be useful elsewhere anyway.

jglick:
Needs some minor work but the meat is there. Shall I take it over?

jglick:
Should also fix script names, which are currently bogus:

* `CpsFlowExecution.parseScript` should not hardcode `WorkflowScript` but should have a constructor parameter for the name. `CpsFlowDefinition` can use whatever; `CpsScmFlowDefinition` should use `scriptFile.getName()` (for example `Jenkinsfile`).
* `LoadStepExecution.start` should use:

```java
FilePath file = cwd.child(step.getPath());
Script script = shell.parse(file.readToString(), file.getName());
```

jglick:
…though `CpsFlowExecution.loadedScripts` seems to assume that script names are unique, so perhaps `CpsGroovyShell.generateScriptName` should be deleted and `CpsGroovyShell.parse` should uniquify names if necessary.

jglick:
Would like to fix script names, but the implications look daunting. I cannot find a place where Groovy allows you to set the filename associated with a class without actually changing the class name, which could have all sorts of evil effects—on compatibility of historical program records, on correctness when multiple scripts with the same filename are loaded, etc. I would not mind if the class name stayed as is, so long as the filename portion reflected the user view:

    at Script3.method(my-workflow-lib.groovy:99)

jglick:
BTW @kohsuke `CpsThreadDumpTest` frequently fails for me locally: stack trace elements are simply missing without explanation. Is there some sort of race condition in the code that would explain this?

jglick:
Specifically `Continuable.getStackTrace()` is sometimes empty when it should not be.

jglick:
Found it: a race condition. I can make the tests pass reliably by forcing `CpsFlowExecution.getThreadDump` to run in the VM thread. But that reduces the usefulness of the feature, since your program might well be stuck in some native method in the VM thread. (Though in that case the current code does offer the native portion of the stack trace, which it really ought to, by capturing the `Thread` stack from `CpsVmExecutorService`.)

jglick:
Listed some open enhancements for this feature, but none critical.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDumpAction.java
```diff
@@ -0,0 +1,71 @@
...
+    @Override
+    public String getUrlName() {
+        return "threadDump";
+    }
+    
+    public String getParentUrl() throws IOException {
+        return execution.getOwner().getUrl();
+    }
+
+    /* for tests */ CpsThreadDump threadDumpSynchonous() throws InterruptedException, ExecutionException {
```
amuniz: Typo

amuniz:
:bee: (as far as I get it - which is not too far).

What about printing the stacktrace unconditionally on build kill request?

jglick:
> What about printing the stacktrace unconditionally on build kill request?

Yes, this is a good idea which I will add to the list of things I am not going to implement now but could be done as a follow-up.

tfennelly:
:bee: insofar as I get it too

Main thing for me is ... seems like a really useful feature + smell nice in general + it has test cases => onward with it

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDumpAction.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public String getIconFileName() {
+        return "gear.png";
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "Thread Dump";
```
Vlatombe: Do we want to allow anyone to run this? Maybe restrict to Item.BUILD or admin?

Vlatombe:
:bee: That would be super useful for hang investigations and support cases.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadDumpAction.java
jglick: The current code lets anyone who has read permission on the build see the thread dump. I do not see any particular reason to restrict it; the steps being run are already visible in the build log, so the only additional information you would get is the function names and line numbers in scripts, which does not seem like it would be secret information. Do you feel otherwise?

Vlatombe: Looks good to me.

jglick:
I filed the outstanding tasks as blocking issues of JENKINS-29705.

PR #250, opened 2015-11-17
Update jetty version for demo

amuniz:
`9.2.12.v20150709` is not available anymore on ibiblio (404 on `http://mirrors.ibiblio.org/eclipse/jetty/9.2.12.v20150709`)

@reviewbybees


recena:
:bee: Did you try with other mirror? Any announce from Jetty project about this?

PR #252, opened 2015-11-17
[JENKINS-25889] InputStepExecution.node null after restart

jglick:
[JENKINS-25889](https://issues.jenkins-ci.org/browse/JENKINS-25889)

- [X] reproduced error in test
- [X] downgraded error to warning
- [x] full fix

@reviewbybees esp. @kohsuke


support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/InputAction.java
```diff
@@ -27,9 +33,51 @@ public void onAttached(Run<?, ?> r) {
     @Override
     public void onLoad(Run<?, ?> r) {
         this.run = r;
-        assert executions != null && !executions.contains(null) : executions;
-        for (InputStepExecution step : executions) {
-            step.run = run;
+        synchronized (this) {
+            if (ids == null) {
+                // Loading from before JENKINS-25889 fix. Load the IDs and discard the executions, which lack state anyway.
+                assert executions != null && !executions.contains(null) : executions;
```
amuniz: Will not be `executions == null` always here?

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/InputAction.java
```diff
@@ -27,9 +33,51 @@ public void onAttached(Run<?, ?> r) {
...
+                        }
+                    }
+                }
+                if (executions.size() < ids.size()) {
+                    LOGGER.log(Level.WARNING, "some input IDs not restored from {0}", run);
+                }
+            } else {
+                LOGGER.log(Level.WARNING, "no flow execution found for {0}", run);
+            }
+            } catch (Exception x) {
```
amuniz: Wrong indentation of the `try` body.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepRestartTest.java
```diff
@@ -24,43 +24,78 @@
...
+    @LocalData // from 1.4.2
+    @Test public void oldFlow() {
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = jenkins().getItemByFullName("p", WorkflowJob.class);
+                assertNotNull(p);
+                WorkflowRun b = p.getLastBuild();
+                assertNotNull(b);
+                assertEquals(1, b.getNumber());
+                proceed(b);
```
amuniz: Trying to proceed on a build that was not started? Or perhaps the LocalData has an unfinished build?

amuniz:
:bee:

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/input/InputStepRestartTest.java
jglick: The local data has a build in progress, paused in `input`. That is the point of this test case: that the old `InputAction` format could be safely read and translated.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/InputAction.java
jglick: Not when we are deserializing an `InputAction` created before this patch, when the field was non-`transient`.

PR #254, opened 2015-11-19
[JENKINS-25550] Make it clearer it's a clickable link

batmat:
Might seem obvious but I've already had many people *not* clicking on
the link thinking this was only something informational.


jglick:
Good point. I also noted in JENKINS-25550 that it would be useful to print something while the 15s timer is counting down, but that is a slightly more complex patch.

batmat:
Yup. I thought about it, but that it would be more involved and that the
current one would already be a (little) improvement.
Maybe I'll see if that could be something I can tackle, though unlikely
next week at least.
Thanks

PR #264, opened 2015-12-04
[JENKINS-31386] Standalone support for checkout scm

jglick:
[JENKINS-31386](https://issues.jenkins-ci.org/browse/JENKINS-31386)

@reviewbybees esp @abayer

### DIFF
--- a/multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMVar.java
+++ b/multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMVar.java
@@ -36,8 +36,11 @@ import java.io.Serializable;
 import jenkins.branch.Branch;
 import jenkins.scm.api.SCMRevisionAction;
 import jenkins.scm.api.SCMSource;
+import org.jenkinsci.plugins.workflow.cps.CpsScmFlowDefinition;
 import org.jenkinsci.plugins.workflow.cps.CpsScript;
 import org.jenkinsci.plugins.workflow.cps.GlobalVariable;
+import org.jenkinsci.plugins.workflow.flow.FlowDefinition;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.pickles.Pickle;
 import org.jenkinsci.plugins.workflow.support.pickles.SingleTypedPickleFactory;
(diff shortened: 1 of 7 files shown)


multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
```diff
@@ -144,171 +104,114 @@ private static void assertRevisionAction(WorkflowRun build) {
     }
 
     @Test public void exactRevisionSubversion() throws Exception {
-        story.addStep(new Statement() {
```
abayer: So there's a pile o' diffs like this - are they largely just indent changes or should I be reading more closely?

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMVar.java
```diff
@@ -62,6 +65,13 @@
         Job<?,?> job = build.getParent();
         BranchJobProperty property = job.getProperty(BranchJobProperty.class);
         if (property == null) {
+            if (job instanceof WorkflowJob) {
+                FlowDefinition defn = ((WorkflowJob) job).getDefinition();
+                if (defn instanceof CpsScmFlowDefinition) {
+                    // JENKINS-31386: retrofit to work with standalone projects, minus the exact revision support.
```
abayer: "exact revision support" meaning the ability to specify the specific revision?

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/SCMBinderTest.java
jglick: There are two commits in the PR. It will probably make more sense to read them individually.

abayer: Gotcha.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/SCMVar.java
jglick: See the series of tests in `SCMBinderTest`. From a multibranch project, `scm` is customized for this build to refer to the exact revision from which `Jenkinsfile` itself was loaded, allowing you to make atomic commits with confidence.

From `CpsScmFlowDefinition` that is not possible (`SCM` provides no suitable API—only multibranch-aware `SCMSource`), so `scm` will be the same configuration as is statically in the project definition, meaning the checkout could be of a newer revision in the same branch (or whatever).

jglick: (also alluded to in JIRA)

abayer: Gotcha.

abayer:
:bee:

PR #266, opened 2015-12-05
[JENKINS-31086] Add useDefaultExcludes option to stash

abayer:
[JENKINS-31086](https://issues.jenkins-ci.org/browse/JENKINS-31086)

Defaults to true, meaning use Ant default excludes for
DirScanner.Glob. If false, don't use the default excludes, to pick up
files like .gitignore etc.

@reviewbybees

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
@@ -29,13 +29,14 @@ import org.jenkinsci.plugins.workflow.flow.StashManager;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
-import org.junit.Test;
-import static org.junit.Assert.*;
 import org.junit.ClassRule;
 import org.junit.Rule;
+import org.junit.Test;
 import org.jvnet.hudson.test.BuildWatcher;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
+
+import static org.junit.Assert.*;
 
 public class StashTest {
 
@@ -69,4 +70,26 @@ public class StashTest {
         assertEquals("{}", StashManager.stashesOf(b).toString());
     }
 
+    @Issue("JENKINS-31086")
+    @Test public void testDefaultExcludes() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition(
+                "node {\n" +
+                        "  writeFile file: 'subdir/.gitignore', text: 'whatever'\n" +
+                        "  writeFile file: 'subdir/otherfile', text: 'whatever'\n" +
+                        "  dir('subdir') {stash name:'has-gitignore', useDefaultExcludes: false}\n" +
+                        "  dir('subdir') {stash name:'no-gitignore' }\n" +
+                        "  dir('first-unstash') {\n" +
+                        "    unstash('has-gitignore')\n" +
+                        "    echo \"gitignore exists? ${fileExists '.gitignore'}\"\n" +
+                        "  }\n" +
+                        "  dir('second-unstash') {\n" +
+                        "    unstash('no-gitignore')\n" +
+                        "    echo \"gitignore does not exist? ${fileExists '.gitignore'}\"\n" +
+                        "  }\n" +
+                        "}", true));
+        WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
+        r.assertLogContains("gitignore exists? true", b);
+        r.assertLogContains("gitignore does not exist? false", b);
+    }
 }
--- a/api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
+++ b/api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
@@ -35,17 +35,6 @@ import hudson.model.TaskListener;
 import hudson.org.apache.tools.tar.TarInputStream;
 import hudson.util.DirScanner;
 import hudson.util.io.ArchiverFactory;
-import java.io.ByteArrayOutputStream;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.util.Map;
-import java.util.TreeMap;
-import javax.annotation.CheckForNull;
-import javax.annotation.Nonnull;
 import jenkins.model.ArtifactManager;
 import jenkins.model.Jenkins;
 import org.apache.commons.io.FileUtils;
@@ -53,6 +42,12 @@ import org.apache.commons.io.IOUtils;
 import org.apache.tools.tar.TarEntry;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import java.io.*;
+import java.util.Map;
+import java.util.TreeMap;
 
 /**
  * Manages per-build stashes of files.
@@ -66,6 +61,12 @@ import org.kohsuke.accmod.restrictions.DoNotUse;
  */
 public class StashManager {
 
+    @Deprecated
+    public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull TaskListener listener,
+                             @CheckForNull String includes, @CheckForNull String excludes) throws IOException, InterruptedException {
+        stash(build, name, workspace, listener, includes, excludes, true);
+    }
+
     /**
      * Saves a stash of some files from a build.
      * @param build a build to use as storage
@@ -73,9 +74,11 @@ public class StashManager {
      * @param workspace a directory to use as a base
      * @param includes a set of Ant-style file includes, separated by commas; null/blank is allowed as a synonym for {@code **} (i.e., everything)
      * @param excludes an optional set of Ant-style file excludes
+     * @param useDefaultExcludes whether to use Ant default excludes
      */
     @SuppressFBWarnings(value="RV_RETURN_VALUE_IGNORED_BAD_PRACTICE", justification="fine if mkdirs returns false")
-    public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull TaskListener listener, @CheckForNull String includes, @CheckForNull String excludes) throws IOException, InterruptedException {
+    public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull TaskListener listener,
+                             @CheckForNull String includes, @CheckForNull String excludes, boolean useDefaultExcludes) throws IOException, InterruptedException {
         Jenkins.checkGoodName(name);
         File storage = storage(build, name);
         storage.getParentFile().mkdirs();
@@ -84,7 +87,7 @@ public class StashManager {
         }
         OutputStream os = new FileOutputStream(storage);
         try {
-            int count = workspace.archive(ArchiverFactory.TARGZ, os, new DirScanner.Glob(Util.fixEmpty(includes) == null ? "**" : includes, excludes));
+            int count = workspace.archive(ArchiverFactory.TARGZ, os, new DirScanner.Glob(Util.fixEmpty(includes) == null ? "**" : includes, excludes, useDefaultExcludes));
             if (count == 0) {
                 throw new AbortException("No files included in stash");
             }
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep.java
@@ -30,8 +30,6 @@ import hudson.FilePath;
 import hudson.Util;
 import hudson.model.Run;
 import hudson.model.TaskListener;
-import javax.annotation.CheckForNull;
-import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.flow.StashManager;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepDescriptorImpl;
@@ -41,11 +39,15 @@ import org.jenkinsci.plugins.workflow.steps.StepContextParameter;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
 
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
 public class StashStep extends AbstractStepImpl {
 
     private final @Nonnull String name;
     private @CheckForNull String includes;
     private @CheckForNull String excludes;
+    private boolean useDefaultExcludes = true;
 
     @DataBoundConstructor public StashStep(@Nonnull String name) {
         Jenkins.checkGoodName(name);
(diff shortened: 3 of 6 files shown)


abayer:
Closing/reopening - those failures look spurious to me.

abayer:
Can't reproduce this locally - had one test failure on {{org.jenkinsci.plugins.workflow.steps.ExecutorStepTest.buildShellScriptAcrossDisconnect(org.jenkinsci.plugins.workflow.steps.ExecutorStepTest)}}, but that looks transient as well. Could be a concurrency issue?

abayer:
Rebased.

abayer:
Woo, and now the tests actually decided to pass. =)

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
```diff
@@ -69,4 +70,30 @@
...
+                        "  dir('subdir') {stash name:'whatever', useDefaultExcludes: false}\n" +
+                        "}\n" +
+                        "node {\n" +
+                        "  dir('elsewhere') {\n" +
+                        "    unstash 'whatever'\n" +
+                        "    echo \"got .gitignore: ${readFile '.gitignore'} other: ${readFile 'other'}\"\n" +
+                        "  }\n" +
+                        "  writeFile file: 'at-top', text: 'ignored'\n" +
+                        "  stash name: 'from-top', includes: 'elsewhere/'\n" +
+                        "  semaphore 'ending'\n" +
```
jglick: :ant: These three lines serve no apparent function in this test case and could have been skipped along with the asserts about `StashManager`.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -25,27 +25,12 @@
 package org.jenkinsci.plugins.workflow.flow;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-import hudson.AbortException;
-import hudson.ExtensionList;
-import hudson.ExtensionPoint;
-import hudson.FilePath;
-import hudson.Util;
+import hudson.*;
```
jglick: Generally prefer to avoid wildcard imports and to lexicographically sort all imports.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -73,9 +64,11 @@
      * @param workspace a directory to use as a base
      * @param includes a set of Ant-style file includes, separated by commas; null/blank is allowed as a synonym for {@code **} (i.e., everything)
      * @param excludes an optional set of Ant-style file excludes
+     * @param useDefaultExcludes whether to use Ant default excludes
      */
     @SuppressFBWarnings(value="RV_RETURN_VALUE_IGNORED_BAD_PRACTICE", justification="fine if mkdirs returns false")
-    public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull TaskListener listener, @CheckForNull String includes, @CheckForNull String excludes) throws IOException, InterruptedException {
+    public static void stash(@Nonnull Run<?,?> build, @Nonnull String name, @Nonnull FilePath workspace, @Nonnull TaskListener listener,
+                             @CheckForNull String includes, @CheckForNull String excludes, boolean useDefaultExcludes) throws IOException, InterruptedException {
```
jglick: :bug: incompatible API change—would lead to errors if someone updated `workflow-api` but not `workflow-support`.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep.java
```diff
@@ -41,11 +39,15 @@
...
 
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
 public class StashStep extends AbstractStepImpl {
 
     private final @Nonnull String name;
     private @CheckForNull String includes;
     private @CheckForNull String excludes;
+    private boolean useDefaultExcludes = true;
```
jglick: :bug: need to also add to `config.jelly`

```xml
<f:entry field="useDefaultExcludes" title="…something…">
    <f:checkbox default="true"/>
</f:entry>
```

and a matching `help-useDefaultExcludes.html`.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep.java
```diff
@@ -72,6 +74,14 @@ public String getExcludes() {
         this.excludes = Util.fixEmpty(excludes);
     }
 
+    public boolean getUseDefaultExcludes() {
```
jglick: `isUseDefaultExcludes` is a little more conventional but I guess it does not matter.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
abayer: Which three lines? Not sure what you're referring to.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
abayer: Yeah, fixing my Intellij settings. Grr.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
abayer: So what should I do? How do we handle adding new arguments like this?

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep.java
abayer: Brain barfed. Will fix.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep.java
abayer: Eh, I'm good with rewording it. I never remember the conventions. =)

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
jglick:
```groovy
writeFile file: 'at-top', text: 'ignored'
stash name: 'from-top', includes: 'elsewhere/'
semaphore 'ending'
```

all copied from `smokes` but not testing anything novel here. The only thing this test case needs to check is that `subdir/.gitignore` was stashed.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
jglick: Retain the original signature `@Deprecated`, delegate to the new method (passing `useDefaultExcludes=true`).

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
abayer: I'd like to keep the `stash` - I like having the test first stash without default excludes and then do a second stash with default excludes enabled to make sure it's covering both options.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
abayer: Gotcha - wasn't sure if we had a different way of handling things here than I'm used to. It's been a while since I've done much Jenkins hacking. =)

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
jglick: But the second `stash` is not even including `subdir`, so it is not testing the implicit `useDefaultExcludes: true`.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
jglick: Yeah…in cases like these, most people will update all the WF plugins simultaneously, but Plugin Manager does not enforce this, so I like to err on the side of caution. If and when we have a system for core & plugins to declare that particular downstream plugins are known to be broken by an update and must be updated in tandem, we can start to delete the deprecated detritus.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
abayer: Yeah, it is - the `unstash` is within `dir('elsewhere')`, and the original `stash` is done from within `subdir`, so we're not actually ever stashing the directory, just its contents.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
abayer: Gotcher.

abayer:
Addressed comments, re-pushed.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
jglick: You could, for example, do something like

```groovy
dir('subdir') {stash 'stripped'}
// …
unstash 'stripped'
echo "have .gitignore? ${fileExists '.gitignore'}"
```

and assert that it prints

    have .gitignore? false

abayer: Ah, ok. That is simpler. Reworking it.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
```diff
@@ -69,4 +70,26 @@
...
+                        "}\n" +
+                        "node {\n" +
+                        "  dir('elsewhere') {\n" +
+                        "    unstash 'whatever'\n" +
+                        "    echo \"got .gitignore: ${readFile '.gitignore'} other: ${readFile 'other'}\"\n" +
+                        "  }\n" +
+                        "  stash name: 'from-top', includes: 'elsewhere/'\n" +
+                        "}", true));
+        WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+        assertEquals("{from-top={elsewhere/other=more}, whatever={.gitignore=whatever, other=more}}", StashManager.stashesOf(b).toString());
```
jglick: :bug: this is not going to work reliably. The build might have completed (deleting the stash) by the time you hit this line.

My suggestion was to delete _both_ the calls to `StashManager` and `SemaphoreStep`, not just one.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
```diff
@@ -74,8 +69,24 @@
      * @param includes a set of Ant-style file includes, separated by commas; null/blank is allowed as a synonym for {@code **} (i.e., everything)
      * @param excludes an optional set of Ant-style file excludes
      */
```
jglick: BTW I would just delete the Javadoc of a deprecated method—noise.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep.java
jglick: (resolved)

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
abayer: Yeah, now I get where you were going and have a new version about to get pushed.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
abayer: Okiedokie!

support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/stash/StashStep/help-useDefaultExcludes.html
```diff
@@ -0,0 +1,3 @@
+<div>
+    If selected, use the default excludes from Ant - see <a href="http://stackoverflow.com/a/6351685">here</a> for the list.
```
jglick: Perhaps better to link directly to the [manual](http://ant.apache.org/manual/dirtasks.html#defaultexcludes).

BTW I like to use `target="_blank"` in all links in help text, to work around [JENKINS-18435](https://issues.jenkins-ci.org/browse/JENKINS-18435).

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
abayer: FYI, to get this to work, I did need to create a second file so that the `stash` with implicit `useDefaultExcludes` doesn't error out due to no files to stash.

abayer:
Yet more addressing and pushing!

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
```diff
@@ -29,14 +29,15 @@
...
-import org.junit.Test;
-import static org.junit.Assert.*;
 import org.junit.ClassRule;
 import org.junit.Rule;
+import org.junit.Test;
 import org.jvnet.hudson.test.BuildWatcher;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
 
+import static org.junit.Assert.assertEquals;
```
jglick: FWIW [usual code style](https://github.com/jenkinsci/branch-api-plugin/blob/master/CONTRIBUTING.md#imports) is to use static imports for these.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
```diff
@@ -69,4 +70,27 @@
...
+                        "    unstash('has-gitignore')\n" +
+                        "    echo \"gitignore exists? ${fileExists '.gitignore'}\"\n" +
+                        "  }\n" +
+                        "  dir('second-unstash') {\n" +
+                        "    unstash('no-gitignore')\n" +
+                        "    echo \"gitignore does not exist? ${fileExists '.gitignore'}\"\n" +
+                        "  }\n" +
+                        "}", true));
+        WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+        r.assertBuildStatusSuccess(r.waitForCompletion(b));
```
jglick: Or more concisely,

```java
WorkflowRun b = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
```

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
abayer: Wildcard, you mean? Ok.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/StashManager.java
jglick: (resolved)

jglick:
:bee:

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
jglick: Sorry, yes, static wildcard imports.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/stash/StashTest.java
abayer: Done.

abayer:
Hopefully one last round of addressing comments. =)

abayer:
fyi, https://jenkins.ci.cloudbees.com/job/plugins/job/workflow-plugin/1955/ is the relevant build for the current state.

jglick:
:bee:

abayer:
Fixed comment.

jglick:
:bee:

abayer:
Rebased, merging.

PR #269, opened 2015-12-10
[JENKINS-31897] Set default parameters on build calls

amuniz:
[JENKINS-31897](https://issues.jenkins-ci.org/browse/JENKINS-31897)

@reviewbybees

### DIFF
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
@@ -1,5 +1,7 @@
 package org.jenkinsci.plugins.workflow.support.steps.build;
 
+import com.google.common.base.Function;
+import com.google.common.collect.Lists;
 import com.google.inject.Inject;
 import hudson.AbortException;
 import hudson.console.ModelHyperlinkNote;
@@ -9,8 +11,10 @@ import hudson.model.CauseAction;
 import hudson.model.Computer;
 import hudson.model.Executor;
 import hudson.model.Job;
+import hudson.model.ParameterDefinition;
 import hudson.model.ParameterValue;
 import hudson.model.ParametersAction;
+import hudson.model.ParametersDefinitionProperty;
 import hudson.model.Queue;
 import hudson.model.Result;
 import hudson.model.Run;
@@ -54,6 +58,7 @@ public class BuildTriggerStepExecution extends AbstractStepExecutionImpl {
         actions.add(new CauseAction(new Cause.UpstreamCause(invokingRun)));
         List<ParameterValue> parameters = step.getParameters();
         if (parameters != null) {
+            completeDefaultParameters(parameters);
             actions.add(new ParametersAction(parameters));
         }
         Integer quietPeriod = step.getQuietPeriod();
@@ -75,6 +80,26 @@ public class BuildTriggerStepExecution extends AbstractStepExecutionImpl {
         } else {
             getContext().onSuccess(null);
             return true;
+        }
+    }
+
+    private void completeDefaultParameters(List<ParameterValue> parameters) {
+        List<String> names = Lists.transform(parameters, new Function<ParameterValue, String>() {
+            @Override public String apply(ParameterValue input) {
+                return input.getName();
+            }
+        });
+        final Job<?,?> project = Jenkins.getActiveInstance().getItem(step.getJob(), invokingRun.getParent(), Job.class);
+        if (project != null) {
+            ParametersDefinitionProperty pdp = project.getProperty(ParametersDefinitionProperty.class);
+            if (pdp != null && pdp.getParameterDefinitionNames().size() > parameters.size()) {
+                for (String name : pdp.getParameterDefinitionNames()) {
+                    ParameterDefinition pDef = pdp.getParameterDefinition(name);
+                    if (!names.contains(pDef.getName()) && pDef.getDefaultParameterValue() != null) {
+                        parameters.add(pDef.getDefaultParameterValue());
+                    }
+                }
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -78,6 +83,26 @@ public boolean start() throws Exception {
         }
     }
 
+    private void completeDefaultParameters(List<ParameterValue> parameters) {
+        List<String> names = Lists.transform(parameters, new Function<ParameterValue, String>() {
+            @Override public String apply(ParameterValue input) {
+                return input.getName();
+            }
+        });
+        final Job<?,?> project = Jenkins.getActiveInstance().getItem(step.getJob(), invokingRun.getParent(), Job.class);
```
jglick: Are we not already doing this calculation? Share the result.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -54,6 +58,7 @@ public boolean start() throws Exception {
         actions.add(new CauseAction(new Cause.UpstreamCause(invokingRun)));
         List<ParameterValue> parameters = step.getParameters();
         if (parameters != null) {
+            completeDefaultParameters(parameters, (Job) project);
```
rsandell: And what happens if there are no parameters in the call? Will core add all the default parameters for you?
I'm asking because I can't remember and at least it would be nice to have a test for it.

amuniz: Yes, the issue only comes up when _some but not all_ parameters are set explicitly.
There is already a [test](https://github.com/amuniz/workflow-plugin/blob/JENKINS-31897/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/build/BuildTriggerStepTest.java#L205) for the _all defaults_ case.

varmenise:
:bee: LGTM, except the 92 failing tests which look being there since a while..?

jglick:
The test failures are probably [JENKINS-30395](https://issues.jenkins-ci.org/browse/JENKINS-30395). This commit has the potential to introduce regressions, so need to verify that at least all the tests in this package still pass.

jglick:
https://github.com/jenkinsci/jenkins/pull/553 is the proposed core change that would obviate such per-plugin logic.

amuniz:
From Sep 3, 2012. Nice.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
jglick: :bug: return a new `List<ParameterValue>` rather than modifying the state of the `Step`.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -78,6 +83,25 @@ public boolean start() throws Exception {
...
 
+    private void completeDefaultParameters(List<ParameterValue> parameters, Job<?,?> project) {
+        List<String> names = Lists.transform(parameters, new Function<ParameterValue, String>() {
+            @Override public String apply(ParameterValue input) {
+                return input.getName();
+            }
+        });
+        if (project != null) {
+            ParametersDefinitionProperty pdp = project.getProperty(ParametersDefinitionProperty.class);
+            if (pdp != null && pdp.getParameterDefinitionNames().size() > parameters.size()) {
```
jglick: :bug: since you are not verifying that the specified parameters are a strict subset, better to limit the test to `pdp != null`.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -78,6 +83,25 @@ public boolean start() throws Exception {
...
+                for (String name : pdp.getParameterDefinitionNames()) {
+                    ParameterDefinition pDef = pdp.getParameterDefinition(name);
+                    if (!names.contains(pDef.getName()) && pDef.getDefaultParameterValue() != null) {
```
jglick: :ant: call `getDefaultParameterValue()` only once (and only for an undefined parameter).

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -78,6 +83,25 @@ public boolean start() throws Exception {
...
+        List<String> names = Lists.transform(parameters, new Function<ParameterValue, String>() {
+            @Override public String apply(ParameterValue input) {
+                return input.getName();
+            }
+        });
+        if (project != null) {
+            ParametersDefinitionProperty pdp = project.getProperty(ParametersDefinitionProperty.class);
+            if (pdp != null && pdp.getParameterDefinitionNames().size() > parameters.size()) {
+                for (String name : pdp.getParameterDefinitionNames()) {
+                    ParameterDefinition pDef = pdp.getParameterDefinition(name);
```
jglick: :ant: bug easier to loop over `getParameterDefinitions()` directly.

amuniz:
It seems that [JENKINS-30395](https://issues.jenkins-ci.org/browse/JENKINS-30395) is hitting again.

amuniz:
Can not reproduce it locally. Since the exception seems to occur during I/O operations (copying files usually), I tried to run a single test (one which is failing in CI) using a fake big dependency (200 MB) to force a big copy that take some time to perform. No luck. Test passed :crying_cat_face:

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/BuildTriggerStepExecution.java
```diff
@@ -78,6 +83,29 @@ public boolean start() throws Exception {
...
     }
 
+    private List<ParameterValue> completeDefaultParameters(List<ParameterValue> parameters, Job<?,?> project) {
+        List<ParameterValue> completeListOfParameters = Lists.newArrayList(parameters);
+        List<String> names = Lists.transform(parameters, new Function<ParameterValue, String>() {
+            @Override public String apply(ParameterValue input) {
+                return input.getName();
+            }
+        });
+        if (project != null) {
```
jglick: :ant: we already did a null check in line 48, so this is unnecessary (and the parameter could be `@Nonnull`).

jglick:
:bee:

PR #270, opened 2015-12-11
Correcting help for the shell step.

jglick:
Amends help in #211 so CC @amuniz. `-xe` are passed only if an explicit interpreter is _not_ defined. [reference](https://github.com/jenkinsci/durable-task-plugin/blob/b27495f9190939590660c29971317ddcd2c774cf/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java#L70-L73)

@reviewbybees


recena:
:bee:

amuniz:
:bee:

abayer:
:bee:

jglick:
Test failures look like [JENKINS-30395](https://issues.jenkins-ci.org/browse/JENKINS-30395), which unfortunately is not fixed in the 1.609.x line.

PR #277, opened 2015-12-16
[JENKINS-30395] Preventing a recurrent ClosedByInterruptException in CI

jglick:
Diagnosis for https://github.com/jenkinsci/jenkins/pull/1826 failed, but found a workaround: clearing the interrupted flag on the current thread before proceeding with the test.

Will file a corresponding upstream patch soon, but fixing this in the `workflow-plugin` CI job is a top priority.

@reviewbybees

PR #278, opened 2015-12-16
[JENKINS-30395] Turn off reuseForks

jglick:
In tandem with #277, I want to check if tests are stable when we give each test class its own JVM.

@reviewbybees


amuniz:
Ok Alice, follow the rabbit :smile:

jglick:
Seems stabler…but an order of magnitude slower.

jglick:
Passes, but very nearly runs over even the expanded build timeout I configured. At least it helps confirm that the problem has something to do with JVM state carried over from one test suite to the next.

PR #279, opened 2015-12-17
[JENKINS-26126] GDSL, DSLD and static HTML Workflow steps docs

abayer:
- [X] GDSL.
- [X] Move the static HTML to a view rather than a big ol' string.
- [X] Add links to the GDSL, DSLD and static HTML.
- [ ] DSLD - still needs testing.
- [ ] Make the HTML not ugly. Halp?
- [ ] Special-case for `ParallelStep`.

cc @reviewbybees and @jglick in particular.


abayer:
Oh, and I'm not touching global variables yet. Again, working on the general guts first.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -329,6 +332,34 @@ private static void describeType(DescribableHelper.ParameterType type, PrintWrit
         }
     }
 
+    @Restricted(NoExternalUse.class)
+    public static final String GDSL_URL = ACTION_URL + "/gdsl";
+
+    @Restricted(DoNotUse.class)
+    public void doGdsl(StaplerRequest req, StaplerResponse rsp) throws Exception {
+        rsp.setContentType("text/plain;charset=UTF-8");
+        req.getView(this,"gdsl.groovy").forward(req, rsp);
```
jglick: Can delete this method. Stapler should find the view anyway. (use `st:header` as needed)

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -331,7 +342,7 @@ static ParameterType of(Type type) {
                     }
                 }
                 if (acceptsList(type)) {
-                    return new ArrayType(of(((ParameterizedType) type).getActualTypeArguments()[0]));
+                    return new ArrayType(List.class, of(((ParameterizedType) type).getActualTypeArguments()[0]));
```
jglick: Might actually be `Collection`.

(Do you want to give `Type` rather than `Class`?)

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -419,6 +434,7 @@ public Schema getType() {
         private final Class<?> supertype;
         private final Map<String,Schema> types;
         HeterogeneousObjectType(Class<?> supertype, Map<String,Schema> types) {
+            super(supertype);
             this.supertype = supertype;
```
jglick: Then you could delete this field and getter.

cps/src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
```diff
@@ -216,4 +216,27 @@ private void assertGenerateSnippet(@Nonnull String json, @Nonnull String respons
...
+    @Issue("JENKINS-26126")
+    @Test public void doDsld() throws Exception {
+        JenkinsRule.WebClient wc = r.createWebClient();
+        String dsld = wc.goTo(Snippetizer.DSLD_URL, "text/plain").getWebResponse().getContentAsString();
+        assertThat("Description is included as doc", dsld, containsString("Build a job"));
+        assertThat("Timeout step appears", dsld, containsString("name: 'timeout'"));
+
+        // Verify valid groovy sntax.
+        GroovyShell shell = new GroovyShell(r.jenkins.getPluginManager().uberClassLoader);
+        shell.parse(dsld);
```
jglick: I thought this was some other format?

jglick:
Merge with `master` to pick up my CI stability fix.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: Not sure what the difference would be between `Type` and `Class`... I'll switch to `Collection`, though, since that sounds right. More tweaking remains to be done on the parameter typing in the DSL.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
abayer: Done.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: Not sure what you mean - whether we're using `Type` or `Class` for the `actualClass`, it still needed to be a new variable in `ParameterType`...

cps/src/test/java/org/jenkinsci/plugins/workflow/cps/SnippetizerTest.java
abayer: DSLD and GDSL are both Groovy - the output's different, but still Groovy in both cases.

abayer:
Force-pushed after a rebase.

abayer:
Added a new .../staticHelp URL, probably going to replace the hard-coded HTML .../static one with it eventually. The new one is a Groovy view, pulls in help properly, loads the sidepanel.jelly, etc.

abayer:
Squashed down my 14 (!!!) commits since, well, no one had really started reviewing in earnest yet and this makes the history a lot less messy. I'll try to limit my future commits to bigger ones so as to keep from sprawling out again. =)

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -272,7 +272,17 @@ public String getDisplayName() {
      * A type of a parameter to a class.
      */
     public static abstract class ParameterType {
-        ParameterType() {}
+        @Nonnull
+        private Type actualClass;
+
+        public Type getActualClass() {
```
jglick: Confusing naming. `Class` is only one kind of `Type`.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -272,7 +272,17 @@ public String getDisplayName() {
      * A type of a parameter to a class.
      */
     public static abstract class ParameterType {
-        ParameterType() {}
+        @Nonnull
+        private Type actualClass;
```
jglick: `final`?

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -343,6 +353,7 @@ static ParameterType of(Type type) {
     public static final class AtomicType extends ParameterType {
         private final Class<?> clazz;
         AtomicType(Class<?> clazz) {
+            super(clazz);
```
jglick: Then it is redundant to keep the `clazz` field.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -360,6 +371,7 @@ static ParameterType of(Type type) {
         private final Class<?> clazz;
         private final String[] values;
         EnumType(Class<?> clazz, String[] values) {
+            super(clazz);
```
jglick: Ditto.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -382,7 +394,8 @@ static ParameterType of(Type type) {
 
     public static final class ArrayType extends ParameterType {
         private final ParameterType elementType;
-        ArrayType(ParameterType elementType) {
+        ArrayType(Class<?> actualClass, ParameterType elementType) {
```
jglick: Better to

```java
ArrayType(Class<?> actualClass) {
    super(actualClass);
    elementType = of(actualClass.getComponentType());
}
```

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -398,7 +411,8 @@ public ParameterType getElementType() {
 
     public static final class HomogeneousObjectType extends ParameterType {
         private final Schema type;
-        HomogeneousObjectType(Schema type) {
+        HomogeneousObjectType(Class<?> actualClass, Schema type) {
```
jglick: ditto

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
jglick: I meant that `HeterogeneousObjectType.supertype` could be deleted, with a typesafe getter retained like

```java
public Class<?> getSupertype() {
    return (Class) getType(); // from superclass
}
```

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
```diff
@@ -231,103 +228,16 @@ public HttpResponse doGenerateSnippet(StaplerRequest req, @QueryParameter String
     }
 
     @Restricted(NoExternalUse.class)
-    public static final String STATIC_URL = ACTION_URL + "/static";
+    public static final String GDSL_URL = ACTION_URL + "/gdsl";
```
jglick: If you are switching to Groovy views, you can get rid of these constants, inlining their code. I defined `STATIC_URL` for the benefit of the Jelly view.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
```diff
@@ -50,6 +50,7 @@ THE SOFTWARE.
             </j:forEach>
         </d:tag>
     </d:taglib>
+    <f:entry title="Workflow DSL Reference" help="/${it.DSL_REF_URL}"/>
```
jglick: I suppose this is intended to supersede the block at lines 98–100, which you forgot to delete (and which is still using the deleted `STATIC_URL`).

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslHelp.groovy
```diff
@@ -0,0 +1,33 @@
+package org.jenkinsci.plugins.workflow.cps.Snippetizer
+
+
+def l = namespace(lib.LayoutTagLib)
+def st = namespace("jelly:stapler")
+
+st.contentType(value: "text/html;charset=UTF-8")
+
+l.layout(title:_("Jenkins Workflow Reference"),permission:app.READ) {
```
jglick: Without `Jenkins.READ` you could not get to this view via Stapler to begin with, so the `permission` is redundant.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
jglick: Anyway I would recommend just using a `f:block` and inline that content here (or use `st:include` if you want it in a separate source file).

jglick: BTW I would suggest swapping the names of `dslHelp` and `dslReference`.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslHelp.groovy
```diff
@@ -0,0 +1,33 @@
+package org.jenkinsci.plugins.workflow.cps.Snippetizer
+
+
+def l = namespace(lib.LayoutTagLib)
+def st = namespace("jelly:stapler")
+
+st.contentType(value: "text/html;charset=UTF-8")
```
jglick: :bug: automatic from `l.layout`.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslHelp.groovy
```diff
@@ -0,0 +1,33 @@
...
+
+l.layout(title:_("Jenkins Workflow Reference"),permission:app.READ) {
+
+    st.include(page: "sidepanel", it: app)
+    l.main_panel {
+        p {
+            raw("Click ")
+            a(href: "${app.rootUrl}${my.GDSL_URL}", target: "_blank") {
```
jglick: :bug: use `rootURL`, not `app.rootUrl`. [prime directive](https://wiki.jenkins-ci.org/display/JENKINS/Hyperlinks+in+HTML)

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslHelp.groovy
```diff
@@ -0,0 +1,33 @@
...
+                raw("here for IntelliJ GDSL")
```
jglick: :ant: use `_` not `raw`. And do not use separate sentence fragments, as this is not localizable.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslHelp.groovy
```diff
@@ -0,0 +1,33 @@
...
+
+        p {
+            raw("Click ")
+            a(href: "${app.rootUrl}${my.DSLD_URL}", target: "_blank") {
+                raw("here for Eclipse DSLD")
+            }
+            raw(".")
+
+        }
+        st.include(page: "dslReferenceContent", it: my)
```
jglick: I think `it`/`my` is automatically passed during `st.include`, no?

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslReferenceContent.groovy
```diff
@@ -0,0 +1,151 @@
...
+import javax.servlet.RequestDispatcher
+
+Snippetizer snippetizer = my;
+
+def l = namespace(lib.LayoutTagLib)
+def st = namespace("jelly:stapler")
+
+
+h1(_("Steps"))
+for (StepDescriptor d : snippetizer.getStepDescriptors(false)) {
```
jglick: Sigh, I was hoping to _not_ have all this crap in Groovy. Maybe it is the only way to get `RequestDispatcher` to work? If so, ick, [JENKINS-31203](https://issues.jenkins-ci.org/browse/JENKINS-31203).

jglick:
> It's still erroring on ParallelStep due to a lack of a Descriptor

Well, it has a `Descriptor`, but this step intentionally does not limit itself to the syntax supported for general steps, and `Snippetizer` cannot deal with it mechanically. You have to special-case it.

jglick:
Title and description of PR both seem out of date now. I would recommend using

```markdown
- [X] checkboxes
- [X] as needed
- [ ] to track your progress
```

abayer:
Title/summary updated.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: Renaming to `actualType` everywhere.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: Will do.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: okiedokie, removing.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: Changing.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: Changing.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: Hrm, actually, we've got `return new ArrayType(Collection.class, of(((ParameterizedType) type).getActualTypeArguments()[0]));` in `ParameterType.of` - I think we still need this?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/Snippetizer.java
abayer: Okiedokie.

abayer: Actually, these are still relevant in the tests...

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
abayer: Deleting 98-100. Not sure what you mean by swapping, though.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslHelp.groovy
abayer: Okiedokie.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslHelp.groovy
abayer: Fixed.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslHelp.groovy
abayer: Okiedokie.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslHelp.groovy
abayer: Ok, so make the whole thing a link, then?

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslHelp.groovy
abayer: Okiedokie.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslReferenceContent.groovy
abayer: So the reason I did it in Groovy is that, well, it was a lot easier to do it here in the view than to pass it to the view in some form and handle all the rendering logic separately in the view. I could certainly be wrong and there may well be a less awkward approach, but, well, this worked. =)

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: Removing.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: Ah, ok.

abayer:
Rebased, pushed new commit attempting to address comments/concerns.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
jglick: > Not sure what you mean by swapping, though.

Well `dslHelp` is really the reference. And `dslReference` is what you see when you click the help button. So the view names seem backwards.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslHelp.groovy
```diff
@@ -0,0 +1,27 @@
...
+
+def l = namespace(lib.LayoutTagLib)
+def st = namespace("jelly:stapler")
+
+l.layout(title:_("Jenkins Workflow Reference")) {
+
+    st.include(page: "sidepanel", it: app)
+    l.main_panel {
+        p {
+            a(href: "${app.rootUrl}${my.GDSL_URL}", target: "_blank") {
```
jglick: :bug: must use `rootURL` in pages served from Jenkins. Never use `app.rootUrl` except when generating HTML to be sent elsewhere. (I thought I mentioned this already? [wiki](https://wiki.jenkins-ci.org/display/JENKINS/Hyperlinks+in+HTML))

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslReference.groovy
```diff
@@ -0,0 +1,28 @@
+package org.jenkinsci.plugins.workflow.cps.Snippetizer
+
+
+def st = namespace("jelly:stapler")
+
+p {
+    a(href: "${rootUrl}${my.DSL_HELP_URL}", target: "_blank") {
```
jglick: I think this is `rootURL`, not `rootUrl`.

jglick: And BTW from Groovy you can use real Java syntax. So `my.DSL_HELP_URL` could be replaced with `Snippetizer.DSL_HELP_URL` probably.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslReferenceContent.groovy
jglick: I just really wanted to avoid having any nontrivial logic in Groovy views. There is limited IDE support, no build-time type checking, etc. That is why I had the original HTML generation in Java code. The one problem that approach had was that I could not figure out how to call from Java code back into Stapler for rendering help views of global variables (the `st.include(page: "help", it: v)` here).

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -331,58 +341,50 @@ static ParameterType of(Type type) {
...
 
     public static final class AtomicType extends ParameterType {
-        private final Class<?> clazz;
         AtomicType(Class<?> clazz) {
-            this.clazz = clazz;
-        }
-        /**
-         * A Java type: either {@link String}, or a {@link Class#isPrimitive} wrapper.
-         */
-        public Class<?> getType() {
```
jglick: My recommendation was to leave the `getType` method, or some variant, so that a caller would not need to cast `getActualType()` to `Class`.

jglick: (And similarly for other `ParameterType`s which have a constrained `type`.)

jglick: (For example as you did for `HeterogeneousObjectType`.)

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
jglick: BTW clicking the help link currently shows

> **ERROR**: Failed to load help file: Internal Server Error

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslReferenceContent.groovy
abayer: So my thinking is that I'd rather have the limitations of Groovy views over trying to generate Groovy and HTML via `String`s in Java. But in the end, I'll defer to you - I just couldn't find a way to do the GDSL/DSLD in particular that I felt comfortable with, and I felt *much* more comfortable doing the HTML in the view as well. *shrug*

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslHelp.groovy
abayer: Missed that one. Fixing.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslReference.groovy
abayer: Fixing both.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: I'm moving this to be in `ParameterType` proper, renaming `HomogenousObjectType.getType` to `getSchemaType`, removing `HeterogenousObjectType.getSupertype` and changing references to that to use `getType` instead.

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/block.jelly
abayer: Flipping 'em, then doing some testing...

abayer:
Ok, it's working - note that there are a bunch of errors that show up on the console due to specific types failing...

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -272,7 +272,21 @@ public String getDisplayName() {
...
-        ParameterType() {}
+        @Nonnull
+        private final Type actualType;
+
+        public Type getActualType() {
+            return actualType;
+        }
+
+        public Class<?> getType() {
+            return (Class) getActualType();
```
jglick: :bug: this should be pushed down into those subclasses which can actually _guarantee_ that the type is in fact a `Class`, and not something else.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -331,58 +345,50 @@ static ParameterType of(Type type) {
                     }
                 }
                 if (acceptsList(type)) {
-                    return new ArrayType(of(((ParameterizedType) type).getActualTypeArguments()[0]));
+                    return new ArrayType(Collection.class, of(((ParameterizedType) type).getActualTypeArguments()[0]));
```
jglick: Should this not be passing `type` as the `actualType`?

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: Ok, I thought I was being tricky. =)

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: Yeah, we can give that a try.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: Which ones can't guarantee it's a `Class`?

jglick: Well, the `ArrayType` I would expect to have a `ParameterizedType` here, so casting to `Class` would throw `ClassCastException`.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: Nope, that ended up causing explosions. Fun!

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: As of now, `ArrayType` is taking a `Class<?>` for `actualClass`...I probably changed that to be that way but can switch to be a `Type` instead. Your call.

jglick: I would expect `actualType` to be the `type` passed to `of`, which is a `Class` except in the second (collection-oriented) call to the `ArrayType` constructor, when it is a `ParameterizedType`.

jglick: (or could be whatever in the case of `ErrorType`)

abayer: Ok. Tweaking, then.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
abayer: Tweaked in re the above, new commit incoming.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -331,44 +341,40 @@ static ParameterType of(Type type) {
...
     public static final class AtomicType extends ParameterType {
-        private final Class<?> clazz;
         AtomicType(Class<?> clazz) {
-            this.clazz = clazz;
+            super(clazz);
         }
-        /**
-         * A Java type: either {@link String}, or a {@link Class#isPrimitive} wrapper.
-         */
+
```
jglick: Of course now you have come full circle and deleted my nice Javadoc. :-) No matter.

jglick:
:bee: assuming the visible stuff more or less works as advertised!

jglick:
From the GDSL,

```groovy
method(name: 'checkout', type: 'Object', namedParams: [parameter(name: 'scm', type: 'Map'), parameter(name: 'changelog', type: 'boolean'), parameter(name: 'poll', type: 'boolean'), ], doc: 'General SCM')
```

is not terribly helpful, but maybe it is best we can do given the schema language, at least pending [JENKINS-29922](https://issues.jenkins-ci.org/browse/JENKINS-29922)?

jglick:
So direct browsing to `/workflow-cps-snippetizer/dsld` is available for those who wish to experiment with DSLD.

jglick:
HTML works, and is a little better than in trunk—so good enough for now.

abayer:
re the GDSL - yeah, I'm also not sure how crazily nested we can go in the GDSL parameters.

abayer:
Rebased on top of the Workflow->Pipeline rename.

jglick:
:bee: assuming your rebase was correct. (Generally I find PRs to be unreviewable when destructive history operations are used. Plain old `git merge master` works fine.)

cps/src/main/resources/org/jenkinsci/plugins/workflow/cps/Snippetizer/dslReferenceContent.groovy
kohsuke: I share a similar sentiment with @jglick that this view contains too much logic that ideally should be in Java classes, then Groovy views can be used to render them, instead of `String` in Java, which is what @abayer is saying.

But such refactoring can come in later.

kohsuke:
:bee:
Let's get something out and iterate over

abayer:
@jglick I will rebase until forced not to. =)

amuniz:
Just noting I was not able to make the DSLD to work in Eclipse (using [Eclipse Groovy](https://github.com/groovy/groovy-eclipse/wiki)). I think I did all the configuration well, so I guess the DSLD is not valid for some reason.

PR #281, opened 2015-12-18
[JENKINS-32067] Use ACE editor annotations to show Groovy syntax errors

amuniz:
[JENKINS-32067](https://issues.jenkins-ci.org/browse/JENKINS-32067)

screenrec

@reviewbybees


cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
```diff
@@ -127,5 +138,28 @@ public FormValidation doCheckScript(@QueryParameter String value, @QueryParamete
             return sandbox ? FormValidation.ok() : ScriptApproval.get().checking(value, GroovyLanguage.get());
         }
 
+        public String doCheckScriptJson(@QueryParameter String value, @QueryParameter boolean sandbox) {
```
jglick: I think you can just return `JSONObject` (or `JSON`) directly.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
```diff
@@ -127,5 +138,28 @@ public FormValidation doCheckScript(@QueryParameter String value, @QueryParamete
...
+            try {
+                new CpsGroovyShell(null).getClassLoader().parseClass(value);
+            } catch (CompilationFailedException x) {
+                return JSONArray.fromObject(CpsFlowDefinitionValidator.toCheckStatus(x).toArray()).toString();
+            }
+            if (sandbox) {
+                return CpsFlowDefinitionValidator.CheckStatus.SUCCESS.toString();
+            } else {
+                FormValidation checking = ScriptApproval.get().checking(value, GroovyLanguage.get());
+                if (StringUtils.isBlank(checking.getMessage())) {
```
jglick: I think you want to compare the result to `OK`.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionValidator.java
```diff
@@ -0,0 +1,110 @@
...
+        public int getLine() {
+            return line;
+        }
+        public void setLine(int line) {
+            this.line = line;
+        }
+        public int getColumn() {
+            return column;
+        }
+        public void setColumn(int column) {
```
jglick: Could delete all the setters and just use a constructor.

cps/src/main/js/workflow-editor.js
```diff
@@ -48,7 +48,46 @@ jenkinsJSModules.import('ace-editor:ace-editor-122')
...
+                    editor.session.clearAnnotations();
+                    var url = textarea.attr("checkUrl") + 'Json';
+
+                    $.ajax({
+                        url: url,
+                        data: {
+                            value: editor.getValue(),
+                            sandbox: $("input[name='_.sandbox']").is(':checked')
+                        },
+                        method: textarea.attr("checkMethod") || "get",
```
jglick: May as well hardcode `POST`. You do not want `GET` as it will not handle large scripts.

cps/src/main/js/workflow-editor.js
```diff
@@ -48,7 +48,46 @@ jenkinsJSModules.import('ace-editor:ace-editor-122')
...
+                                // Top level statuses
+                                if (data.status === 'success') {
+                                    return;
+                                }
+                                if (data.status === 'approval') {
+                                    annotations.push({
+                                        row: 0,
+                                        column: 0,
+                                        text: data.message,
+                                        type: "warning"
```
jglick: This would better be served with form validation I think. Simpler code too: the JSON response could concentrate on Groovy errors.

jglick:
:bee:

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionValidator.java
amuniz: I followed the "only required fields in constructor" as in `@DataboundConstrucotr` (unrelated here, I know, but sensible). `column` is not required under some circumstances, so having a constructor with this field will lead in weird calls to `new CheckStatus("mesg", "success", 0, 0)` where `0, 0` does not really have any semantic. That's why I added this setters.

amuniz:
The script approval validation is now excluded - it is done as it was before in `doCheckScript`.

@jglick could you review the last commit please?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinitionValidator.java
```diff
@@ -0,0 +1,111 @@
...
+            this.column = column;
+        }
+
+        public JSONObject asJSON() {
+            return JSONObject.fromObject(this);
```
jglick: Would be clearer to either have all the JSON creation in `doCheckScriptJson`, or all of it here.

cps/src/main/js/workflow-editor.js
```diff
@@ -48,7 +48,34 @@ jenkinsJSModules.import('ace-editor:ace-editor-122')
                 });
 
                 editor.on('blur', function() {
-                    textarea.trigger('change');
```
jglick: :bug: AFAICT this is still suppressing `doCheckScript`, which you now again rely on for script approval warnings.

amuniz: Well, it's true but `doCheckScript` is still called when you check/uncheck the sandbox checkbox (sorry for the tongue-twister).

Sandbox is activated by default, so if you write a script requiring approval and uncheck the sandbox then regular validation is shown.

Do we want to run sandboxing related validation on textarea changes too? If so, we have to go back to the previous commit and manage both approval and syntax validations in the JSON callback.

jglick: When `sandbox` is _unchecked_ then any change to `script` requires `ScriptApproval` to be consulted for `FormValidation`, in case a non-admin is editing the script.

I do not get why you cannot simply restore the call to `trigger` here, so `doCheckScript` will verify script approval exactly as it did before this PR, while `doCheckScriptJson` will handle line-by-line Groovy errors.

amuniz: > I do not get why you cannot simply restore the call to trigger here

I could, but then two AJAX calls will be done on textarea `blur`, one to `doCheckScript` and a second one to `doCheckScriptJson` which is a kind of duplicated work to validate the script. Moreover, the ugly error message under the textarea will be shown (which is redundant since we have the error in-lined in the editor).

I think the best option is to remove the last commit and show both syntax errors and approval requirements as ACE editor annotations. We could even remove the old `doCheckScript`. Do you agree?

jglick: Obviously you would delete the Groovy compilation from `doCheckScript`.

Approval warnings just look weird as editor annotations, when it is not about a specific line.

You could use one AJAX method if you dug up the code to display `FormValidation` from `hudson-behavior.js`.

amuniz: > Obviously you would delete the Groovy compilation from doCheckScript.

Ok, it was not so obvious to me - it will break compatibility. I hope nobody is relying on the existence of `doCheckScript` to request some ad-hoc script syntax validation (for some weird reason).

So, let's go that way.

jglick: > I think the best option is to remove the last commit and show both syntax errors and approval requirements as ACE editor annotations.

Not ideal, but an option if it is looking hard to use the normal `FormValidation` UI: script approval warnings are not common in practice, since if you are not an admin the sandbox mode is checked by default, so not many people would see this anyway.

tfennelly:
:bee: 

Nice :smile: :+1:

jglick:
:bug: pending resolution of script approval warnings.

amuniz:
@jglick I finally implemented it as you suggested. Now two AJAX calls are done on each change in the text area, but the second call - to check for approval requirements - is only done if the script is syntactically correct (result of the first call).

amuniz:
The test failure seems unrelated...

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java
```diff
@@ -119,12 +131,21 @@ public FormValidation doCheckScript(@QueryParameter String value, @QueryParamete
             if (j == null) {
                 return FormValidation.ok();
             }
+            return sandbox ? FormValidation.ok() : ScriptApproval.get().checking(value, GroovyLanguage.get());
+        }
+
+        public JSON doCheckScriptCompile(@QueryParameter String value, @QueryParameter boolean sandbox) {
```
jglick: :ant: you do not need the `sandbox` parameter here AFAICT.

cps/src/main/js/workflow-editor.js
```diff
@@ -48,7 +48,36 @@ jenkinsJSModules.import('ace-editor:ace-editor-122')
...
+                    editor.session.clearAnnotations();
+                    var url = textarea.attr("checkUrl") + 'Compile';
+
+                    $.ajax({
+                        url: url,
+                        data: {
+                            value: editor.getValue(),
+                            sandbox: $("input[name='_.sandbox']").is(':checked')
```
jglick: :ant: as above.

jglick:
> The test failure seems unrelated

Does look that way, but it also failed five times in a row, so I want a rebuild to check. Other than that, :bee:

amuniz:
The same test fails again, going to check locally.

amuniz:
Just checked that the same test is failing locally under:
 - This PR as it is.
 - This PR after merging master.
 - master.

There is some test dependency issue. I don't know why other PRs are building successfully, they shouldn't.

jglick:
:bee:

If you claim the test is failing locally in `master`, then I presume this PR is not introducing the regression, but please take the lead on making it pass.

amuniz:
Ok, will do it.

@reviewbybees done

jglick:
> I presume this PR is not introducing the regression

Not so sure now! `doCheckScript` is the method being called in the failing line.

jglick:
IOW I would _expect_ that test to fail until updated. Not sure why it would be failing for you in `master`—are you positive that is the same cause? After a clean build?

jglick:
:bug: for now.

amuniz:
As you pointed, the exception is printed when running on master, but the test passes. Going to check why it make the test fail in the PR.

@reviewbybees undone :smile:

amuniz:
All that exceptions distracted me. Fixed.

jglick:
:bee: assuming the tests now pass.

amuniz:
@reviewbybees done

PR #282, opened 2015-12-19
Add a pointer to some example workflows

rodrigc:


jglick:
Yes, just saw this now, sorry.

PR #285, opened 2015-12-30
[JENKINS-32225] - Add the functionality for Node usage tracking

oleg-nenashev:
- [x] - Add FlowListener extension point
- [x] - Add method allowing to retrieve executors from FlowExecutions
- [x] - Unit test demonstrating the use-case

http://issues.jenkins-ci.org/browse/JENKINS-32225

@reviewbybees

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
@@ -25,6 +25,7 @@
 package org.jenkinsci.plugins.workflow.steps;
 
 import hudson.model.Computer;
+import hudson.model.Executor;
 import hudson.model.Node;
 import hudson.model.Queue;
 import hudson.model.User;
@@ -50,12 +51,15 @@ import java.util.logging.ConsoleHandler;
 import java.util.logging.Handler;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.CheckForNull;
 import jenkins.model.Jenkins;
 import org.apache.tools.ant.util.JavaEnvUtils;
 import org.jenkinsci.plugins.workflow.SingleJobTestBase;
 import org.jenkinsci.plugins.workflow.actions.WorkspaceAction;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
+import org.jenkinsci.plugins.workflow.cps.nodes.StepStartNode;
+import org.jenkinsci.plugins.workflow.flow.FlowListener;
 import org.jenkinsci.plugins.workflow.graph.FlowGraphWalker;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
@@ -66,11 +70,16 @@ import org.junit.AfterClass;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static org.hamcrest.Matchers.*;
+import org.jenkinsci.plugins.workflow.support.steps.ExecutorStep;
+import org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution;
+import static org.junit.Assert.assertThat;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 import org.junit.runners.model.Statement;
 import org.jvnet.hudson.test.Issue;
+import org.jvnet.hudson.test.TestExtension;
 
 /** Tests pertaining to {@code node} and {@code sh} steps. */
 public class ExecutorStepTest extends SingleJobTestBase {
@@ -414,5 +423,80 @@ public class ExecutorStepTest extends SingleJobTestBase {
             }
         });
     }
+    
+    /**
+     * Ensures that {@link FlowListener} callback is being invoked for the {@link ExecutorStepExecution}.
+     * It allows to record the executor, which is being utilized by the flow.
+     */
+    @Test
+    public void shouldInvokeFlowListenerCallbacksAndRecordNodes() {
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                DumbSlave s = createSlave(story.j);
+                s.setLabelString("remote quick");
+                s.getNodeProperties().add(new EnvironmentVariablesNodeProperty(
+                        new EnvironmentVariablesNodeProperty.Entry("ONSLAVE", "true")));
 
+                p = jenkins().createProject(WorkflowJob.class, "demo");
+                p.setDefinition(new CpsFlowDefinition(
+                    "node('" + s.getNodeName() + "') {\n" +
+                    "    sh('echo before=`basename $PWD`')\n" +
+                    "    sh('echo ONSLAVE=$ONSLAVE')\n" +
+                    "}"));
+
+                // Run build and wait for its completion
+                startBuilding();
+                while (!e.isComplete()) {
+                    e.waitForSuspension();
+                }
+
+                // Get the node invocation monitor
+                NodeInvocationListener listener = story.j.jenkins.getExtensionList(FlowListener.class)
+                        .get(NodeInvocationListener.class);
+                assertThat("NodeInvocationListener must be specified in the test", listener, notNullValue());
+                Node builtOn = listener.getBuiltOn();
+                assertThat("Node has not been recorded", builtOn, notNullValue());
+                assertThat("Node name differs from the expected one", builtOn.getNodeName(), equalTo(s.getNodeName()));
+            }
+        });
+    }
+    
+    /**
+     * Test extension, which locates the last usage of node declared by {@link ExecutorStep}.
+     */
+    @TestExtension("shouldInvokeFlowListenerCallbacksAndRecordNodes")
+    public static class NodeInvocationListener extends FlowListener {
+    
+        private Node builtOn;
+
+        @CheckForNull
+        public Node getBuiltOn() {
+            return builtOn;
+        }
+        
+        @Override
+        public void onNewHead(FlowNode node) { 
+            // We try to catch nodes, which indicate the start of the node() execution
+            if (!(node instanceof StepStartNode)) {
+                return;
+            }
+            
+            List<FlowNode> parents = node.getParents();
+            if (parents.isEmpty()) {
+                // Should never happen
+                return;
+            }
+            FlowNode firstParent = parents.get(0);
+            if (firstParent instanceof StepStartNode) { // Parent may be an ExecutorStep
+                StepStartNode candidate = (StepStartNode)firstParent;
+                if (candidate.getDescriptor() instanceof ExecutorStep.DescriptorImpl) {
+                    // We found ExecutorStep => we are within the node() call, which has been initialized
+                    Executor found = ExecutorStepExecution.findExecutor(candidate.getExecution());
+                    
+                    // Record the discovered node
+                    builtOn = found != null ? found.getOwner().getNode() : null;
+                }
+            }
+        }
+    }
 }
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowListener.java
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Notifies all listeners about the new flow node creation.
+     * @param node Newly created node
+     */
+    public static void fireNewHead(@Nonnull FlowNode node) {
+        for (FlowListener listener : all()) {
+            try {
+                listener.onNewHead(node);
+            } catch (Exception ex) { // Prevent failures on runtime exceptions
+                LOGGER.log(Level.SEVERE, "Runtime exception during the flow execution callback processing: " + listener, ex);
```
jglick: Copy `ItemListener.forAll` pending JENKINS-21224.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/GraphListener.java
```diff
@@ -27,6 +27,8 @@
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 
 /**
+ * A low-level implementation of the Flow listener.
+ * A classic Jenkins extension point is implemented as {@link FlowListener}.
```
jglick: Confusing. Why not just put the two new methods here in `GraphListener`?

api/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowListener.java
```diff
@@ -0,0 +1,125 @@
...
+public abstract class FlowListener implements ExtensionPoint {
+    
+    private final static Logger LOGGER = Logger.getLogger(FlowListener.class.getName());
+    
+    /**
+     * Notifies that a new execution node has been added to the flow.
+     * {@link FlowExecution} should batch up changes to a group and call this once, 
+     * as opposed to call this for every new node added.
+     *
+     * One of the use cases of this listener is to persist the state of {@link FlowExecution}.
```
jglick: Really? I do not think so. Looks like you just copied this from `GraphListener`.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
```diff
@@ -66,11 +70,16 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static org.hamcrest.Matchers.*;
+import org.jenkinsci.plugins.workflow.support.steps.ExecutorStep;
+import org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution;
```
jglick: Useless imports.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
```diff
@@ -66,11 +70,16 @@
...
+import static org.junit.Assert.assertThat;
```
jglick: Can just switch to a wildcard static import (see lines 70–72).

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
```diff
@@ -414,5 +423,80 @@ private void startJnlpProc() throws Exception {
             }
         });
     }
+    
+    /**
+     * Ensures that {@link FlowListener} callback is being invoked for the {@link ExecutorStepExecution}.
+     * It allows to record the executor, which is being utilized by the flow.
+     */
+    @Test
```
jglick:
```java
@Issue("JENKINS-32225")
```
?

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
```diff
@@ -414,5 +423,80 @@ private void startJnlpProc() throws Exception {
...
+        public void onNewHead(FlowNode node) { 
+            // We try to catch nodes, which indicate the start of the node() execution
+            if (!(node instanceof StepStartNode)) {
+                return;
+            }
+            
+            List<FlowNode> parents = node.getParents();
+            if (parents.isEmpty()) {
+                // Should never happen
+                return;
```
jglick:
```java
assertNotEqual(Collections.emptyList(), parents);
```

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -519,6 +520,49 @@ public String getUrl() {
             private static final long serialVersionUID = 1L;
         }
     }
+    
+    /**
+     * Looks for the executor, which is acquired by {@link ExecutorStep}.
+     * @param flowExecution Flow execution
+     * @return Found {@link Executor}. Null if it cannot be found
```
jglick: :bug: does not make sense. What is this supposed to return when there are two `node` blocks active in `parallel`?

I think you need to make it take `StepExecution`, or just be an instance method on `ExecutorStepExecution`.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
```diff
@@ -414,5 +423,80 @@ private void startJnlpProc() throws Exception {
...
+     * It allows to record the executor, which is being utilized by the flow.
+     */
+    @Test
+    public void shouldInvokeFlowListenerCallbacksAndRecordNodes() {
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                DumbSlave s = createSlave(story.j);
+                s.setLabelString("remote quick");
+                s.getNodeProperties().add(new EnvironmentVariablesNodeProperty(
+                        new EnvironmentVariablesNodeProperty.Entry("ONSLAVE", "true")));
```
jglick: How is this variable relevant to the test?

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
```diff
@@ -414,5 +423,80 @@ private void startJnlpProc() throws Exception {
...
+                    "node('" + s.getNodeName() + "') {\n" +
+                    "    sh('echo before=`basename $PWD`')\n" +
+                    "    sh('echo ONSLAVE=$ONSLAVE')\n" +
+                    "}"));
 
+                // Run build and wait for its completion
+                startBuilding();
+                while (!e.isComplete()) {
+                    e.waitForSuspension();
+                }
```
jglick: Deprecated idiom, do not copy. Just

```java
story.j.assertBuildStatusSuccess(p.scheduleBuild2(0));
```

jglick:
I am missing an overall description of what you are ultimately trying to accomplish here, and the JIRA issue does not give much background.

* The listener seems potentially useful, but
  * the `onStepExecutionStarted` and `onBodyExecutionStarted` events are both untested and unused
  * it might make more sense to use the existing `FlowExecution.addListener` for details, while adding a new API to notify global listeners of newly started or finished `FlowExecution`s (you can almost get there via `FlowExecutionOwner.Executable` + `RunListener` but you would need to poll on `getOrNull` unless `WorkflowRun.getExecutionPromise` were lifted into the interface)
* The method returning `Executor` strikes me as weird. And unnecessary: we already record `WorkspaceAction`, which should be all you need to determine from an `onNewHead` event that a node is being used. Without casting to `StepStartNode` or other tricks.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/GraphListener.java
oleg-nenashev: GraphListener is not a Jenkins extension point => it's external usage seems to be a hack.

api/src/main/java/org/jenkinsci/plugins/workflow/flow/FlowListener.java
oleg-nenashev: This extension point may be used as well, but I agree it's a rudiment to be deleted from Javadoc

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
oleg-nenashev: Agreed

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
oleg-nenashev: yep

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
oleg-nenashev: agreed

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
oleg-nenashev: Agreed. Maybe it was an over-engineering. I need to check why I did it

api/src/main/java/org/jenkinsci/plugins/workflow/flow/GraphListener.java
jglick: Well you could make it implement `ExtensionPoint`.

oleg-nenashev: It's problematic due to the current plugin's API. It's hard to prevent usage conflicts since there are explicit methods for adding GraphListeners

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
oleg-nenashev: Nope. These classes belong to <code>org.jenkinsci.plugins.workflow.SUPPORT</code>, so they are not visible by default

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
oleg-nenashev: Temporarily added some Javadoc

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
oleg-nenashev: yep, it's not required

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
oleg-nenashev: fixed

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
```diff
@@ -414,5 +420,77 @@ private void startJnlpProc() throws Exception {
...
+
+                // Run build and wait for its completion
+                story.j.assertBuildStatusSuccess(p.scheduleBuild2(0));
 
+                // Get the node invocation monitor
+                NodeInvocationListener listener = story.j.jenkins.getExtensionList(FlowListener.class)
+                        .get(NodeInvocationListener.class);
+                assertThat("NodeInvocationListener must be specified in the test", listener, notNullValue());
+                Node builtOn = listener.getBuiltOn();
+                assertThat("Node has not been recorded", builtOn, notNullValue());
```
amuniz: Using a negative sentence seems confusing, "Node has to be recorded" is better IMO so it reads "Assert that: node has to be recorded".

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/ExecutorStepTest.java
```diff
@@ -414,5 +420,77 @@ private void startJnlpProc() throws Exception {
...
+                assertThat("Node name differs from the expected one", builtOn.getNodeName(), equalTo(s.getNodeName()));
```
amuniz: Same here.

amuniz:
:bee:

jglick:
:bug: until the basic design is sorted out.

jglick:
`FlowListener` is probably salvageable with some minor work, if there is a use case for it. `getExecutor` does not make sense as an API. Closing since this seems abandoned.

oleg-nenashev:
@jglick Yes, the approach has been declined by the stakeholders. Currently we're working on another implementation

PR #289, opened 2016-01-05
[JENKINS-31614] Avoid acquiring the Queue lock while holding other locks

jglick:
[JENKINS-31614](https://issues.jenkins-ci.org/browse/JENKINS-31614)

Trying to cover all three known cases.

@reviewbybees


stephenc:
:bee: as I understand it

PR #291, opened 2016-01-07
getSCMDescriptors cleanup

jglick:
See https://github.com/jenkinsci/branch-api-plugin/pull/24 for background.

@reviewbybees


amuniz:
Why not remove the method? Since it is removed in https://github.com/jenkinsci/branch-api-plugin/pull/24. Or at least, remove the `@Override`.

jglick:
> Why not remove the method?

Cannot, it is `abstract`.

> Or at least, remove the `@Override`.

The annotation is useful as a reminder to delete the override when updating the `branch-api` dep.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
```diff
@@ -76,10 +80,22 @@ public WorkflowMultiBranchProject(ItemGroup parent, String name) {
         }
 
         @Override public List<SCMDescriptor<?>> getSCMDescriptors() {
-            // TODO this is a problem. We need to select only those compatible with Workflow.
-            // Yet SCMDescriptor.isApplicable requires an actual Job, whereas we can only pass WorkflowJob.class!
-            // Can we create a dummy WorkflowJob instance?
-            return SCM.all();
+            throw new IllegalStateException("TODO unused: https://github.com/jenkinsci/branch-api-plugin/pull/24");
```
oleg-nenashev: Maybe better to just return an empty list and log the error.
It's a public API, so no idea how it could be used by external plugins and groovy scripts

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
```diff
@@ -76,10 +80,22 @@ public WorkflowMultiBranchProject(ItemGroup parent, String name) {
...
+        }
+
+        @Override public boolean isApplicable(Descriptor descriptor) {
+            if (descriptor instanceof SCMDescriptor) {
+                SCMDescriptor d = (SCMDescriptor) descriptor;
+                // TODO would prefer to have SCMDescriptor.isApplicable(Class<? extends Job>)
+                try {
+                    if (!d.isApplicable(new WorkflowJob(null, null))) {
```
oleg-nenashev: OK, there is a notice about the hack below

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
jglick: The method is scheduled for _deletion_ so I would rather not be quiet about calls to it.

oleg-nenashev: Uh oh, ok

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectTest.java
```diff
@@ -116,4 +125,35 @@
...
+    @Test public void applicableSCMs() throws Exception {
+        final WorkflowMultiBranchProject mp = r.jenkins.createProject(WorkflowMultiBranchProject.class, "p");
+        List<Class> scmTypes = new ArrayList<Class>();
+        List<SCMDescriptor<?>> scmDescriptors = SingleSCMSource.DescriptorImpl.getSCMDescriptors(mp);
+        for (SCMDescriptor<?> scmDescriptor : scmDescriptors) {
+            scmTypes.add(scmDescriptor.clazz);
+        }
+        assertThat(scmTypes, hasItem((Class) GitSCM.class));
+        assertThat(scmTypes, not(hasItem((Class) OldSCM.class)));
+        /* More realistic variant:
```
oleg-nenashev: Makes sense to add TODO or something else to explain why we keep this commented code

jglick: Well my initial attempt at a fix was to use the equivalent of `@AncestorInPath` to look up the `WorkflowJob` instance. For that, a more realistic test was needed. Then I realized that such a trick would not work anyway, since when configuring a new multibranch project there is no branch project yet. So I just kept this in case some future iteration of the fix actually is more sensitive to the request environment.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
jglick: See upstream PR.

amuniz:
:bee:

PR #298, opened 2016-01-12
[JENKINS-28649] Add exported beans needed for Queue API

amuniz:
[JENKINS-28649](https://issues.jenkins-ci.org/browse/JENKINS-28649). Supersedes #137 

@reviewbybees

### DIFF
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
@@ -28,11 +28,19 @@ import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.runners.model.Statement;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.RestartableJenkinsRule;
+import org.xml.sax.SAXException;
+
 import static org.junit.Assert.*;
+
+import java.io.IOException;
+import java.util.concurrent.ExecutionException;
 
 import com.gargoylesoftware.htmlunit.Page;
 
@@ -43,30 +51,115 @@ import net.sf.json.JSONObject;
 
 public class BuildQueueTasksTest {
 
-    @Rule public JenkinsRule r = new JenkinsRule();
+    @Rule public RestartableJenkinsRule story = JenkinsRuleExt.workAroundJenkins30395Restartable();
 
     @Issue("JENKINS-28649")
     @Test public void queueAPI() throws Exception {
-        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
-        // try to use non-existent node to keep the build queued
-        p.setDefinition(new CpsFlowDefinition("node('linux') { echo 'test' }"));
+        // This is implicitly testing ExecutorStepExecution$PlaceholderTask as exported bean
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                // use non-existent node label to keep the build queued
+                p.setDefinition(new CpsFlowDefinition("node('linux') { echo 'test' }"));
+
+                WorkflowRun b = scheduleAndWaitQueued(p);
+                assertQueueAPIStatusOKAndAbort(b);
+            }
+        });
+    }
+
+    @Issue("JENKINS-28649")
+    @Test public void queueAPIRestartable() throws Exception {
+        // This is implicitly testing AfterRestartTask as exported bean
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                // use non-existent node label to keep the build queued
+                p.setDefinition(new CpsFlowDefinition("node('linux') { echo 'test' }"));
+                scheduleAndWaitQueued(p);
+                // Ok, the item is in he queue now, restart
+            }
+        });
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.getItemByFullName("p", WorkflowJob.class);
+                WorkflowRun b = p.getBuildByNumber(1);
+
+                assertQueueAPIStatusOKAndAbort(b);
+            }
+        });
+    }
+
+    @Issue("JENKINS-28649")
+    @Test public void computerAPI() throws Exception {
+        // This is implicitly testing ExecutorStepExecution$PlaceholderTask$PlaceholderExecutable as exported bean
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                p.setDefinition(new CpsFlowDefinition(
+                        "node {\n" +
+                        "  echo 'test'\n " +
+                        "  semaphore 'watch'\n " +
+                        "}"));
+
+                WorkflowRun b = p.scheduleBuild2(0).getStartCondition().get();
+                SemaphoreStep.waitForStart("watch/1", b);
+
+                assertComputerAPIStatusOK();
+
+                SemaphoreStep.success("watch/1", null);
+                story.j.waitUntilNoActivity();
+            }
+        });
+    }
+
+    @Issue("JENKINS-28649")
+    @Test public void computerAPIRestartable() throws Exception {
+        // This is implicitly testing AfterRestartTask#run as exported bean
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                p.setDefinition(new CpsFlowDefinition(
+                        "node {\n" +
+                        "  echo 'test'\n " +
+                        "  semaphore 'watch'\n " +
+                        "}"));
+
+                WorkflowRun b = p.scheduleBuild2(0).getStartCondition().get();
+                SemaphoreStep.waitForStart("watch/1", b);
+            }
+        });
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                assertComputerAPIStatusOK();
+
+                SemaphoreStep.success("watch/1", null);
+                story.j.waitUntilNoActivity();
+            }
+        });
+    }
+
+    private WorkflowRun scheduleAndWaitQueued(WorkflowJob p) throws InterruptedException, ExecutionException {
         QueueTaskFuture<WorkflowRun> build = p.scheduleBuild2(0);
 
         WorkflowRun b = build.getStartCondition().get();
-        CpsFlowExecution e = (CpsFlowExecution) b.getExecutionPromise().get();
         int secondsWaiting = 0;
         while (true) {
             if (secondsWaiting > 5) {
                 assertTrue("No item queued after 5 seconds", false);
             }
-            if (r.jenkins.getQueue().getItems().length > 0) {
+            if (story.j.jenkins.getQueue().getItems().length > 0) {
                 break;
             }
             Thread.sleep(1000);
             secondsWaiting++;
         }
+        return b;
+    }
 
-        JenkinsRule.WebClient wc = r.createWebClient();
+    private void assertQueueAPIStatusOKAndAbort(WorkflowRun b)
+            throws IOException, SAXException, InterruptedException, ExecutionException {
+        JenkinsRule.WebClient wc = story.j.createWebClient();
         Page queue = wc.goTo("queue/api/json", "application/json");
 
         JSONObject o = JSONObject.fromObject(queue.getWebResponse().getContentAsString());
@@ -75,6 +168,7 @@ public class BuildQueueTasksTest {
         // Not going into de the content in this test
         assertEquals(1, items.size());
 
+        CpsFlowExecution e = (CpsFlowExecution) b.getExecutionPromise().get();
         e.interrupt(Result.ABORTED);
 
         queue = wc.goTo("queue/api/json", "application/json");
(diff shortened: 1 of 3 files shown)


job/src/main/java/org/jenkinsci/plugins/workflow/job/AfterRestartTask.java
```diff
@@ -33,15 +33,18 @@
...
 /**
  * Represents a {@link WorkflowRun} still running after a Jenkins restart.
  * Could be a {@code ContinuedTask}, though not really necessary since it is a {@link FlyweightTask} which would never be blocked anyway.
  */
+@ExportedBean
 class AfterRestartTask extends AbstractQueueTask implements Queue.FlyweightTask, Queue.TransientTask {
 
-    private final WorkflowRun run;
-    
+    @Exported private final WorkflowRun run;
```
oleg-nenashev: Maybe better to implement the public getter method. Is there any reason to hide this field?

oleg-nenashev:
:bee: with a minor comment.
I would bee also happy to see a test for the AfterRestart task (or do we test it implicitly?)

job/src/main/java/org/jenkinsci/plugins/workflow/job/AfterRestartTask.java
amuniz: It was already `private`, I just added the `@Exported` annotation.

oleg-nenashev: Yes. I just wonder if it maies sense to make it public with a getter. May be useful for workflow-specific task listeners

jglick: @oleg-nenashev this class is already not `public`, so cannot be used as an API (anyway I would not want it to be).

Does `@Exported` support private members? This seems very uncomfortable. :ant: Prefer to make the field `public` to emphasize that it is being exported. Otherwise you would be led to believe that you could rename or restructure the field without consequence, which is not true.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -283,6 +286,7 @@ private Object readResolve() {
             return hasPermission(Item.CANCEL);
         }
 
+        @Exported
```
jglick: :bug: `@Exported` may be used on fields or getter methods only.

job/src/main/java/org/jenkinsci/plugins/workflow/job/AfterRestartTask.java
jglick: Or add a public getter method with the annotation.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
```diff
@@ -0,0 +1,77 @@
...
+        e.waitForSuspension();
+
+        JenkinsRule.WebClient wc = r.createWebClient();
+        Page queue = wc.goTo("queue/api/json", "application/json");
+
+        JSONObject o = JSONObject.fromObject(queue.getWebResponse().getContentAsString());
+        JSONArray items = o.getJSONArray("items");
+        assertEquals(1, items.size());
+        assertEquals(true, items.getJSONObject(0).getBoolean("stuck"));
+        assertEquals(true, items.getJSONObject(0).getBoolean("buildable"));
```
jglick: :ant: it does not seem that you are even testing the existence of the `@Exported` properties. May as well delete them if you are not going to verify them. (The reported priority bug is that the task as a whole lacks `@ExportedBean`, breaking even unrelated queries; lack of information would be an RFE.)

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
```diff
@@ -0,0 +1,77 @@
...
+    @Issue("JENKINS-28649")
+    @Test public void queueAPI() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        // try to use non-existent node to keep the build queued
+        p.setDefinition(new CpsFlowDefinition("node('linux') { echo 'test' }"));
+        QueueTaskFuture<WorkflowRun> build = p.scheduleBuild2(0);
+
+        WorkflowRun b = build.getStartCondition().get();
+        CpsFlowExecution e = (CpsFlowExecution) b.getExecutionPromise().get();
+        e.waitForSuspension();
```
jglick: :bug: do not use `CpsFlowExecution.waitForSuspension` in tests; it is unreliable. Rather schedule the build, then wait for the queue to become nonempty.

jglick:
:ant: no test for the modification to `AfterRestartTask`. Would I think need to use `RestartableJenkinsRule` for this.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
amuniz: Fixing. BTW It seems to be used [everywhere](https://github.com/jenkinsci/workflow-plugin/blob/3d85711de54cd43f1b703e5a9dc0345d23730ced/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/StageTest.java#L79) [in](https://github.com/jenkinsci/workflow-plugin/blob/53e756588642a39d4e2c74590c317f0a904b26b1/cps/src/test/java/org/jenkinsci/plugins/workflow/cps/CpsScriptTest.java#L26) [test](https://github.com/jenkinsci/workflow-plugin/blob/3d85711de54cd43f1b703e5a9dc0345d23730ced/aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryTest.java#L71) [code](https://github.com/jenkinsci/workflow-plugin/blob/3d85711de54cd43f1b703e5a9dc0345d23730ced/aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/steps/RestartingLoadStepTest.java#L73).

jglick: @kohsuke created it and once it was the only option. In most cases `SemaphoreStep` is simpler and more reliable, and I have over time replaced most—but not at all—original calls. In your case you cannot use that since you want to block (IIUC) between the start of the `node` step and the start of its body. We could if necessary introduce a utility allowing a test to wait for a `FlowNode` to appear with a specific step name, etc.

jglick:
:bug: the patch is incomplete. You have made the `Task`s (equivalent of `Job` for traditional projects) exportable, which is good because it allows `Queue.Item.task` to be written from `/queue/api`. But you have not made the `Executable`s (equivalent of `Run`) exportable, so `Executor.currentExecutable` cannot be written (unclear which API requests would hit this). Look carefully at the stack traces in both JENKINS-28649 and JENKINS-31471. Should be possible to reproduce both failure modes.

amuniz:
@jglick 
> the patch is incomplete

Fixed. And added tests for:

 1. `AfterRestartTask`
 1. `PlaceholderTask`
 1. `ExecutorStepExecution$PlaceholderTask$PlaceholderExecutable`
 1. `AfterRestartTask#run` (it's a `WorkflowRun` which is already `@ExportedBean`)

Is there any other case I'm missing?

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
```diff
@@ -0,0 +1,192 @@
...
+    private WorkflowRun scheduleAndWaitQueued(WorkflowJob p) throws InterruptedException, ExecutionException {
+        QueueTaskFuture<WorkflowRun> build = p.scheduleBuild2(0);
+
+        WorkflowRun b = build.getStartCondition().get();
+        int secondsWaiting = 0;
+        while (true) {
+            if (secondsWaiting > 5) {
+                assertTrue("No item queued after 5 seconds", false);
+            }
+            if (story.j.jenkins.getQueue().getItems().length > 0) {
```
jglick: :ant: `Queue.getItem(p) != null` is safer—will not falsely trigger if some other task gets scheduled. Not that it should, in a test like this, but still.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
```diff
@@ -0,0 +1,192 @@
...
+    @Rule public RestartableJenkinsRule story = JenkinsRuleExt.workAroundJenkins30395Restartable();
+
+    @Issue("JENKINS-28649")
+    @Test public void queueAPI() throws Exception {
+        // This is implicitly testing ExecutorStepExecution$PlaceholderTask as exported bean
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                // use non-existent node label to keep the build queued
+                p.setDefinition(new CpsFlowDefinition("node('linux') { echo 'test' }"));
```
jglick: I tend to use the label `nonexistent` for clarity.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+                WorkflowRun b = scheduleAndWaitQueued(p);
+                assertQueueAPIStatusOKAndAbort(b);
+            }
+        });
+    }
+
+    @Issue("JENKINS-28649")
+    @Test public void queueAPIRestartable() throws Exception {
+        // This is implicitly testing AfterRestartTask as exported bean
```
jglick: Is it really? I.e., did you actually confirm that this throws an exception mentioning `AfterRestartTask` when you comment out `@ExportedBean` there? I doubt it—`AfterRestartTask` is a replacement for `WorkflowJob` used when a build is restarted. It is flyweight, meaning it is moved onto the `master` node immediately upon the first call to `Queue.maintain`, so it would be difficult to even make a test pause long enough to reproduce an error related to it—normally it is gone from the queue almost immediately after being added.

And using a blocked `node` task as you did in `queueAPI` is pointless, since the only thing in the queue will be a `PlaceholderTask`, which you already tested. You could just have a script

```groovy
semaphore 'wait'
```

to force the build to stay running through the restart.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    @Issue("JENKINS-28649")
+    @Test public void computerAPIRestartable() throws Exception {
+        // This is implicitly testing AfterRestartTask#run as exported bean
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                p.setDefinition(new CpsFlowDefinition(
+                        "node {\n" +
```
jglick: Why the `node`?

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
```diff
@@ -0,0 +1,192 @@
...
+                        "  semaphore 'watch'\n " +
+                        "}"));
+
+                WorkflowRun b = p.scheduleBuild2(0).getStartCondition().get();
+                SemaphoreStep.waitForStart("watch/1", b);
+
+                assertComputerAPIStatusOK();
+
+                SemaphoreStep.success("watch/1", null);
+                story.j.waitUntilNoActivity();
```
jglick: Unnecessary.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
```diff
@@ -0,0 +1,192 @@
...
+            }
+        });
+    }
+
+    @Issue("JENKINS-28649")
+    @Test public void computerAPIRestartable() throws Exception {
```
jglick: :ant: why not just fold this into the previous test? `waitForStart`, do your assertions, restart, do your assertions again.

jglick:
:bee: assuming you can confirm that all the new tests actually test what you think they are testing.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
amuniz: You are right. I'm trying to modify this test to adapt it to what I wanted, although I don't know if it's possible.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
amuniz: The script is not the same in both, the first test requires a node but the second does not necessarily.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
amuniz: Easier like it is - given the conditions - as `p` is not  what I want to check in the queue but the `PlaceholderTask` instance and I don't see an straightforward way to retrieve it.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
amuniz: I can not imagine how to test `AfterRestartTask` (unless mocking a lot of things to insert wait points). If you have some straightforward way to do it in mind, please tell me, I'll be happy to implement it. If not, sorry, I give up.

amuniz:
@jglick 

> assuming you can confirm that all the new tests actually test what you think they are testing

No, I can not. Sorry. Fixed some minor details in tests - and added inline comment about testing `AfterRestartTask`.

Should I remove the test?

jglick:
> Should I remove the test?

Well if you did not follow TDD, then the correct practice is to go back when you think you are done and comment out/revert your fix (`src/main/`) and rerun the tests and verify that they fail in the expected ways. If they pass, or fail for other reasons, then your tests are not performing their advertised function and it would be clearer to delete them (if you cannot rework them).

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
jglick: I do not know of a straightforward way to test it. Probably it is possible via `QueueTaskDispatcher`, but too awkward to be worth it. (For the same reason, I doubt people are likely to encounter the error on `AfterRestartTask` often to begin with—i.e., much lower priority than the other fixes here.)

aggregator/src/test/java/org/jenkinsci/plugins/workflow/BuildQueueTasksTest.java
jglick: Ah, I see. Possible using `getOwnerTask` but overkill.

amuniz:
Ok. Checked that the following tests fail after reverting the changes: `queueAPI`, `queueAPIRestartable`, `computerAPI`, but not `computerAPIRestartable`, so removing it.

jglick:
:bee:, thank you for your patience.

amuniz:
@reviewbybees done

PR #299, opened 2016-01-12
Escape method name.

KostyaSha:
Confuses me with `No such DSL method step found among [...`


daniel-beck:
Maybe rephrase to *No DSL method named 'foo' found…*?

KostyaSha:
Maybe, but current message confused me, maybe it also makes sense propose install missing plugins...

jglick:
Builds are broken, trying to debug.

PR #303, opened 2016-01-14
Experimental cleanup using new plugin parent POM

andresrc:
Experimental Parent POM used: https://github.com/andresrc/plugin-pom
Won't build as cannot fetch parent POM (unreleased).

DO NOT MERGE.

@reviewbybees esp. @jglick and @stephenc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,8 +27,8 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
-        <artifactId>plugin</artifactId>
-        <version>1.609.2</version> <!-- but see Jenkins-Version override below -->
+        <artifactId>plugin-pom</artifactId>
+        <version>0.1-SNAPSHOT</version> <!-- but see Jenkins-Version override below -->
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-pom</artifactId>
@@ -49,7 +49,8 @@
       <tag>HEAD</tag>
   </scm>
     <properties>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <jenkins.version>1.609.2</jenkins.version>
+        <hpi-plugin.version>1.110</hpi-plugin.version>
     </properties>
     <modules>
         <module>step-api</module>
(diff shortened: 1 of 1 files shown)


amuniz:
LGTM!

pom.xml
```diff
@@ -49,7 +49,7 @@
       <tag>HEAD</tag>
   </scm>
     <properties>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <jenkins.version>1.609.2</jenkins.version>
```
jglick: readd comment from the `parent` section (it is important)

pom.xml
```diff
@@ -217,7 +168,6 @@
                     <plugin>
                         <groupId>org.codehaus.mojo</groupId>
                         <artifactId>findbugs-maven-plugin</artifactId>
-                        <version>3.0.2</version>
                         <configuration>
                             <xmlOutput>true</xmlOutput>
                             <failOnError>false</failOnError>
```
jglick: Could probably be deleted.

BTW with the `findbugs` profile (see a few lines up), I am setting up the POM so that when you

    mvn -DskipTests clean install

(i.e., a normal developer build) you do not pay the price of running FindBugs analysis, which can be quite slow: you do the minimum necessary to get artifacts, and perhaps later run _selected_ tests.

CI builds, or careful developer builds, do not set `skipTests` and so we both run tests _and_ static analysis. Perhaps an abuse of `skipTests` but quite convenient.

Also the `maven-javadoc-plugin` execution here is perhaps superseded by your inclusion of this in the release profile, though I recommend the parent POM be made to do it the way I do here: so that CI builds, or local builds in “careful” mode, verify the absence of Javadoc errors (especially on JDK 8). This ensures that mistakes are caught during PRs, well before an attempt is made to cut a release.

pom.xml
andresrc: I've kept the override. I assume is the thing to do as the new 1.13 version keeps it.

amuniz: BTW I don't see a clear reason for that core version discrepancy (introduced [here](https://github.com/andresrc/workflow-plugin/commit/0a5dbf7407c6fb60893954f89ef86d5a76fb9d47) without issue linked)

amuniz: Ok, it was done to [pick up JENKINS-28840](https://github.com/jenkinsci/workflow-plugin/pull/184) to avoid failing tests, but in that moment 1.609.2 was just released and forcing a core update just for testing purposes was avoided (thus #197).

I guess the override could be removed now (there is even a new LTS line out there).

andresrc:
Adapted to 2.0-SNAPSHOT and rebased. Further cleanup could be done in modules poms (TODO).

pom.xml
```diff
@@ -49,7 +49,9 @@
       <tag>HEAD</tag>
   </scm>
     <properties>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <jenkins.version>1.609.2</jenkins.version>
+        <!-- TODO: Compilation errors to investigate when using 2.0 (htmlunit related) -->
+        <jenkins-test-harness.version>1.609.2</jenkins-test-harness.version>
```
andresrc: I suppose this is because of the `HtmlUnit` changes mentioned here https://github.com/jenkinsci/copyartifact-plugin/pull/76 so I have to apply those changes.

pom.xml
jglick: @amuniz yes we could probably bump up to 1.609.3 at this point.

jglick: BTW #197 is the relevant PR if you are wondering about the override.

pom.xml
jglick: This code is already prepared, just not in `master`. We could backport most of [this](https://github.com/jenkinsci/workflow-plugin/compare/master...post-625).

PR #315, opened 2016-01-20
Document currentBuild.absoluteUrl

jtnord:
@reviewbybees


jglick:
:bug: this property is not writable, and is available for all builds, not just the current one. It is already documented AFAIK, under the `build` step.

jglick:
What would probably be useful is some better way of sharing the list of properties between the two help pages.

PR #316, opened 2016-01-21
[FIXED JENKINS-32547] Make `durable-task-step` play nicely with WinRM connections

arthurschreiber:
[JENKINS-32547](https://issues.jenkins-ci.org/browse/JENKINS-32547)

WinRM is slow. Like, really, __really__ slow. So slow, that streaming
back the results of a `DurableTaskStep` can easily take more than
the previously set timeout of 3 seconds. This would cause even
super-simple builds steps like `bat 'echo TEST'` to run for very
long times (in some cases more than 20 minutes).

Bumping the internal timeout of a `DurableTaskStep` to 10 seconds seems
to fix any issues with slaves connected over WinRM, and I've not seen
a single `InterruptedException` show up in the log files.

### DIFF
--- a/durable-task-step/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/durable-task-step/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -167,7 +167,7 @@ public abstract class DurableTaskStep extends AbstractStepImpl {
             if (workspace == null) {
                 return; // slave not yet ready, wait for another day
             }
-            // Do not allow this to take more than 3s for any given task:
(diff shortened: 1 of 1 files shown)


jglick:
AFAIK changing this timeout is harmless enough (and could probably be made a system property), but I have to ask whether it is even worth using this kind of slave if it responds so slowly?

arthurschreiber:
For now, there's no other way to use Windows Slaves in combination with the ec2-plugin.

Also, you have to keep in mind that this is only the communication over the WinRM transport which is slow, not the build execution itself. I'm not sure what specifically is the root of the slowness, it might be the way WinRM is implemented in the ec2-plugin and not something inherent to WinRM itself (but from what I've read, WinRM is a big mess so it might also be inherent).

PR #318, opened 2016-01-22
Fixed onclick handler

jglick:
Fixes a trivial but longstanding annoyance: that the console log scrolls back up to the top of the screen when you click on _Proceed_ from an `input` step (for example).

The comment about `crumb` was deleted because I confirmed that it works, due to [this hack](https://github.com/jenkinsci/jenkins/blob/1f89b5a9096f8d4946c1797e79ca5e2f494175ee/war/src/main/webapp/scripts/prototype.js#L1535-L1541).

@reviewbybees


amuniz:
über-:bee:
I was facing this continuously in a talk using the demo image yesterday.

PR #260, opened 2015-12-01
[JENKINS-31458] Taking advantage of nested object construction in StepConfigTester

jglick:
Based on #229 (just to get everything set up for Jenkins dev builds) but downstream of https://github.com/jenkinsci/jenkins/pull/1936.

[Incremental diff](https://github.com/jglick/workflow-plugin/compare/JobPropertyStep+OptionalJobProperty+BuildDiscarderProperty...DataBoundConstructor-JENKINS-31458)

@reviewbybees

PR #229, opened 2015-10-27
JobPropertyStep with OptionalJobProperty & fixed ParametersDefinitionProperty & BuildDiscarderProperty

jglick:
Downstream of #228 and https://github.com/jenkinsci/jenkins/pull/1889 (which are in turn second-level PRs).

[Incremental diff](https://github.com/jenkinsci/workflow-plugin/compare/post-625...jglick:JobPropertyStep+OptionalJobProperty+BuildDiscarderProperty)

@reviewbybees

PR #321, opened 2016-01-22
Changes for 1.642.x

jglick:
Subsumes #229 and #260 and #322.

Only “in progress” insofar as I am reluctant to start publishing releases for this line already.

@reviewbybees


amuniz:
:bee:

PR #323, opened 2016-01-22
[JENKINS-30759] Race condition initializing fields in RunningTask

jglick:
[JENKINS-30759](https://issues.jenkins-ci.org/browse/JENKINS-30759)

@reviewbybees

### DIFF
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -362,7 +362,11 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                     return;
                 }
                 final AsynchronousExecution execution = runningTask.execution;
-                assert execution != null && runningTask.launcher != null;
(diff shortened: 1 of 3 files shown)


support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -362,7 +362,11 @@ private static void finish(@CheckForNull String cookie) {
                     return;
                 }
                 final AsynchronousExecution execution = runningTask.execution;
-                assert execution != null && runningTask.launcher != null;
+                if (execution == null) {
```
amuniz: How is this possible? A call to finish on a `node` step when its execution is not even scheduled?

jglick: Its body task _has_ been scheduled. Creating the `AsynchronousExecution` happens slightly later. If the body is very fast (typically just because there are no external tasks, just a `println`), then the body can finish in between the two. Definitely a corner case.

amuniz:
:bee:

PR #317, opened 2016-01-21
New POM

jglick:
Supersedes #303.

- [X] use new parent POM
- [X] use split test harness
- [X] switch baseline to 1.609.3
- [x] fix `java.lang.IncompatibleClassChangeError: Implementing class` (cf. https://github.com/jenkinsci/jenkins-test-harness/pull/2)

@reviewbybees esp. @andresrc

PR #325, opened 2016-01-22
[JENKINS-30395] Deleting workaround

jglick:
Taking advantage of #317 to revert a workaround from #277 and #287.

@reviewbybees


amuniz:
:bee:

PR #327, opened 2016-01-25
Noting radiatorview and chuck compatibilities

batmat:


batmat:
Flaky build, seems like.

PR #334, opened 2016-02-02
[JENKINS-32727] Rerun a build with script edits

jglick:
[JENKINS-32727](https://issues.jenkins-ci.org/browse/JENKINS-32727)

- [X] working prototype
- [X] test support for script loaded from SCM incl. `Jenkinsfile`
- [x] fix permission model
- [x] pass parameters from original
- [x] support editing `loadedScripts`
- [x] use ACE editor for main script
- [x] `make -C demo run-snapshot` and verify
- [x] fix ACE editor for loaded scripts
- [x] warn on syntax errors
- [x] inline help
- [x] CLI command
- [x] display diff relative to original script (`diff4j` or `java-diff-utils`)
- [x] reconsider naming of feature
- [x] adjust CD demo to use `load`
- [x] `IllegalArgumentException: Failed to parse 'hudson.model.Run.Rerun:dev' --- no such permission`
- [x] marshal `CpsFlowExecution.sandbox`

@reviewbybees


cps/src/main/java/org/jenkinsci/plugins/workflow/cps/rerun/RerunAction.java
```diff
@@ -0,0 +1,152 @@
...
+    /* accessible to Jelly */ public boolean isEnabled() {
+        Job job = run.getParent();
+        if (!job.hasPermission(Item.BUILD)) {
+            return false;
+        }
+        if (!job.hasPermission(Item.EXTENDED_READ)) {
+            // CONFIGURE would fail on a branch project, and anyway we are not changing the job configuration per se.
+            // We do want to be able to see the script (if CpsFlowDefinition).
+            // TODO but for multibranch we arguably want to limit this to people who could have changed the MultiBranchProject or OrganizationFolder.
+            // Do we need a new permission just for this action?
```
jglick: As written, this is probably a security hole. Say someone created a job using folder-scoped credentials to do some sensitive deployment, then granted `EXTENDED_READ` on the folder to others (which lets them see the name of the credentials, but of course not the password, plus the job configuration generally). Allowing them to rerun with an edited script would allow them to hijack the credentials.

Ideally for `CpsScmFlowDefinition` and multibranch the requisite permission would match a user’s ability to actually commit to the selected SCM: in other words, to make such an edit persistent. But Jenkins has no way of checking that directly (short of using the `github-oauth` plugin or similar). We could check whether the `User` has recorded commits in the job’s changelog, though this seems easily spoofed (for example by editing your user profile to include someone else’s email address). The only foolproof and generic way to determine if someone _could_ commit to a given SCM repository (and branch, which may have its own ACL) is to challenge them to do so: display a random token and ask them to commit it to some file. This seems overly burdensome, and it is unclear how you would cache this approval in a way that allows it to be retracted if SCM permissions are changed.

Enforcing `CONFIGURE` at the `WorkflowJob` level would prevent this from being used under any circumstances on multibranch projects, so that is a nonstarter. Might suffice to check for `CONFIGURE` at this level _or above_, which would then give a pass to anyone who would anyway be able to edit the `WorkflowMultiBranchProject` or `OrganizationFolder`. On the other hand it would block non-admins from using the feature in a system which is 100% pipeline-as-code (only `READ` + `BUILD` for developers).

A new permission would work, but adds to the `AuthorizationStrategy` configuration UI, and imposes a burden on installations using this feature to grant it. Perhaps it could be `impliedBy(CONFIGURE)`, which I *think* would result in it being granted even on a branch project if the user had `CONFIGURE` at the multibranch or higher level in the `AuthorizationStrategy`. Tricky because it depends on the order of interpretation of the `WorkflowJob.getACL` override vs. `Permission.impliedBy` search in the `AuthorizationStrategy`: the blocking of `CONFIGURE` & `DELETE` on branch projects and organization folder children is done “at the last minute” (invisible to the `AuthorizationStrategy`).

@stephenc might have some advice.

tfennelly:
@jglick I created a [PR on your branch](https://github.com/jglick/workflow-plugin/pull/2) with what I hope will fix your ACE editor issues. Maybe give it a try and let me know. Hopefully it works.

tfennelly:
@jglick hmmm ... those failures might be my fault ?? ... sorry. Will have a look in a bit.

jglick:
@tfennelly no I think they are mine.

cps/src/main/js/workflow-editor.js
```diff
@@ -12,98 +9,115 @@ $('.textarea-handle', wrapper).remove();
...
+            
+            // The 'ace-editor:ace-editor-122' plugin supplies an "ACEPack" object.
+            // ACEPack understands the hardwired async nature of the ACE impl and so
+            // provides some async ACE script loading functions.
+            
+            acePack.edit(editorId, function() {
+                var ace = acePack.ace;
+                var editor = this.editor;
                 
-                editor.$blockScrolling = Infinity;
```
jglick: [Ignore whitespace](https://github.com/jenkinsci/workflow-plugin/pull/334/files?w=1#diff-63ad0d7784893a9dca6c64f8c55aa693) to review.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayActionTest.java
```diff
@@ -0,0 +1,205 @@
...
+    @ClassRule public static BuildWatcher buildWatcher = new BuildWatcher();
+    @Rule public JenkinsRule r = new JenkinsRule();
+
+    @Test public void editSimpleDefinition() throws Exception {
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("echo 'first script'", false));
+        WorkflowRun b1 = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
+        r.assertLogContains("first script", b1);
+        WorkflowRun b2;
+        /* TODO broken by ACE integration, figure out how to resurrect:
```
amuniz: Perhaps you can get the underlying `<textarea>` which has `name=_.script`

jglick: That is what I _was_ doing (line 80), but when ACE is in use it behaves very differently and the same test began failing.

jglick: Specifically with this patch

```diff
diff --git a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayActionTest.java b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayActionTest.java
index ffc8382..084ad86 100644
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayActionTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayActionTest.java
@@ -24,6 +24,9 @@
 
 package org.jenkinsci.plugins.workflow.cps.replay;
 
+import com.gargoylesoftware.htmlunit.html.HtmlForm;
+import com.gargoylesoftware.htmlunit.html.HtmlPage;
+import com.gargoylesoftware.htmlunit.html.HtmlTextArea;
 import org.jenkinsci.plugins.workflow.cps.replay.ReplayAction;
 import org.jenkinsci.plugins.workflow.cps.replay.ReplayCause;
 import hudson.FilePath;
@@ -70,7 +73,6 @@ public class ReplayActionTest {
         WorkflowRun b1 = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
         r.assertLogContains("first script", b1);
         WorkflowRun b2;
-        /* TODO broken by ACE integration, figure out how to resurrect:
         { // First time around, verify that UI elements are present and functional.
             ReplayAction a = b1.getAction(ReplayAction.class);
             assertNotNull(a);
@@ -87,8 +89,6 @@ public class ReplayActionTest {
             b2 = p.getBuildByNumber(2);
             assertNotNull(b2);
         } // Subsequently can use the faster whitebox method.
-        */
-        b2 = (WorkflowRun) b1.getAction(ReplayAction.class).run("echo 'second script'").get();
         r.assertLogContains("second script", r.assertBuildStatusSuccess(b2));
         ReplayCause cause = b2.getCause(ReplayCause.class);
         assertNotNull(cause);
```

the test just begins using 100% CPU:

```
"Executing editSimpleDefinition(org.jenkinsci.plugins.workflow.cps.replay.ReplayActionTest)" … waiting for monitor entry […]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at com.gargoylesoftware.htmlunit.javascript.JavaScriptEngine$HtmlUnitContextAction.run(JavaScriptEngine.java:847)
	- waiting to lock <…> (a com.gargoylesoftware.htmlunit.html.HtmlPage)
	at net.sourceforge.htmlunit.corejs.javascript.Context.call(Context.java:628)
	at …
	at com.gargoylesoftware.htmlunit.html.HtmlInput.executeOnChangeHandlerIfAppropriate(HtmlInput.java:511)
	at com.gargoylesoftware.htmlunit.html.HtmlTextArea.setText(HtmlTextArea.java:125)
	at org.jenkinsci.plugins.workflow.cps.replay.ReplayActionTest.editSimpleDefinition(ReplayActionTest.java:84)
"JS executor for org.jvnet.hudson.test.JenkinsRule$WebClient@…" … runnable […]
   java.lang.Thread.State: RUNNABLE
	at java.util.AbstractCollection.toArray(AbstractCollection.java:139)
	at java.util.ArrayList.<init>(ArrayList.java:177)
	at com.gargoylesoftware.htmlunit.html.HtmlPage.safeGetAttributeListeners(HtmlPage.java:2097)
	- locked <…> (a java.lang.String)
	at com.gargoylesoftware.htmlunit.html.HtmlPage.fireHtmlAttributeAdded(HtmlPage.java:2060)
	at com.gargoylesoftware.htmlunit.html.HtmlElement.setAttributeNS(HtmlElement.java:203)
	at com.gargoylesoftware.htmlunit.html.DomElement.setAttribute(DomElement.java:327)
	at com.gargoylesoftware.htmlunit.javascript.host.css.CSSStyleDeclaration.writeToElement(CSSStyleDeclaration.java:696)
	at com.gargoylesoftware.htmlunit.javascript.host.css.CSSStyleDeclaration.replaceStyleAttribute(CSSStyleDeclaration.java:614)
	at com.gargoylesoftware.htmlunit.javascript.host.css.CSSStyleDeclaration.setStyleAttribute(CSSStyleDeclaration.java:586)
	at com.gargoylesoftware.htmlunit.javascript.host.css.CSSStyleDeclaration.setStyleAttribute(CSSStyleDeclaration.java:565)
	at com.gargoylesoftware.htmlunit.javascript.host.css.CSSStyleDeclaration.setStyleAttributePixel(CSSStyleDeclaration.java:6080)
	at com.gargoylesoftware.htmlunit.javascript.host.css.CSSStyleDeclaration.setHeight(CSSStyleDeclaration.java:2043)
	at sun.reflect.GeneratedMethodAccessor70.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at net.sourceforge.htmlunit.corejs.javascript.MemberBox.invoke(MemberBox.java:153)
	at net.sourceforge.htmlunit.corejs.javascript.ScriptableObject$GetterSlot.setValue(ScriptableObject.java:311)
	at net.sourceforge.htmlunit.corejs.javascript.ScriptableObject$RelinkedSlot.setValue(ScriptableObject.java:384)
	at net.sourceforge.htmlunit.corejs.javascript.ScriptableObject.putImpl(ScriptableObject.java:2887)
	at net.sourceforge.htmlunit.corejs.javascript.ScriptableObject.put(ScriptableObject.java:547)
	at com.gargoylesoftware.htmlunit.javascript.host.css.CSSStyleDeclaration.put(CSSStyleDeclaration.java:2558)
	at net.sourceforge.htmlunit.corejs.javascript.ScriptableObject.putProperty(ScriptableObject.java:2570)
	at net.sourceforge.htmlunit.corejs.javascript.ScriptRuntime.setObjectProp(ScriptRuntime.java:1674)
	at net.sourceforge.htmlunit.corejs.javascript.ScriptRuntime.setObjectProp(ScriptRuntime.java:1669)
	at net.sourceforge.htmlunit.corejs.javascript.Interpreter.interpretLoop(Interpreter.java:1255)
	at net.sourceforge.htmlunit.corejs.javascript.Interpreter.interpret(Interpreter.java:798)
	at net.sourceforge.htmlunit.corejs.javascript.InterpretedFunction.call(InterpretedFunction.java:105)
	at net.sourceforge.htmlunit.corejs.javascript.ContextFactory.doTopCall(ContextFactory.java:411)
	at com.gargoylesoftware.htmlunit.javascript.HtmlUnitContextFactory.doTopCall(HtmlUnitContextFactory.java:310)
	at net.sourceforge.htmlunit.corejs.javascript.ScriptRuntime.doTopCall(ScriptRuntime.java:3286)
	at com.gargoylesoftware.htmlunit.javascript.JavaScriptEngine$4.doRun(JavaScriptEngine.java:789)
	at com.gargoylesoftware.htmlunit.javascript.JavaScriptEngine$HtmlUnitContextAction.run(JavaScriptEngine.java:850)
	- locked <…> (a com.gargoylesoftware.htmlunit.html.HtmlPage)
	at net.sourceforge.htmlunit.corejs.javascript.Context.call(Context.java:628)
	at net.sourceforge.htmlunit.corejs.javascript.ContextFactory.call(ContextFactory.java:513)
	at com.gargoylesoftware.htmlunit.javascript.JavaScriptEngine.callFunction(JavaScriptEngine.java:796)
	at com.gargoylesoftware.htmlunit.javascript.JavaScriptEngine.callFunction(JavaScriptEngine.java:768)
	at com.gargoylesoftware.htmlunit.InteractivePage.executeJavaScriptFunctionIfPossible(InteractivePage.java:201)
	at com.gargoylesoftware.htmlunit.javascript.background.JavaScriptFunctionJob.runJavaScript(JavaScriptFunctionJob.java:53)
	at com.gargoylesoftware.htmlunit.javascript.background.JavaScriptExecutionJob.run(JavaScriptExecutionJob.java:103)
	at com.gargoylesoftware.htmlunit.javascript.background.JavaScriptJobManagerImpl.runSingleJob(JavaScriptJobManagerImpl.java:428)
	at com.gargoylesoftware.htmlunit.javascript.background.DefaultJavaScriptExecutor.run(DefaultJavaScriptExecutor.java:168)
	at java.lang.Thread.run(Thread.java:745)
```

and finally times out.

Perhaps we need a way for the `workflow-editor` form control to switch to a plain textarea when run inside HtmlUnit.

cps/src/main/resources/org/jenkinsci/plugins/workflow/editor/workflow-editor.jelly
```diff
@@ -0,0 +1,22 @@
...
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:f="/lib/form">
+    <st:documentation>
+        <st:attribute name="script">
+            The script.
+        </st:attribute>
+    </st:documentation>
+    <j:choose>
+        <j:when test="${h.isUnitTest}">
+            <!-- HtmlUnit and ACE do not seem to mix. -->
```
tfennelly: Yeah, it does a lot of async initialization, which means it's a PITA to integrate with HtmlUnit.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayActionTest.java
```diff
@@ -0,0 +1,203 @@
...
+        r.assertLogContains("received original text", b2);
+        assertEquals(0, new CLICommandInvoker(r, "replay-pipeline").withStdin(IOUtils.toInputStream("'new text'")).invokeWithArgs("p", "-n", "1", "-s", "Script1").returnCode());
+        r.waitUntilNoActivity();
+        WorkflowRun b3 = p.getLastBuild();
+        assertEquals(3, b3.getNumber());
+        // Main script picked up from #1, not #2.
+        r.assertLogContains("got new text", b3);
+    }
+
+}
```
tfennelly: These tests are awesome !! But ... (nit) they be even more awesome if there were some comments in there explaining what's being tested. I find it hard to review a sea of code.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
```diff
@@ -305,6 +306,14 @@ public GroovyShell getShell() {
...
         return storage;
     }
+    
+    public String getScript() {
+        return script;
+    }
+
+    public Map<String,String> getLoadedScripts() {
+        return ImmutableMap.copyOf(loadedScripts);
+    }
```
tfennelly: The above are ok for external usage?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
```diff
@@ -0,0 +1,242 @@
...
+    private ReplayAction(Run run) {
+        this.run = run;
+    }
+
+    @Override public String getDisplayName() {
+        return "Replay";
+    }
+
+    @Override public String getIconFileName() {
+        return isEnabled() ? "clock.png" : null;
```
tfennelly: All of those icons suck, but maybe `redo.png` would be more appropriate for this use case?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
```diff
@@ -0,0 +1,242 @@
...
+        CpsFlowExecution execution = getExecution();
+        return execution != null ? execution.getLoadedScripts() : /* ? */Collections.<String,String>emptyMap();
+    }
+
+    /* accessible to Jelly */ public Run getOwner() {
+        return run;
+    }
+
+    @RequirePOST
+    public void doRun(StaplerRequest req, StaplerResponse rsp) throws ServletException, IOException {
```
tfennelly: `@Restricted` ?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
```diff
@@ -0,0 +1,242 @@
...
+        if (!isEnabled()) {
+            throw new AccessDeniedException("not allowed to replay"); // AccessDeniedException2 requires us to look up the specific Permission
+        }
+        JSONObject form = req.getSubmittedForm();
+        Map<String,String> otherScripts = new HashMap<String,String>();
+        for (String clazz : getOriginalLoadedScripts().keySet()) {
+            otherScripts.put(clazz, form.getString(clazz));
```
tfennelly: And what if that script is not in `form` ?

tfennelly: `JSONObject.getString` will throw an exception, right?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
```diff
@@ -0,0 +1,242 @@
...
+        ParametersAction.class,
+        SCMRevisionAction.class
+    );
+
+    /** For whitebox testing. */
+    public @CheckForNull QueueTaskFuture/*<Run>*/ run(String script) {
+        return run(script, Collections.<String,String>emptyMap());
+    }
+
+    public @CheckForNull QueueTaskFuture/*<Run>*/ run(String script, Map<String,String> otherScripts) {
```
tfennelly: Can the args to these methods be `null`?

tfennelly: And we're ok with them being exposed to external use? (not `@Restricted`)

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
```diff
@@ -0,0 +1,242 @@
...
+            actions.addAll(run.getActions(c));
+        }
+        return new ParameterizedJobMixIn() {
+            @Override protected Job asJob() {
+                return run.getParent();
+            }
+        }.scheduleBuild2(0, actions.toArray(new Action[actions.size()]));
+    }
+
+    public String getDiff() {
```
tfennelly: So this is only being used by tests, right?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayCause.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Marker that a run is a replay of an earlier one.
+ */
+public class ReplayCause extends Cause {
+
+    private final int originalNumber;
+    private transient Run<?,?> run;
+
+    ReplayCause(Run<?,?> original) {
```
tfennelly: `@Nonnull` on the arg?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayCommand.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.Map;
+import org.apache.commons.io.IOUtils;
+import org.jenkinsci.plugins.workflow.cps.replay.Messages;
+import org.kohsuke.args4j.Argument;
+import org.kohsuke.args4j.CmdLineParser;
+import org.kohsuke.args4j.Option;
+import org.kohsuke.args4j.OptionDef;
+import org.kohsuke.args4j.spi.Setter;
+
+@Extension public class ReplayCommand extends CLICommand {
```
tfennelly: Would a user really do this from the CLI? Or maybe you're doing this to support IDE integration?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayCommand.java
```diff
@@ -0,0 +1,106 @@
...
+    @Override public String getName() {
+        return "replay-pipeline";
+    }
+
+    @Override public String getShortDescription() {
+        return Messages.ReplayCommand_shortDescription();
+    }
+
+    @Override protected int run() throws Exception {
+        Run<?,?> run = number == 0 ? job.getLastBuild() : job.getBuildByNumber(number);
```
tfennelly: I see that `job` is required in the `@Argument` def. Does that "required" mean that the arg must be supplied, or does it also mean that it must be a valid job name?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayFlowFactoryAction.java
```diff
@@ -0,0 +1,103 @@
...
+        String script = mainScript;
+        mainScript = null; // minimize build.xml size
+        return new CpsFlowExecution(script, sandbox, owner);
+    }
+
+    @Override public boolean shouldSchedule(List<Action> actions) {
+        return true; // do not coalesce
+    }
+
+    @Extension public static class ReplacerImpl implements LoadStepExecution.Replacer {
```
tfennelly: `LoadStepExecution.Replacer` funkiness :smile:

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayFlowFactoryAction.java
```diff
@@ -0,0 +1,103 @@
...
+                    } else {
+                        LOGGER.log(Level.FINE, "{0} was not a replay", executable);
+                    }
+                } else {
+                    LOGGER.log(Level.FINE, "{0} was not a run at all", executable);
+                }
+            } catch (IOException x) {
+                LOGGER.log(Level.WARNING, null, x);
+            }
+            return text;
```
tfennelly: So what is the `text` arg used for? I see (later) that it's "the original text", which seems strange.

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/ReplayActionTest.java
```diff
@@ -0,0 +1,161 @@
...
+    private static boolean canReplay(WorkflowRun b, String user) {
+        final ReplayAction a = b.getAction(ReplayAction.class);
+        return ACL.impersonate(User.get(user).impersonate(), new NotReallyRoleSensitiveCallable<Boolean,RuntimeException>() {
+            @Override public Boolean call() throws RuntimeException {
+                return a.isEnabled();
+            }
+        });
+    }
+
+}
```
tfennelly: Again... awesome tests ... great detail, coverage etc, but reading it is like wading through sweet treacle :smile:

tfennelly:
:bee: insofar as I get how it works.

One observation ... from a UX perspective, it seems a bit of a pity that the user has to go to another screen to use the replay functionality (Vs being able to do it from the config page).

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowExecution.java
jglick: Harmless so far as I know.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
jglick: Good idea. There is no 32×32 but maybe it does not matter. Will check how it looks.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
jglick: Yes, `JSONException`. Conceivable in case you try to replay a _running_ build; in such a case I can just make it fall back to the original script.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
jglick: > And we're ok with them being exposed to external use?

Yes, to the extent you can access `ReplayAction` from an API, it makes sense to call `run` on it.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
jglick: No, used in `diff.jelly`.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayCommand.java
jglick: Mainly to support IDE integration, though you could use it directly from a shell if you wanted:

```bash
$ jed Jenkinsfile
$ git diff
$ java -jar /tmp/jenkins-cli.jar -s http://jenkins/ replay mypipeline < Jenkinsfile
$ git add Jenkinsfile
$ git commit -m 'Fixed bug in Jenkins build'
```

aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayActionTest.java
jglick: (resolved from my PoV)

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayCommand.java
jglick: `required` in that context means just that the argument must be supplied. If it is not a valid job name, `JobHandler` ought to be failing the command with an appropriate error message and exit code.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayFlowFactoryAction.java
jglick: In this implementation the `text` argument is unused. I left the option in the SPI for an implementation to somehow process incoming script text.

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/ReplayActionTest.java
jglick: Added some comments which should make it like wading through _warm_ treacle.

jglick:
> it seems a bit of a pity that the user has to go to another screen to use the replay functionality (Vs being able to do it from the config page)

It would be easy enough to add a link to `lastBuild/replay` from somewhere at the job level, if that seems intuitive. My intuition was that you are not replaying the _job_, you are replaying a particular _build_ that somehow went astray.

tfennelly:
re :bee:

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/ReplayActionTest.java
```diff
@@ -0,0 +1,171 @@
...
+    @Test public void permissions() throws Exception {
+        File clones = tmp.newFolder();
+        sampleRepo.init();
+        sampleRepo.write("Jenkinsfile", "");
+        sampleRepo.git("add", "Jenkinsfile");
+        sampleRepo.git("commit", "--all", "--message=init");
+        sampleRepo.git("clone", ".", new File(clones, "one").getAbsolutePath());
+        // Set up a secured instance with an organization folder.
+        // Developers have varying permissions set at the topmost (configurable) level.
+        r.jenkins.setSecurityRealm(r.createDummySecurityRealm());
```
svanoort: I like that this is included in test coverage

multibranch/src/test/java/org/jenkinsci/plugins/workflow/multibranch/ReplayActionTest.java
svanoort: Much better than treacle to me now.  I'm happy for once with the amount of explanation in the comments, and agree that the test coverage is excellent.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayActionTest.java
```diff
@@ -0,0 +1,222 @@
...
+        // Can take a look at the build.xml and see that we are not duplicating script content once edits are applied (not yet formally asserted).
+        System.out.println(new XmlFile(new File(b2.getRootDir(), "build.xml")).asString());
+        // Diff should reflect both sets of changes.
+        String diff = b2.getAction(ReplayAction.class).getDiff();
+        assertThat(diff, containsString("+echo 'trying edits'"));
+        assertThat(diff, containsString("Script1"));
+        assertThat(diff, containsString("-echo 'original loaded text'"));
+        assertThat(diff, containsString("+echo 'new loaded text'"));
+        System.out.println(diff);
+        // TODO test multiple loads
```
svanoort: I'd really like to see this covered on a basic level, with two loads.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayActionTest.java
svanoort: <3 those comments, it's enough to make me happy and I'm on the extreme end about wanting comments.

cps/pom.xml
```diff
@@ -90,6 +94,17 @@
...
             <version>1.0.1</version>
         </dependency>
+        <dependency>
+            <groupId>com.cloudbees</groupId>
+            <artifactId>diff4j</artifactId>
+            <version>1.2</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-io</groupId>
+                    <artifactId>commons-io</artifactId>
```
svanoort: Why was the exclusion added?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    @Restricted(DoNotUse.class)
+    @RequirePOST
+    public void doRun(StaplerRequest req, StaplerResponse rsp) throws ServletException, IOException {
+        if (!isEnabled()) {
+            throw new AccessDeniedException("not allowed to replay"); // AccessDeniedException2 requires us to look up the specific Permission
+        }
+        JSONObject form = req.getSubmittedForm();
+        Map<String,String> otherScripts = new HashMap<String,String>();
```
svanoort: This really needs an explaining comment.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
```diff
@@ -0,0 +1,254 @@
...
+
+    /** For whitebox testing. */
+    public @CheckForNull QueueTaskFuture/*<Run>*/ run(String script) {
+        return run(script, Collections.<String,String>emptyMap());
+    }
+
+    /**
+     * For whitebox testing.
+     * @param script main script; replacement for {@link #getOriginalScript}
+     * @param otherScripts auxiliary scripts, keyed by class name; replacement for {@link #getOriginalLoadedScripts}
```
svanoort: I feel like "otherScripts" is uninformative & confusing enough a name to merit a :bug:  here.  I'd suggest this should be "replacementScripts"

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/LoadStepExecution.java
```diff
@@ -56,4 +71,20 @@ public void stop(Throwable cause) throws Exception {
 
     private static final long serialVersionUID = 1L;
 
+    /** Allows content of the loaded script to be substituted. */
+    @Restricted(NoExternalUse.class) // for now anyway
+    public interface Replacer extends ExtensionPoint {
```
svanoort: I'm wondering why our replacer does a *single* replacement rather than taking a Map<String,String> of replacements or similar?

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/LoadStepExecution.java
```diff
@@ -27,14 +34,22 @@
...
     public boolean start() throws Exception {
         CpsStepContext cps = (CpsStepContext) getContext();
         CpsThread t = CpsThread.current();
 
-        GroovyShell shell = t.getExecution().getShell();
+        CpsFlowExecution execution = t.getExecution();
 
-        Script script = shell.parse(cwd.child(step.getPath()).readToString());
+        String text = cwd.child(step.getPath()).readToString();
+        for (Replacer replacer : ExtensionList.lookup(Replacer.class)) {
```
svanoort: Is there a reason why it wouldn't make more sense for the Replacer to do a bulk replacement of scripts?  We could still chain them if needed, I suppose.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayFlowFactoryAction.java
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Attached to a run that is a replay of an earlier one.
+ */
+class ReplayFlowFactoryAction extends InvisibleAction implements CpsFlowFactoryAction2, Queue.QueueAction {
+
+    private static final Logger LOGGER = Logger.getLogger(ReplayFlowFactoryAction.class.getName());
+
+    private String mainScript;
+    private final Map<String,String> otherScripts;
```
svanoort: The naming here is confusing enough to merit a :bug:  - really this should be "replacementScripts" or something like that.  otherScripts suggests they're adjuncts, and not replacements and I guarantee someone will get tripped up.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayFlowFactoryAction.java
svanoort: Tagging on a mention here too that I think it makes more sense for this to do batching and operate on a whole set of scripts at a time.

cps/pom.xml
jglick: Clashes with a newer version of the dependency already pulled in from Jenkins core. Maven’s dependency strategy is “closest-first” so without this exclusion you get massive linkage errors.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
jglick: Well `script` is also a replacement script, just the main script rather than one of the other scripts. So the naming is intentional here.

svanoort:
I think a lot of this is solid, but the use of 'otherScripts' is confusing enough to merit a :bug:. 

For the Replacer, I am really wondering why we're doing one script replacement at a time, rather than a batch?  

The tests are generally excellent, however, with the comments added.  They're finnicky on my system at the moment, but since CI is playing nice I'm assuming that's a local issue.

cps/pom.xml
svanoort: Works for me!

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayAction.java
jglick: Nonetheless I will try to clarify it.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayFlowFactoryAction.java
jglick: > do batching and operate on a whole set of scripts at a time

Not sure how you would expect that to happen. `load` steps can be run at any time, potentially separated by days and Jenkins restarts. The SPI simply defines what the `load` step should load if and when it runs; the rest is up to the implementation.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/LoadStepExecution.java
jglick: Because it is only called one script at a time. See my previous response.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/LoadStepExecution.java
jglick: Yes there is a reason, see above, all duplicate comments I think?

jglick:
> I am really wondering why we're doing one script replacement at a time, rather than a batch?

For the third time, see my first response. :-)

> They're finnicky on my system at the moment

Something is failing for you?

> the use of 'otherScripts'

Replaced everywhere it occurs.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayFlowFactoryAction.java
svanoort: Thanks, that helps; I think I misunderstood the intent here; but I'm still a little confused what the intent is for multiple replacements here?  Does it make sense to have multiple extension point impls, any one of which defines the final returned text?  We've marked the ExtensionPoint as NoExternalUse, is the for loop here mostly just boilerplate and something that will need to change if we allow for other extensions anyway? 

I suppose what I'm getting at here is wondering if the API allows for future extension as it stands.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/steps/LoadStepExecution.java
svanoort: Yes, duplicates to tag the different places here.

svanoort:
@jglick I like the new variable names for modified script and the replacementLoadedScripts, they make it much easier to follow - thanks!

tfennelly:
re re :bee:

svanoort:
The extended test coverage is *awesome.*  I love it.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayFlowFactoryAction.java
svanoort: @jglick Thinking about it:  does it really make sense to create this as an extension point?  Or do we have a notion about how we'd use multiple Replacer implementations without knowing the order they'll evaluate?  (I guess relying on setting ordinal for ordering?)

jglick:
Random failure I think.

svanoort:
:bee: I'm going to drop the replacer comments - (I still think it doesn't entirely make sense as an extension point yet, but since it's pretty much internal it's potentially changeable still).

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/replay/ReplayFlowFactoryAction.java
jglick: > Does it make sense to have multiple extension point impls, any one of which defines the final returned text?

As currently written, they all get a chance to adjust the result.

> is the for loop here mostly just boilerplate and something that will need to change if we allow for other extensions

No, my intent was that the loop as written should work with multiple implementations.

> wondering if the API allows for future extension as it stands

Well, that remains to be determined—if and when we find the need for some others, we will see if the API is right. In the meantime, it is restricted anyway.

PR #333, opened 2016-02-01
[JENKINS-32561] Adding NotExecutedNodeAction

tfennelly:
[See JENKINS-32561](https://issues.jenkins-ci.org/browse/JENKINS-32561)

PR #336, opened 2016-02-03
PR 333 fixups for `NotExecutedNodeAction`

tfennelly:
See #333

### DIFF
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
@@ -41,7 +41,6 @@ import hudson.XmlFile;
 import hudson.model.Action;
 import hudson.util.RobustReflectionConverter;
 import hudson.util.XStream2;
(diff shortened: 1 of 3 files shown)


tfennelly:
@jenkinsci/code-reviewers @reviewbybees

svanoort:
:bee:

amuniz:
:bee: I like it @tfennelly :smile:

support/src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
```diff
@@ -151,7 +150,6 @@ private Tag(@Nonnull FlowNode node, @Nonnull List<Action> actions) {
     private static final Field FlowNode$parentIds;
 
     static {
-        XSTREAM.addCompatibilityAlias("com.cloudbees.workflow.cps.checkpoint.NotExecutedNodeAction", NotExecutedNodeAction.class);
```
jglick: The agreement was that this alias could be registered in the plugin from whence `NotExecutedNodeAction` came, rather than here.

PR #338, opened 2016-02-07
logPrefixCache was not intended to be serialized

jglick:
Mistake in #213 which leads to junk in `build.xml` in 1.13.

@reviewbybees


amuniz:
:bee:

svanoort:
:bee:

PR #339, opened 2016-02-08
[JENKINS-32819] Fix script compilation check when CSRF is enabled

amuniz:
[JENKINS-32819](https://issues.jenkins-ci.org/browse/JENKINS-32819)

@reviewbybees

### DIFF
--- a/cps/src/main/js/workflow-editor.js
+++ b/cps/src/main/js/workflow-editor.js
@@ -51,21 +51,21 @@ jenkinsJSModules.import('ace-editor:ace-editor-122')
                     editor.session.clearAnnotations();
                     var url = textarea.attr("checkUrl") + 'Compile';
 
-                    $.ajax({
-                        url: url,
-                        data: {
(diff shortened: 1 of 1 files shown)


tfennelly:
:bee: LGTM

tfennelly:
@amuniz however, as we were saying yesterday, it would be nice to create an NPM package that contains a module to do this kind of thing without resorting to global poop (`Ajax`).

oleg-nenashev:
:bee:

amuniz:
@tfennelly 
> it would be nice to create an NPM package that contains a module to do this kind of thing without resorting to global poop (Ajax).

Yeah, agree, but let's fix this and let's do that in different PR, ok?

amuniz:
@reviewbybees done

tfennelly:
> Yeah, agree, but let's fix this and let's do that in different PR, ok?

Oh absolutely !!  Sorry, I don't mean that as a distraction here.

PR #342, opened 2016-02-12
Deleting abandoned stm plugin

jglick:
Originally the idea was to create an alternate implementation of `FlowExecution` and associated APIs, as a state transition machine. But this work was abandoned long ago and there is no point keeping it in trunk sources.

@reviewbybees


andresrc:
:bee:

PR #343, opened 2016-02-17
[JENKINS-33005] Hang in StageStepExecution.println

jglick:
[JENKINS-33005](https://issues.jenkins-ci.org/browse/JENKINS-33005)

@reviewbybees

### DIFF
--- a/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/StageStepExecution.java
+++ b/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/StageStepExecution.java
@@ -11,7 +11,6 @@ import hudson.model.TaskListener;
 import hudson.model.listeners.RunListener;
 import java.io.File;
 import java.io.IOException;
-import java.io.Serializable;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -222,10 +221,14 @@ public class StageStepExecution extends AbstractStepExecutionImpl {
     }
 
     private static void println(StepContext context, String message) {
+        if (!context.isReady()) {
+            LOGGER.log(Level.FINE, "cannot print message ‘{0}’ to dead {1}", new Object[] {message, context});
+            return;
+        }
         try {
             context.get(TaskListener.class).getLogger().println(message);
         } catch (Exception x) {
-            LOGGER.log(WARNING, null, x);
+            LOGGER.log(WARNING, "failed to print message to dead " + context, x);
         }
     }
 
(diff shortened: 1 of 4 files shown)


support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/StageStepExecution.java
```diff
@@ -222,18 +222,26 @@ private static void cleanUp(Job<?,?> job, String jobName) {
     }
 
     private static void println(StepContext context, String message) {
+        if (!context.isReady()) {
```
jglick: Really it would be preferable to send all calls to `get`, `onSuccess`, and `onFailure` to a per-context asynchronous queue, so that the `StageStepExecution.class` monitor is never held while these calls are made. Seems overly complex for now.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/StageStepExecution.java
```diff
@@ -222,18 +222,26 @@ private static void cleanUp(Job<?,?> job, String jobName) {
...
     private static void cancel(StepContext context, StepContext newer) throws IOException, InterruptedException {
-        println(context, "Canceled since " + newer.get(Run.class).getDisplayName() + " got here");
-        println(newer, "Canceling older " + context.get(Run.class).getDisplayName());
-        context.onFailure(new FlowInterruptedException(Result.NOT_BUILT, new CanceledCause(newer.get(Run.class))));
+        if (context.isReady() && newer.isReady()) {
+            println(context, "Canceled since " + newer.get(Run.class).getDisplayName() + " got here");
+            println(newer, "Canceling older " + context.get(Run.class).getDisplayName());
+            context.onFailure(new FlowInterruptedException(Result.NOT_BUILT, new CanceledCause(newer.get(Run.class))));
+        } else {
+            LOGGER.log(WARNING, "cannot cancel dead {0} or {1}", new Object[] {context, newer});
```
amuniz: So in this case, the older build will no be cancelled and the new one will proceed without any confirmation of the old one being killed. I think this could lead to weird situations if the old build comes back to live sometime... but unfortunately I don't see any _secure_ fix for this corner case.

amuniz:
:bee:

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/StageStepExecution.java
jglick: I _suspect_ that in this case the older build is never coming back. `getFlowExecution()` returned, yet its `programPromise` never got set, perhaps because there was a fatal error loading the build. If I knew how to reproduce, a better fix might be possible.

PR #284, opened 2015-12-29
[JENKINS-32214] Polling baseline with in-progress builds

jglick:
[JENKINS-32214](https://issues.jenkins-ci.org/browse/JENKINS-32214)

- [X] reproduced in test
- [ ] fix somehow

PR #344, opened 2016-02-18
[JENKINS-32214] Polling baseline with in-progress builds

jglick:
Replaces #284. [JENKINS-32214](https://issues.jenkins-ci.org/browse/JENKINS-32214)

- [X] reproduced in test
- [x] fix somehow
- [x] handle more complex scenarios

@reviewbybees

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Only noting significant user changes, not internal code cleanups and minor bug f
 
 * Simple `git` step now checks out a branch, not a detached head, for ease of committing to the workspace.
 * [JENKINS-33005](https://issues.jenkins-ci.org/browse/JENKINS-33005): hang running `stage` step which tries to cancel an earlier build that could not be loaded.
+* [JENKINS-32214](https://issues.jenkins-ci.org/browse/JENKINS-32214): polling for Subversion and Mercurial did not take into account changes already checked out in a running build.
 
 ## 1.14-beta-1 (Feb 10 2016)
 
--- a/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/scm/SubversionStepTest.java
+++ b/aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/scm/SubversionStepTest.java
@@ -122,7 +122,6 @@ public class SubversionStepTest {
     @Test public void pollDuringBuild() throws Exception {
         sampleRepo.init();
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
-        p.addTrigger(new SCMTrigger(""));
         p.setDefinition(new CpsFlowDefinition(
             "node {\n" +
             "  svn '" + sampleRepo.trunkUrl() + "'\n" +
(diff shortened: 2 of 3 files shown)


jglick:
I am not sure if this fix is really complete, especially in the face of concurrent builds with varying timing. The [logic in `AbstractProject`](https://github.com/jenkinsci/jenkins/blob/5627b7b456b24aca96f672c887bbf6c643415040/core/src/main/java/hudson/model/AbstractProject.java#L1332-L1345) is [rather complex](https://github.com/jenkinsci/jenkins/blob/5627b7b456b24aca96f672c887bbf6c643415040/core/src/main/java/hudson/model/AbstractProject.java#L1444-L1447) (and probably has poor test coverage).

Possibly we need to search for baselines in _every_ running build (terminated at the last completed, since there is no efficient way to find these without a boundary), presumably giving preference to the more recent ones.

Possibly we also need to record baselines from the polling itself, as `AbstractProject` appears to do, I suppose in case one polling put a build in the queue and then polling is run again before that build can start to go through a checkout.

jglick:
Handled both requirements, I think. There are probably still some corner cases involving builds in progress during a Jenkins restart triggering one extra round of queue additions. `AbstractProject` does not seem to fully handle this either, though it is possible that the lack of durability makes it irrelevant in that case. Could perhaps be solved by persisting `$JENKINS_HOME/jobs/whatever/pollingBaselines.xml` if it turns out to be an issue.

svanoort:
:bee:  after some deep diving to understand what this is actually doing.   It's powerful, and seems to cover all the cases I can come up with.

It feels like the test could use comments to explain exactly what's happening there (since it's not intuitive what's happening there unless you're quite familiar with how the polling statuses work) but that's an :ant: at worst.

amuniz:
Since the implementation is using a reference to the `lastBuild` workspace, I think the polling for SVN will not work when the last build was run in an `EphemeraNode` (IIUC). Anyway that case was not working before neither, so :bee:

jglick:
> I think the polling for SVN will not work when the last build was run in an `Ephemera[l]Node` (IIUC)

Well first of all there is existing handling for those SCMs which `requiresWorkspaceForPolling`, however discouraged it may be. It _does_ require that the node still exist, so we can grab a workspace. This logic could probably be made more robust following the [tricks done in `AbstractProject.getSomeBuildWithExistingWorkspace`](https://github.com/jenkinsci/jenkins/blob/6597ba8a2454788394433f18c1a10d6d27b91a64/core/src/main/java/hudson/model/AbstractProject.java#L579-L586). At any rate, this behavior should not be affected by this PR so far as I can see; a future PR could add tests for workspace-based polling and improve its coverage.

Anyway the Subversion plugin [does not require a workspace for polling](https://github.com/jenkinsci/subversion-plugin/blob/57460ce3cfb0885fce878c47eafec45e83e2fc00/src/main/java/hudson/scm/SubversionSCM.java#L814-L820) so it should always go through the simpler logic that deals only with revision states, regardless of what happened to the agent afterwards.

jglick:
> It feels like the test could use comments to explain exactly what's happening there

True. There also just need to be a lot more tests for this stuff, since there are plenty of boundary cases that I do not pretend to be covering yet, and I am not entirely sure that the user would _expect_ to happen in each case. “Deep thought” should at some point result in some kind of clear invariant that can point the way to specific test cases: for example

> Regardless of concurrency level, a job configured to build a single branch and using master-based polling should show each commit made to that branch in exactly one build’s changelog, and no build should have an empty changelog except under such-and-such conditions when it is unavoidable, and when the system is quiescent there must be a build covering the most recent commit (branch head).

I doubt these tests exist in any comprehensive form for freestyle projects, whose behavior seems to be an accretion of years of patches tuning various special cases. I can find a couple of things in `ProjectTest` and `AbstractProjectTest`.

PR #347, opened 2016-02-25
Simple form of sleep step would fail in sandbox

jglick:
`CpsClosure2.sleep` was already `@Whitelisted`, but this one (used _outside_ blocks/closures) was forgotten.

@reviewbybees


amuniz:
:bee:

off-topic: Why do we have two `sleep` impls? What's different between them? and which one is being called when using the step `sleep` in a Pipeline script?

jglick:
There is a `SleepStep`, with function name `sleep`. The weirdness comes from the fact that `DefaultGroovyMethods` also defines a method of the same name that calls `Thread.sleep`, which you assuredly do _not_ want to do in the CPS VM thread. So we try to overload `sleep` methods wherever they might occur, to delegate to the step: in `CpsScript`, for calls from top level, and in `CpsClosure2`, for calls from inside closures (such as `node {}` blocks).

PR #349, opened 2016-02-27
propose forward2docker as an alternative to `boot2docker ip`

bsideup:


KostyaSha:
Well, boot2docker cli is deprecated at all. Better use `docker-machine` instead.

bsideup:
@KostyaSha btw, forward2docker supports docker-machine as well

KostyaSha:
Yeah, but in general doc shouldn't suggest deprecated and ancient tools :)

PR #351, opened 2016-03-03
[JENKINS-32925] Add tracking to prevent stack overflows

kwhetstone:
[JENKINS-32925](https://issues.jenkins-ci.org/browse/JENKINS-32925)

To prevent stack overflows, especially in the Multiple SCMs case, track the subtypes to make sure the type isn't trying to discover itself.  When that case is detected, don't log, and return back up the stack.  This will only prevent duplication vertically, not horizontally.

@reviewbybees  esp @jglick & @abayer

### DIFF
--- a/step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
+++ b/step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
@@ -301,6 +301,9 @@ public class DescribableHelper {
 
         static ParameterType of(Type type, Stack<String> tracker) {
             try {
+                if(tracker == null){ //in case called externally
+                    tracker = new Stack<String>();
+                }
                 if (type instanceof Class) {
                     Class<?> c = (Class<?>) type;
                     if (c == String.class || Primitives.unwrap(c).isPrimitive()) {
--- a/step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
+++ b/step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
@@ -337,20 +337,16 @@ public class DescribableHelper {
                                 subtypesBySimpleName.put(simpleName, bySimpleName = new ArrayList<Class<?>>());
                             }
                             bySimpleName.add(subtype);
-                            LOG.log(Level.INFO, "Value of subtype: " + simpleName);
                         }
                         Map<String,Schema> types = new TreeMap<String,Schema>();
                         for (Map.Entry<String,List<Class<?>>> entry : subtypesBySimpleName.entrySet()) {
                             if (entry.getValue().size() == 1) { // normal case: unambiguous via simple name
                                 try {
                                     String key = entry.getKey();
-                                    LOG.log(Level.INFO, "entry value size=1 key: "+key+" stack size: "+tracker.size());
                                     if(tracker.search(key) < 0) {
                                         tracker.push(key);
                                         types.put(key, schemaFor(entry.getValue().get(0), tracker));
                                         tracker.pop();
-                                    } else {
-                                        LOG.log(Level.INFO, "I safelly triggered the stack!");
                                     }
                                 } catch (Exception x) {
                                     LOG.log(Level.FINE, "skipping subtype", x);
@@ -359,14 +355,11 @@ public class DescribableHelper {
                                 for (Class<?> subtype : entry.getValue()) {
                                     try {
                                         String name = subtype.getName();
-                                        LOG.log(Level.INFO, "entry value size!=1 name: "+name+" stack size: "+tracker.size());
                                         if(tracker.search(name) < 0) {
                                             tracker.push(name);
                                             types.put(name, schemaFor(subtype, tracker));
                                             tracker.pop();
-                                        } else {
-                                        LOG.log(Level.INFO, "I safelly triggered the stack!");
-                                    }
+                                        }
                                     } catch (Exception x) {
                                         LOG.log(Level.FINE, "skipping subtype", x);
                                     }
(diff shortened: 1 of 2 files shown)


abayer:
Oooo, nicely done. :bee:

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -147,6 +148,10 @@
     public static Schema schemaFor(Class<?> clazz) {
         return new Schema(clazz);
     }
+    
+    public static Schema schemaFor(Class<?> clazz, Stack<String> tracker) {
```
jglick: :bug: should be `private`

jglick:
Good enough for the short term: addresses the stack overflow by simply omitting a legitimate choice. Later, if needed, we can introduce a new parameter type like `Error` which explicitly indicates recursion (or even a horizontal backreference), but that would require changes to the content generators too.

jglick:
BTW there is an open issue for this on issues.jenkins-ci.org, assigned to you IIRC; link to it please.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -158,20 +163,27 @@ public static Schema schemaFor(Class<?> clazz) {
...
 
         Schema(Class<?> clazz) {
+            this(clazz, null);
+        }
+
+        Schema(Class<?> clazz, Stack<String> tracker) {
             this.type = clazz;
+            if(tracker == null){
+                tracker = new Stack<String>();
+            }
```
jglick: Or more simply, pass `new Stack<String>()` as the argument in line 166.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -282,9 +294,16 @@ public Type getActualType() {
...
         }
+        
+        static ParameterType of(Type type){
+            return of(type, new Stack<String>());
+        }
 
-        static ParameterType of(Type type) {
+        static ParameterType of(Type type, Stack<String> tracker) {
             try {
+                if(tracker == null){ //in case called externally
```
jglick: Why not just make this overload `private`? Then the `tracker` can be `@Nonnull`.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -323,14 +342,24 @@ static ParameterType of(Type type) {
                         for (Map.Entry<String,List<Class<?>>> entry : subtypesBySimpleName.entrySet()) {
                             if (entry.getValue().size() == 1) { // normal case: unambiguous via simple name
                                 try {
-                                    types.put(entry.getKey(), schemaFor(entry.getValue().get(0)));
+                                    String key = entry.getKey();
```
jglick: I guess there is a bit of confusion here about what the values in the stack represent—in some cases `simpleName`s, in others FQNs. Probably it does not really matter: at worst you could recurse once and add both to the stack before stopping.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -323,14 +342,24 @@ static ParameterType of(Type type) {
...
+                                    if(tracker.search(key) < 0) {
```
jglick: Arguably clearer to use `contains`.

kwhetstone: I'm not sure why I didn't use that in the first place... updating.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
kwhetstone: Added the `@Nonnull` annotation to the Stack in the constructor.  I really like that annotation.

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
kwhetstone: I figured that it didn't matter what specific value ended up in the stack, just that the same value would show up again indicating a possible stack overflow.  Luckily they were both strings, so it made comparisons a bit easier.

kwhetstone:
Thanks for the reviews.  I wanted to make a fairly minimal change so there wouldn't be that huge disruption to all the content generators.  I definitely will keep introducing the error type in the back of my mind if this ends up being needed down the road.

jglick:
:bee:

kohsuke:
While studying this fix, I noticed that this change missed the use of `of` in ArrayType instantiation [here](https://github.com/jenkinsci/workflow-plugin/pull/351/files#diff-5f889e800bbe1383fa455077e89bb626R370) so the stack overflow continues with multiple-scm plugin.

The recursive portion of the stack is as follows (as of c85a422f2fdfa67e3e228623b5cf97b211db1f8b):
```
	at org.jenkinsci.plugins.workflow.structs.DescribableHelper$Schema.<init>(DescribableHelper.java:179)
	at org.jenkinsci.plugins.workflow.structs.DescribableHelper.schemaFor(DescribableHelper.java:152)
	at org.jenkinsci.plugins.workflow.structs.DescribableHelper.access$400(DescribableHelper.java:77)
	at org.jenkinsci.plugins.workflow.structs.DescribableHelper$ParameterType.of(DescribableHelper.java:344)
	at org.jenkinsci.plugins.workflow.structs.DescribableHelper$ParameterType.of(DescribableHelper.java:298)
	at org.jenkinsci.plugins.workflow.structs.DescribableHelper$ParameterType.of(DescribableHelper.java:369)
	at org.jenkinsci.plugins.workflow.structs.DescribableHelper$ParameterType.access$200(DescribableHelper.java:285)
	at org.jenkinsci.plugins.workflow.structs.DescribableHelper$Schema.<init>(DescribableHelper.java:179)
```

jglick:
Probably better to leave a comment in https://github.com/jenkinsci/structs-plugin/issues/1 requesting a test case be added that has something like `Impl3` but whose polymorphic attribute is of type `Base[]`.

PR #355, opened 2016-03-04
[JENKINS-27039] New step: milestone

amuniz:
Based on discussion on [JENKINS-27039](https://issues.jenkins-ci.org/browse/JENKINS-27039) (and also covers [JENKINS-32829](https://issues.jenkins-ci.org/browse/JENKINS-32829)).

The implementation is based on the `stage` step code and @jglick's proposal in the linked JIRA. There is currently some duplicated code which will be there until `concurrency` is deprecated in `stage`.

This step has been designed to accomplish the following:
 1. Builds pass through the `milestone` in order (taking the build number as sorter field).
 1. Older builds will not proceed (they are aborted) if a newer one already entered the milestone.
 1. Once a build enters the milestone, it will be never aborted by a newer build which did reach the milestone yet.
 1. When a build passes a milestone, any older build that passed the previous milestone but not this one is aborted.

I used this Pipeline script for manual testing (thanks @stephenc):

```
milestone
node {
    def delay = new Random().nextInt(60)
    echo "Sleeping for ${delay}s..."
    sleep time: delay, unit: 'SECONDS'
}
milestone
node {
    def delay = new Random().nextInt(60)
    echo "Sleeping for ${delay}s..."
    sleep time: delay, unit: 'SECONDS'
}
```

- [x] Base implementation of `milestone` step
- [x] Automated tests
- [x] Clean up non existing `milestone`s in config file (`$JENKINS_HOME/org.jenkinsci.plugins.workflow.support.steps.MilestoneStep.xml`)
- (Delaying as SNAPSHOTs can not be used in demo `plugins.txt` file) Update the CD stock demo to use `milestone` and `lock` in tandem (it requires upstream https://github.com/jenkinsci/lockable-resources-plugin/pull/25 to be merged and released)

@reviewbybees


mbidewell:
Would it be possible to set the result of the build when it is aborted at a milestone?  I am thinking of a pipeline like the following:
Build -> Test ->Approve to Production

Even if a build's tests succeed would it be possible for it not to be deployed to production.  In this case it would be nice for the build status to reflect the test status.

amuniz:
@mbidewell 
> Even if a build's tests succeed would it be possible for it not to be deployed to production. In this case it would be nice for the build status to reflect the test status.

If the build is cancelled, the build status is set to `Result.NOT_BUILT` which is the correct status since the build didn't finished. I think you can achieve what you want by splitting a monolithic job into `test` and `deploy` jobs (sharing artifacts between the first and the second), so the status of the first one would be `Result.SUCCESS` and the second `Result.NOT_BUILT` when cancelled.

rsandell:
> I think you can achieve what you want by splitting a monolithic job into test and deploy jobs 

Isn't that what the pipeline job feature is trying desperately to avoid you having to do?

jglick:
Do we really need `concurrency` to be an `int`? In practice I never heard of any use case for a finite concurrency >1. Would suggest simplifying this to a `boolean` like `singleBuild`.

jglick:
> I think you can achieve what you want by splitting a monolithic job

We do not want to force users to do that.

`NOT_BUILT` is technically correct for an overall build result: the pipeline did not run to completion. You can still look at test results if you are interested. For purposes of visualization, we would like to be able to display results of parts of the build, which is what [JENKINS-26522](https://issues.jenkins-ci.org/browse/JENKINS-26522) is designed to cover.

jglick:
> Would suggest simplifying this to a `boolean` like `singleBuild`.

For that matter, perhaps we can simplify further by removing this option altogether, and implementing [JENKINS-30269](https://issues.jenkins-ci.org/browse/JENKINS-30269) to provide a general-purpose lock system independent of milestones, for those pipelines which need exclusive access to singleton resources such as staging servers.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStepExecution.java
```diff
@@ -0,0 +1,356 @@
...
+            waitingContext = null;
+            waitingBuild = null;
+            waitingBuildExternalizableId = null;
+        }
+    }
+
+    /**
+     * Records that a flow was canceled while waiting in a stage step because a newer flow entered that stage instead.
+     */
+    public static final class CanceledCause extends CauseOfInterruption {
```
rsandell: Consider having a more descriptive name, since some semi badly written statistics calculators are using the simpleName of the class as category display name.

support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep/help.html
```diff
@@ -0,0 +1,12 @@
+<div>
+    <p>By default, Pipeline builds can run concurrently.</p>
+    <p>
+        The milestone step forces all builds to go through in order, so an older build will never be allowed to
+        go into the milestone (it is aborted) if a newer one already passed it.
+    </p>
+    <p>
+        It also lets you mark certain sections of a build as being constrained by limited concurrency. Newer builds are 
+        always given priority when entering such a throttled milestone; older waiting builds will simply exit early if 
+        they are preëmpted.
```
rsandell: `ë`?

support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep/config.jelly
```diff
@@ -0,0 +1,34 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-->
+
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <f:entry field="ordinal" title="Milestone ordinal">
```
rsandell: What does ordinal mean? There is no help text for it.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStepExecution.java
```diff
@@ -0,0 +1,356 @@
...
+public class MilestoneStepExecution extends AbstractStepExecutionImpl {
+
+    private static final Logger LOGGER = Logger.getLogger(MilestoneStepExecution.class.getName());
+
+    @Inject(optional=true) private transient MilestoneStep step;
+    @StepContextParameter private transient Run<?,?> run;
+    @StepContextParameter private transient FlowNode node;
+    @StepContextParameter private transient TaskListener listener;
+
+    private static Map<String, Map<Integer, Milestone>> milestonesByOrdinalByJob;
```
rsandell: Static context has a tendency to behave iffy in Jenkins. Have you considered putting this as a member of the Descriptor who's life cycle is controlled by Jenkins?

jglick:
> provide a general-purpose lock system independent of milestones

For example

```groovy
milestone 1
// build and test
lock('staging') {
  milestone 2
  // deploy to staging
}
lock('production') {
  milestone 3
  // deploy to production
}
```

where if some older builds were queued up for the `staging` lock, once the original build released this lock the `lock` step would preferentially grant it to the newest waiter, at which point that new build would pass milestone 2 and so the intermediate builds could never make it. Ideally they would be immediately aborted rather than sticking around waiting to get the lock only to be aborted as soon as they got it.

jglick:
It seems that the implementation has been copied pretty much unmodified from `StageStepExecution`, which is not correct. In particular I see nothing which compares the `ordinal`s, which was the fundamental problem with `stage`—that it had no inherent order to stages and thus was not able to properly handle builds that moved through stages at varying speeds. You should also think in terms of _passing_ a milestone, not _entering_ a milestone: unlike stages, you are not “in” milestone 3; you have “passed” milestones 1, 2, and 3, but not yet 4. You should probably also be allowed to simply skip irrelevant milestones. My expectation would be that if the first build B1 has passed the tenth milestone M10, B2 has passed M3, and B3 has also passed M3, then if B3 runs `milestone 5` (skipping 4), B2 should be either immediately aborted or aborted once it tries to run `milestone 4`. In other words the data structure should be organized differently: you want to track a list of builds, by (strictly) decreasing build number, which have passed milestones in (non-strictly) increasing order:

* B3 past M3
* B2 past M3
* B1 past M10

→

* B3 past M5
* (B2 already aborted, or scheduled to be)
* B1 past M10

And of course it should be an error to attempt to enter a milestone earlier than one you have already said you passed.

amuniz:
> It seems that the implementation has been copied pretty much unmodified from StageStepExecution, which is not correct

Yes, because I wanted to migrate (as much as possible) the `stage` behavior to `milestone` so we can deprecated all `concurrent` features on stage. BTW some slightly changes were done which drastically change how the whole thing works. If we finally decide to remove the `concurrency` behavior in favor of a new `lock` step, then all that code will be simply removed.

> My expectation would be that if the first build B1 has passed the tenth milestone M10, B2 has passed M3, and B3 has also passed M3, then if B3 runs milestone 5 (skipping 4), B2 should be either immediately aborted or aborted once it tries to run milestone 4

This is what actually happens. B2 would be aborted once it reaches M4, as M4 records the last passed build (which was B3 in the example), so B2 < B3 leads to B2 being cancelled. Or perhaps I'm misunderstanding what you mean by _skipping_ milestone: that's not possible in _my_ model, I mean, if a B1 passed M10, then it means that it passed all previous milestones (there is no way to reach M10 without passing through M1, M2, ..., M9.

> Would suggest simplifying this to a boolean like singleBuild.

It's ok for me.

jglick:
> perhaps I'm misunderstanding what you mean by _skipping milestone_: that's not possible in _my_ model

Which is OK, so long as the step enforces that you are passing every milestone in order.

amuniz:
> so long as the step enforces that you are passing every milestone in order.

Yes, [it is](https://github.com/jenkinsci/workflow-plugin/pull/355/files#diff-b49845f1c0f3bfa0c52b992c2dfde425R118). Currently writing a test that proves it.

jglick:
And what about

* B2 past M2
* B1 past M2

and B2 exits successfully. B1 must now be aborted. IOW there should be an implicit “milestone at infinity” at the end of a successful build, or you should document that users are expected to conclude a script with a final `milestone` step.

jglick:
> so we can deprecate all concurrent features on `stage`

Well the object is not to deprecate an existing feature with an exact replacement, if that feature was already misguided. The object is to provide a set of features that do what we need, and `demo/repo/Jenkinsfile` should be adjusted to show it.

jglick:
>> so long as the step enforces that you are passing every milestone in order.
> Yes, it is. Currently writing a test that proves it.

Does it? I see no indication that

```groovy
milestone 1
milestone 3
```

would be rejected by the current code.

jglick:
I do not even think I see how

```groovy
milestone 2
milestone 1
```

would be rejected.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStepExecution.java
jglick: Can we fix those badly written statistics calculators, wherever they are?

support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep/config.jelly
```diff
@@ -0,0 +1,34 @@
...
+        <f:textbox/>
```
jglick:
```xml
<f:number clazz="positive-number"/>
```

support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep/help.html
jglick: I am [archaic](https://en.wikipedia.org/wiki/Diaeresis_(diacritic)#English).

jglick:
> Once a build enters the milestone, it will be never aborted.

So how does the stock continuous delivery demo look with this step? The originally reported problem, which #177 purports to fix, was precisely that we needed the `input` step to be aborted when a superseding build arrived.

amuniz:
@jglick 

> Does it? I see no indication that

As I suspected I was misunderstanding what you meant with _skipped_ milestones. I understand now. I was supposing that the user would always add milestones in the right order. Let me fix it.

> The originally reported problem, which #177 purports to fix, was precisely that we needed the input step to be aborted when a superseding build arrived.

See the last commit. I implemented the linear approach you suggested in a JIRA comment: when a build reaches milestone X (or finishes), then any older build still in milestone X - 1 is cancelled. So if you approve a build from `input`, any older build waiting on that `input` is cancelled.

#177 (and `eager` proposal) is wrong, because it could lead to _never deploy to production_ if the pipeline is loaded (which means a new-build rate lower than the average build time - uncommon, but possible). 

The linear approach is better, since the user can select which build should continue (from all waiting on `input` and any older build waiting is cancelled), but the *newer ones will be kept*, which is good as they can be approved later (or cancelled). If the newer one is approved, then any other waiting build is cancelled.

jglick:
> The linear approach is better, since …

From a casual reading it sounds right. Will need some careful testing to make sure it really works.

Any further thought on whether a fix of JENKINS-30269 would subsume the interesting use cases for `concurrency: 1`?

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/MilestoneStepTest.java
```diff
@@ -0,0 +1,106 @@
...
+import org.junit.Test;
+import org.junit.runners.model.Statement;
+import org.jvnet.hudson.test.RestartableJenkinsRule;
+
+import hudson.model.Result;
+
+public class MilestoneStepTest {
+
+    @Rule
+    public RestartableJenkinsRule story = new RestartableJenkinsRule();
```
jglick: Recommend adding `BuildWatcher`—makes it much easier to develop & debug this kind of test.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStepExecution.java
amuniz: What descriptor? `AbstractStepExecutionImpl` is not `Describable`.

amuniz:
@jglick 

> Any further thought on whether a fix of JENKINS-30269 would subsume the interesting use cases for concurrency: 1?

Yes, I think having lockable resources in the build scope would cover any use case of `concurrency: 1`, and I'd be more than happy to remove such logic from `milestone`.In your example above, the current implementation of `milestone` would cancel any build waiting for the `lock` when a newer build reaches the next milestone (usually the last step inside the `lock` block), which seems OK to me.

In the other hand, adapting `lockable-resources-plugin` to be usable in Pipeline builds does not seem trivial, it would require some work to make the lock state non-transient, thus it would block this PR for an undefined amount of time.

But I think there is a third option: simply remove `concurrency` from milestone. As `milestone ordinal: 1, concurrency: 1` is equivalent to `milestone 1; stage concurrency: 1` it would be a valid _idiom_ to workaround JENKINS-27039 and JENKINS-32829. So `milestone` could be released and JENKINS-30269 could be worked separately without blocking anything. Once it is fixed the idiom could be replaced by `milestone 1; lock(resource) { [...steps...] milestone 2 }`

For example:

```
milestone 1
stage name: 'deploy', concurrency: 1
input 'Continue?'
milestone 2
sh '... deploy ...'
```

Builds will reach `stage` in order (because of `milestone 1`) so it covers JENKINS-32829 and once one build is approved on `input` any older build waiting in `stage` is aborted (because of `milestone 2`) so it covers JENKINS-27039.

jglick:
> As `milestone ordinal: 1, concurrency: 1` is equivalent to `milestone 1; stage concurrency: 1` it would be a valid idiom to workaround JENKINS-27039 and JENKINS-32829.

This does not feel like a sound approach. We need fully tested and integrated features, not workarounds, and I would like to be able to fully deprecate `stage`’s `concurrency` (or even the whole step, in favor of a proposed `label`). If a `lock`-like step is needed to implement critical use cases, then we need to write that, test it in conjunction with `milestone`, and produce documented usage examples. In the course of integration testing you may realize that the behavior is not what you thought, and releasing a fundamentally flawed implementation of `milestone` prematurely is far worse than a slight delay.

> adapting `lockable-resources-plugin` to be usable in Pipeline builds does not seem trivial, it would require some work to make the lock state non-transient

I did not claim it would be trivial, but I believe there is no serious obstacle to doing it. `StageStepExecution` already implements a _more complicated_ concurrency primitive, with persistent state. (Admittedly it does not currently implement `stop`.) Or if interoperation with existing (freestyle) projects is not considered important, it may be sufficient to write a fresh locking step independently of that plugin, initially only supporting simple per-job mutual exclusion locks; extending that to support globally-defined locks sounds straightforward, and if there is demand it could be extended to support other lock semantics (e.g., finite locker bounds greater than one).

As an aside, removing all blocking behavior from `MilestoneStepExecution`—so that the step only ever proceeds, or aborts, or interrupts another build—means that it can become a synchronous step (returning `true` or throwing an exception from `start`).

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStepExecution.java
jglick: There is an associated `StepDescriptor`. Anyway it does not need to be a `Descriptor`; any `Extension` will do. `StageStepExecution` is implemented incorrectly in this respect, as [already noted](https://github.com/jenkinsci/workflow-plugin/blob/5b2dd6f5a070d1543511d1d2c64af3311d2b5738/support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/StageStepExecution.java#L80).

amuniz:
> Or if interoperation with existing (freestyle) projects is not considered important, it may be sufficient to write a fresh locking step independently of that plugin

It could be important IMO. Specially for users migrating to Pipeline, as they probably have a mix of Freestyle and Pipeline jobs. Using Freestyle jobs for concrete tasks because plugin X is not still compatible with Pipeline is another _common_ use case nowadays. So I think we should add a Pipeline step to `lockable-resources-plugin` (and probably improve its documentation, which is none).

jglick:
> Specially for users migrating to Pipeline, as they probably have a mix of Freestyle and Pipeline jobs.

Note that the question is whether they have a mixture of freestyle and Pipeline jobs accessing the _same resource_. Less likely, but certainly plausible.

jglick:
Note that there are few users of the plugin currently (whereas there are already a significant number of votes and watchers for JENKINS-30269). Also beware that PRs in this plugin seem to be ignored currently; it may be abandoned.

michaelneale:
Naming suggestion: milestone -> gate - is this closer to what it is?

amuniz:
@michaelneale 

> Naming suggestion: milestone -> gate - is this closer to what it is?

Once JENKINS-30269 is fixed and `concurrency` options deleted from this PR, a milestone is something that a build _passes_ not _enters_, so `gate` would not be accurate IMO.

Anyway, switching to another name is trivial, so if others feel like you, I can change it.

jglick:
> a milestone is something that a build passes not enters, so gate would not be accurate

Right—as currently conceived, this step would not block, whereas “gate” suggests a barrier that you must wait to pass through.

michaelneale:
@jglick gotcha, I was conflating it with lock which I saw elsewhere

amuniz:
`concurency` parameter has been removed as the feature is available through the `lock` step (see [PR](https://github.com/jenkinsci/lockable-resources-plugin/pull/25/) in `lockable-resources-plugin`).

amuniz:
PR builder seems to be failing due to javadoc errors in code unrelated to this PR...

jglick:
@amuniz it is related:

```
support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep.java:16: error: reference not found
* When {@link #concurrency} is set the number of builds inside the milestone will be limited to that value and
```

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep.java
```diff
@@ -0,0 +1,46 @@
...
+import hudson.Extension;
+
+/**
+ * This step can be used to grant:
+ * <ol>
+ *   <li>Builds pass through the step in order (taking the build number as sorter field)</li>
+ *   <li>Older builds will not proceed (they are aborted) if a newer one already entered the milestone</li>
+ * </ol>
+ *
+ * When {@link #concurrency} is set the number of builds inside the milestone will be limited to that value and
```
jglick: :bug: Javadoc error

support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep/config.jelly
```diff
@@ -0,0 +1,34 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-->
+
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <f:entry field="ordinal" title="Milestone ordinal">
+        <f:number clazz="positive-number"/>
+    </f:entry>
+    <f:entry field="concurrency" title="Concurrency">
```
jglick: :bug: no such field

support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep/help-concurrency.html
```diff
@@ -0,0 +1,8 @@
+<p>
+    Concurrency level.
```
jglick: :bug: no such field

support/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep/help.html
```diff
@@ -0,0 +1,12 @@
+<div>
+    <p>By default, Pipeline builds can run concurrently.</p>
+    <p>
+        The milestone step forces all builds to go through in order, so an older build will never be allowed to
+        go into the milestone (it is aborted) if a newer one already passed it.
+    </p>
+    <p>
+        It also lets you mark certain sections of a build as being constrained by limited concurrency. Newer builds are
```
jglick: No such feature any more IIUC.

HRMPW:
I'm not a big fan of the term 'milestone' because the purpose of this function is not to show that a build has passed an actual milestone but as a way to control the build order of successive jobs. Nothing in the word 'milestone' is intuitive regarding its function. A milestone to me should have a name and should be reported because a job should work towards milestones. 

I would much prefer '[juncture](http://www.thefreedictionary.com/juncture)' because it defines the actual use case.  "A point in time, especially one requiring a decision to be made"  As I understand it this is exactly the purpose of this step.

Regarding the numbers, as a user my reaction would be, "Isn't the pipeline execution order based on the order in which they appear in the script?" "When would they not execute in the order of the script?"

In addition to numbering the execution order in my script I am also forced to know the execution order and 'milestone/juncture' numbers for all steps defined in external scripts if they are used at all.  Forcing a user to know how many junctures are in other scripts that might be used as well as limiting the portability of the script makes the use of sequential numbering seem like a really bad user experience.

jglick:
> A milestone to me should have a name

Uh, actual milestones have, well, numbers—of miles. :-)

We could probably drop the number and treat it as implicit by execution order.

amuniz:
> Regarding the numbers, as a user my reaction would be, "Isn't the pipeline execution order based on the order in which they appear in the script?" "When would they not execute in the order of the script?"

Initially (when I started writing this code) the idea was to allow non-consecutive ordinals, that's why that parameter exists. So yes, I think it can be implicitly calculated now that we force consecutive ordinals. Let me try it.

abayer:
fwiw, my concern is that a milestone feels like something you pass but not something that keeps others from passing too. "milestone", "stage" and the like all feel like the same general concept, but not a control flow sort of concept. I jokingly want to call this step "thunderdome" - "two builds enter, one build leaves!" but, well, yeah.

jtnord:
I've been trying to follow the comments and the associated JIRA but am confused.

there is no linking in a miltestone - so how do you know that one is later than another.
A flow can have conditional logic such that the 3rd milestone entered in one case is a completely different milestone from the another case.

If I have 2 staging environments how would I prevent old builds from sitting forever waiting for them?

Seem like the milestone won't help me as that only supports one thing.

Locks would not help me as builds will just queue up waiting for the locks and the order of locking may be out of order....

jglick:
> A flow can have conditional logic such that the 3rd milestone entered in one case is a completely different milestone from the another case.

Not sure I follow your use case here. The expectation is that all builds have comparable milestones. You might of course have conditional logic _between_ milestones, but should not have it _around_ milestones.

jtnord:
@jglick well that's a major flaw.  I have certainly done that in the past.

The other thing to point out is that the main flow may not even control it's milestones.  For instance say it includes some other "sub flow" logic.

Say these flows run for days.
What happens if that sub flow logic is changed and a new milestone is added to it.  For 2 days all your flows are going to be broken with potentially disastrous consequences.

Or hey - just add a few new milestone to an existing flow early on (say you split your one testing milestone into "smoke testing" and "bigger testing")

so pre change you have 
```
milestone < testing
milestone < deployment
```
after you have
milestone < smoke
milestone < integration tests
milestone < deployment
```
so Build 100 which is the old flow spends 1 day in testing... and then finishes
but build 101 with the new flow takes 2 hours in smoke, then finishes and enters integration tests...
build 100 will then come to deployment and well - it will be killed...

And then you can look at what happens if milestones are removed....

If you have no ability to label your milestones you are effectivly forced to know all of your future milestones upfront...

**Comment by jglick** [comment]
So are you arguing that we should go back to my original proposal, which was that `milestone` take a mandatory nonnegative integer parameter, which must be monotonically increasing but need not be sequential?

**Comment by jtnord** [comment]
@jglick well even that has issues - in that you may not be in control of the sub flow (re-susable 3rd party (ie from a different team) scripts/sub flows).

Not sure what the solution is - even using montotonically increasing numbers would have some issues.
It's almost like you need multiple scopes - i.e. this is the "deploy" scope and it is the nth milestone (where n comes from a monotonically increasing number - and the good practice is to start at 100 and increase by 100 so you have room to fill in...) - the scope can (should?) come from a different command to the bit that controls n - such that it is it ommitted it is considered a scope with name "" (so the scope can be set from a higher level flow...

Now this was just something that came out  of my head and I'm sure there are issues with this approach also.

**Comment by amuniz** [comment]
> So are you arguing that we should go back to my original proposal, which was that milestone take a mandatory nonnegative integer parameter

I guess so. It would solve the use case although it would be really annoying for people doing it right (avoid putting milestones inside conditionals) so all builds have comparable milestones.

> Say these flows run for days. What happens if that sub flow logic is changed and a new milestone is added to it

Well, I think you are trying to support use cases that can generate bizarre results not only in this case but is most cases. Editing a flow and running new builds while there are other builds running the old version is kind of _let's go to break everything_. Actually I think it would break `stage` and stage views also. There are much more corner cases when two consecutive builds are not comparable, and not only here, but even in freestyle jobs using plain publishers and mutual exclusion. Do we actually want to take this use case into account?

**Comment by jtnord** [comment]
>  it would be really annoying for people doing it right 

it's not about right or wrong - even if you do it right as you say what you have is broken, as a newer version of a flow can introduce new milestones in arbitrary places and you then end up comparing an apple with an orange.

**Comment by jglick** [comment]
I think of milestones as something that would really live at top level (and be used sparingly), not something you would ever want to pull in from a library, because of their very nature.

Being able to support evolution of pipeline definitions without halting work in progress is indeed important.

**Comment by amuniz** [comment]
> you then end up comparing an apple with an orange.

Yes, and do we want to manage and control that case? or just say _hey, if you do this you will be comparing apples and oranges, just don't do it_.

Moreover, the wrong corner case would be visible _only_ during the lapse of time where an old build is running and new builds - whit the new configuration - overlap. So if the script is modified when there are no running builds, everything is ok.

**Comment by jtnord** [comment]
I feel I need to :bug: this as the use case in what this is purporting to do is no clear to me.  It is (in conjuntion with the lockable resources plugin support) supposed to be a replacement for `stage concurrency: somenumber` but infact I fail to see how this can actually be the case.

It doesn;t seem to work with imported flow snippets that themselves may be doing locking or milestones.  So the actual use case here seems a little restricted from what I beleive this use case needs to support - which was the original intention of the JamesNord opperator - however it did not work due to out of order builds.

**Comment by amuniz** [comment]
@jtnord 

> supposed to be a replacement for stage concurrency: somenumber

Well, it would be a replacement of `stage concurrency: 1`

> It doesn;t seem to work with imported flow snippets that themselves may be doing locking

No, it works.

> or milestones

Right. Already discussed in previous comments. I guess this is your main concern, right?

> So the actual use case here seems a little restricted from what I beleive this use case needs to support - which was the original intention of the JamesNord opperator

Could you be more specific please? What would need to be fixed in this PR to be acceptable for you? To consider the _edit + trigger new builds_ case?

File: support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep.java [hunk 11]
```diff
@@ -0,0 +1,44 @@
+package org.jenkinsci.plugins.workflow.support.steps;
```
jglick: FYI in #369 I expect to move `StageStep` out of its current location, so expect this to have a new home too.

HRMPW:
Maybe I'm missing something but based on my understanding this feature is designed to ensure that older runs of the same pipeline stop execution when they reach a milestone already passed by a newer run. This can help to ensure that lockable resources aren't locked by obsolete builds, that older code is never deployed over newer code, and just to save processing cycles wasted on older builds that might have gotten hung up along the way.

If this is the case, why does the order of milestones within a specific build matter at all? What matters is what comes after the milestone not what comes before it. If each milestone is unique (regardless of order) then it is comparable across runs.

In @jtnord example:

Before a change:
```
milestone < testing
milestone < deployment
```

After a change:

```
milestone < smoke
milestone < integration tests
milestone < deployment
```

If we assume that each milestone is unique based on its designation then the 'testing' milestone disappears and two new milestones are created for 'smoke' and 'integration tests' while 'deployment' stays the same.

Build 100 sits in 'testing' for 2 days while Build 101 passes quickly through 'smoke' before starting 'integration tests'. However, because 'deployment' is the same unique milestone across both builds if Build 101 finishes 'integration tests' before Build 100 finishes 'testing' then the 'deployment' milestone would stop Build 100. Otherwise, if they reach 'deployment' in order then both pass through to whatever comes after that milestone.

Forcing uniqueness across milestones might cause problems with naming but that is an easy problem for an end user to solve. They don't have to worry about execution order or anything else. They only have to make unique milestones at important junctures of their pipeline.

michaelneale:
"They only have to make unique milestones at important junctures of their pipeline."

Wouldn't it make sense to at least let people group this with stage/label/whatever @HRMPW ?

@jglick says "I think of milestones as something that would really live at top level (and be used sparingly)"
once again - a simple step that can be thrown anywhere doesn't seem to suit milestone well. It sounds like there is yearning for more structure in the script than what current solutions are proposing. 

In current terminology and thinking (even outside jenkins pipeline) *stage* is what most people reach for here (so, kinda back where we started).

amuniz:
@HRMPW 

> ensure that older runs of the same pipeline stop execution when they reach a milestone already passed by a newer run

Right.

>  that older code is never deployed over newer code

Right.

> This can help to ensure that lockable resources aren't locked by obsolete builds

No. If a build already locked a resource a newer build reaching the lock will just wait. But this behavior is not in `milestone` but in `lock` ([PR in lockable-resources-plugin](https://github.com/jenkinsci/lockable-resources-plugin/pull/25)). However, if an older build is locking a resource and a newer one passes the next milestone, then the older build is cancelled, for example in this script:

```
milestone
lock('resource') {
  [do stuff]
}
milestone
```

The aim of `milestone` is just keep the builds passing in ascending order.

>  If each milestone is unique (regardless of order) then it is comparable across runs.

Each `milestone` is granted to be unique by build, but not across builds. Here resides the issue pointed by @jtnord.

In conclusion: using a positive incremental integer parameter (not necessarily sequential) covers all use cases (except the sub-flows loaded via `load`).

@michaelneale 

> a simple step that can be thrown anywhere doesn't seem to suit milestone well.

Using the _numbered_ approach you could use milestones everywhere (except inside `parallel` branches), taking care about numbering in sub-flows.

amuniz:
To summarize, we have to way to go here:

 1. `milestone` without any required argument but not managing _run after script milestones edition when builds already in flight_
 1. `milestone` with a required positive incremental integer parameter (no necessarily consecutive) with the resulting UX issue when using externally loaded scripts via `load`.

What do you think? If some of you have an idea on how to manage both cases at the same time, please throw it here.

michaelneale:
In the case of people changing the flow while things are in progress - wouldn't this be best solved by multibranch? ie with multibranch, the pipeline stays with the branch, and thus is static by definition (to change the pipeline is to create a new branch). Does this solve some confusion?

jtnord:
@michaelneale I don't see how multibranch would solve this?

you have a long running pipeline `Jenkinsfile` on `master`.  You need to make a change to this.  Are you seriously suggesting making users create a new branch `newmaster` just so that they can make changes and build the workflow?  How would old builds then get killed off (e.g. if you introduced one new milestone - It wouldn't as its a different branch - and as such should be completely unaffected by another branches milestones.
If that is what you are suggesting apart from the obvious what branch are we working on today issue, and old builds not getting killed by newer builds - you have lost all of your trend reporting by using a new branch.

:fearful:

michaelneale:
James, not suggesting users do anything. I am tying to reconcile how the
idea of "changing a pipeline definition" works with multibranch. If you
have a long running pipeline and you push a change to master (of the
Jenkinsfile) then you get a new build (it doesn't do anything with the
inflight one). The inflight one continues on with the Jenkinsfile as it was
right? (Ie a previous commit). If I am wrong then yes this is still the
same problem with multibranch.

michaelneale:
@jtnord Sorry I may be confusing things. I am saying "mutlibranch" mostly for the purposes of Jenkinsfile in a repo triggering a build. As there is no way to change in flight now with multibranch (changing Jenkinsfile say on master just triggers a new build also for master, current running one on master continues as it was with its older version of Jenkinsfile) - there is no "change in flight" problem as you can't do it, but I suspect I miscomprehended the question (and you would like to change a long running inflght pipeline, which you can do now with pipeline jobs that aren't defined in a repo)? 

I guess my point is how common would inflight modifications be vs kill and start again? This is similar to the 90s BPM hassle of long running workflow systems and making changes (in general the default they settled on was to keep the process as it was when it was launched - which was fair as most processes were human based things like insurance claims - not fair to change the process after it is started!). There are other problems beyond milestone if you talk about changing the logic of a long running pipeline - the user may have been asked for input that is no longer relevant, or an artefact is now needed at a later stage that is not there in an earlier version of the pipeline logic. In this case, it is up to the care of the pipeline author to make sure it handles it right (basically they have to think about data migration) isn't it?  

(it is that way now, does milestone/whatever really make things worse in that regard)? 

cc @amuniz

amuniz:
> (and you would like to change a long running inflght pipeline, which you can do now with pipeline jobs that aren't defined in a repo)?

I don't think so. You can not edit the script being executed by an _in flight_ build, the problem pointed by @jtnord is editing the inline defined script in the job configuration (or the `Jenkinsfile` if we are in multibrach, it does not matter) and triggering a new build, then build 1 and build 2 could potentially be non-comparable.

> I guess my point is how common would inflight modifications be vs kill and start again?

Yes, agree. IMO we are currently covering the biggest part of the use cases and remaining corner cases could be warned by adding a documentation entry saying _If you want to edit your milestones, do it when there are no builds running. Any other modification is allowed_. In the worse case (edit the flow when a build is running), a newer build is cancelled erroneously which can be fixed just by manually triggering a new build afterwards.

> (it is that way now, does milestone/whatever really make things worse in that regard)?

Not if we add `ordinal` as a required parameter so every milestone is uniquely identified - as `stage` currently does with `name` - (beyond the UX problem pointed before in Option 2).

And if some of you are wondering why don't use `name` instead of `ordinal` - so the UX issue is gone -, it's not possible, since the key improvement of `milestone` over `stage` is just the ordering based on the comparability of any two `milestone`s in a build.

jglick:
> If you want to edit your milestones, do it when there are no builds running.

I am not convinced this is an unacceptable restriction, even if documented. Some users will require uninterrupted builds in process. Multibranch is a red herring here since sooner or later changes will be merged into an existing branch.

Now whether it is probable that you would _want_ to change your milestone list for an existing job is another question. I suspect the set of milestones should be small and permanent in most cases. But we need to see some realistic use cases—such as a CD demo patch—to evaluate this seriously.

My inclination is to prefer the explicit monotonically increasing ordinal. Then you do not need to guess whether M3 in build 101 corresponds to M2 or M3 or M4 in build 102. You have written `milestone 3` and that is that.

> the resulting UX issue when using externally loaded scripts via `load`

What is the issue, exactly? You can pull in scripts via `load`, typically for the purpose of having some bootstrap code in the main script and the real work defined somewhere else. It would not generally make sense to have two different scripts used in a given build both be calling `milestone`, since the step is by its very nature only meaningful at the build granularity.

> a simple step that can be thrown anywhere doesn't seem to suit milestone well

_Can_ is different than _should_. A block scope is not well suited to this task. Nor was it for `stage` with `concurrency`; one problem with `stage` is that it is too tempting to use it solely as a display marker, which is a use case that _is_ well suited to block scope and inclusion inside `parallel` branches and so on.

jglick:
> In current terminology […] _stage_ is what most people reach for here

Perhaps, but that name is taken.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/MilestoneStepTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+    @ClassRule
+    public static BuildWatcher buildWatcher = new BuildWatcher();
+
+    @Rule
+    public RestartableJenkinsRule story = new RestartableJenkinsRule();
+
+    @Before
+    public void clear() {
+        MilestoneStepExecution.clear();
```
jglick: :ant: do not copy my bad example; store state in an `Extension`-scoped instance field.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStepException.java
```diff
@@ -0,0 +1,11 @@
+package org.jenkinsci.plugins.workflow.support.steps;
+
+public class MilestoneStepException extends RuntimeException {
+
+    public MilestoneStepException(String message) {
```
jglick: :ant: delete this and use `AbortException` instead.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep.java
```diff
@@ -0,0 +1,58 @@
...
+
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+import hudson.Extension;
+
+/**
+ * This step can be used to grant:
+ * <ol>
+ *   <li>Builds pass through the step in order (taking the build number as sorter field)</li>
+ *   <li>Older builds will not proceed (they are aborted) if a newer one already entered the milestone</li>
+ *   <li>Once a build enters the milestone, it will be never aborted</li>
```
jglick: Seems to contradict the following point, does it not?

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/MilestoneStepTest.java
```diff
@@ -0,0 +1,166 @@
...
+                        "milestone()\n" +
+                        "echo 'First milestone'\n" +
+                        "semaphore 'wait'\n" +
+                        "milestone()\n" +
+                        "echo 'Second milestone'\n"));
+                WorkflowRun b1 = p.scheduleBuild2(0).waitForStart();
+                SemaphoreStep.waitForStart("wait/1", b1);
+                WorkflowRun b2 = p.scheduleBuild2(0).waitForStart();
+                SemaphoreStep.waitForStart("wait/2", b2);
+                // Now both #1 and #2 are into milestone 1
```
jglick: I find it easier to think, and write, in terms of “passing” rather than “entering”. They are between M1 and M2.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/MilestoneStepTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+                // Let #2 continue so it leaves milestone 1
+                SemaphoreStep.success("wait/2", null);
+                story.j.waitForCompletion(b2);
+
+                // Once #2 continues and reaches milestone 2 then #1 is automatically cancelled
+                // TODO: determine why it's getting ABORTED instead of NOT_BUILT
```
jglick: :bug: somewhere (possibly elsewhere in infrastructure)

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep.java
jglick: If I understand correctly, this third bullet can simply be deleted.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep.java
```diff
@@ -0,0 +1,58 @@
...
+ *   <li>When a build passes a milestone, any older build that passed the previous milestone but not this one is aborted.</li>
+ * </ol>
+ */
+public class MilestoneStep extends AbstractStepImpl {
+
+    /**
+     * Optional milestone label.
+     */
+    private String label;
```
jglick: Do we really need a label? Seems like a distraction to me.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/MilestoneStepTest.java
```diff
@@ -0,0 +1,166 @@
...
+                SemaphoreStep.waitForStart("wait/1", b1);
+                WorkflowRun b2 = p.scheduleBuild2(0).waitForStart();
+                SemaphoreStep.waitForStart("wait/2", b2);
+                // Now both #1 and #2 are into milestone 1
+
+                // Let #2 continue so it leaves milestone 1
+                SemaphoreStep.success("wait/2", null);
+                story.j.waitForCompletion(b2);
+
+                // Once #2 continues and leaves milestone 1 then #1 is automatically cancelled
```
jglick: You mean, once it exits.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStepExecution.java
```diff
@@ -0,0 +1,443 @@
...
+    public boolean start() throws Exception {
+        if (step.getLabel() != null) {
+            node.addAction(new LabelAction(step.getLabel()));
+        }
+        Integer ordinal = processOrdinal();
+        tryToPass(run, getContext(), ordinal);
+        return true;
+    }
+
+    private synchronized Integer processOrdinal() {
```
jglick: :ant: can and should be `int` I think.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStepExecution.java
```diff
@@ -0,0 +1,443 @@
...
+
+    // Used in tests only (where the static context is kept between tests somehow)
+    @Restricted(DoNotUse.class)
+    public static void clear() {
+        milestonesByOrdinalByJob = null;
+    }
+
+    @Override
+    public void stop(Throwable cause) throws Exception {
+        throw new UnsupportedOperationException();
```
jglick: Can just leave the body empty; ought not be called in any event, since we do not block.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStepExecution.java
```diff
@@ -0,0 +1,443 @@
...
+    private static Map<String, Map<Integer, Milestone>> milestonesByOrdinalByJob;
+
+    @Override
+    public boolean start() throws Exception {
+        if (step.getLabel() != null) {
+            node.addAction(new LabelAction(step.getLabel()));
+        }
+        Integer ordinal = processOrdinal();
+        tryToPass(run, getContext(), ordinal);
+        return true;
```
jglick: Could use `AbstractSynchronousStepExecutionImpl`, since we unconditionally return true from `start` (never block).

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStepExecution.java
```diff
@@ -0,0 +1,443 @@
...
+    public static void clear() {
+        milestonesByOrdinalByJob = null;
+    }
+
+    @Override
+    public void stop(Throwable cause) throws Exception {
+        throw new UnsupportedOperationException();
+    }
+
+    private static synchronized void tryToPass(Run<?,?> r, StepContext context, Integer ordinal) {
```
jglick: :ant: use `int`

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStepExecution.java
```diff
@@ -0,0 +1,443 @@
...
+         * Last build that passed through the milestone, or null if none passed yet.
+         */
+        @CheckForNull
+        Integer lastBuild;
+
+        /**
+         * Last build extenalizable ID hat passed through the milestone, or null if none passed yet.
+         */
+        @CheckForNull
+        String lastBuildExternalizableId;
```
jglick: :ant: redundant to store this when we have `lastBuild`; the ID simply consists of the job name (which we already know) and the build number (for Jenkins 1.597+, which we must be running for this new step).

amuniz:
> My inclination is to prefer the explicit monotonically increasing ordinal. Then you do not need to guess whether M3 in build 101 corresponds to M2 or M3 or M4 in build 102

Agree. The implementation would be actually easier than the current one (which is traversing the flow graph to search the previous ordinal).

> What is the issue, exactly? You can pull in scripts via load, typically for the purpose of having some bootstrap code in the main script and the real work defined somewhere else. It would not generally make sense to have two different scripts used in a given build both be calling milestone, since the step is by its very nature only meaningful at the build granularity.

That's my understanding too, but @jtnord is really concerned about the maintainability of milestones when using them in externally loaded scripts.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep.java
amuniz: It was fixed in `help.html` but I forgot to update it here. Fixed now.

support/src/main/java/org/jenkinsci/plugins/workflow/support/steps/MilestoneStep.java
amuniz: No, not needed, but someone (actually more than one I think) asked me somewhere to add this **optional** label just for convenience (added as a `LabelAction` later in `MilestoneStepExecution` so it's usable by others.

aggregator/src/test/java/org/jenkinsci/plugins/workflow/steps/MilestoneStepTest.java
amuniz: Yes. The cause of this is that a call to `Run#getExecutor().interrupt(Result, CauseOfInterruption ... causes)` in a `WorkflowRun` in turn calls `AsynchronousExecution#interrupt(boolean)` which calls the specific implementation of `AsynchronousExecution` in `WorkflowRun`, which  finally calls `execution.interrupt(Result.ABORTED);`, so the initial `Result` is omitted.

The above behavior is a bug IMO, but not fixable without API changes I think.

Anyway it's locally avoidable by calling `((FlowExecutionOwner.Executable) r).asFlowExecutionOwner().get().interrupt(Result, CauseOfInterruption ...);` directly in `MilestoneStepExecution`. Fixing...

jglick: > The above behavior is a bug IMO, but not fixable without API changes I think.

I suspect this suffices:

```diff
diff --git a/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java b/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
index 92ca05c..450fe84 100644
--- a/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
+++ b/job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
@@ -227,7 +227,7 @@ public final class WorkflowRun extends Run<WorkflowJob,WorkflowRun> implements F
                     return;
                 }
                 try {
-                    execution.interrupt(Result.ABORTED);
+                    execution.interrupt(getExecutor().abortResult());
                 } catch (Exception x) {
                     LOGGER.log(Level.WARNING, null, x);
                 }
```

Please try it and see.

michaelneale:
Still :bug: as making users explicitly order things like BASIC for the rare case of wanting to modify in-flight pipelines seems wrong.

jglick:
What about having an _optional_ ordinal, inferred by counting if unspecified?

My preference is still to have a mandatory parameter. I do not see how

```groovy
doSomething()
milestone 1
doSomethingElse()
milestone 2
```

is any harder to either type or read than

```groovy
doSomething()
milestone()
doSomethingElse()
milestone()
```

which is more complex to implement and more prone to subtle errors.

michaelneale:
@jglick if you reduce it to something trivial of course it is easy. It is more realistic people will have reams of things between milestones (as we have seen) then it will be quite different and more confusing. Optional would be preferable. 

Is GOTO already taken ? ;)

jglick:
I actually would love to implement INTERCAL’s `COME FROM`.

michaelneale:
yes `COME FROM` from [intercal](https://en.wikipedia.org/wiki/INTERCAL) could be good for handling errors and sending notification emails. Also `PLEASE GIVE UP` is a good way to cancel a build at any point.

TG9541:
From a background of a more exotic application (embedded control, e.g. in the automotive industry), I've read the comments here with the greatest interest. Since the decisions taken here are likely to have some impact on the usability of the product "Jenkins Pipeline" I'd like to shed some light on some aspects as long as the  topic is still "fresh".

While I mostly agree with @jglick 's conclusions in the answer 6 days ago, but for different reasons. The statement "Multibranch is a red herring here since sooner or later changes will be merged into an existing branch" gets my attention, as this doesn't apply to the case where branches represent configurations (-> [ISO10007](https://www.iso.org/obp/ui/?_escaped_fragment_=iso:std:36644:en#!iso:std:36644:en)), and thus @jglick 's argument is only valid in a limited field (feature branches). IMO, **this** is a "red herring"!

IMO, one really needs to separate the concerns of a the "business process" (like quality assessments, or release decisions), and the  "build mechanics" (like parallelism, concurrency, or resource locking), and clear semantics of arguments in this discussion is necessary. I think that this is what @jglick had in mind when he stated that "the step is by its very nature only meaningful at the build granularity". 

In the field of embedded software, one must keep the scopes of "product configuration", and "development life-cycle" separate. In fact, milestones in different "configuration branches" **must not** interfere with each other, as they may be applied to different deliverables (an exception is when common components are used, and the quality process applies to components!). So called "variant builds" are a different case, since here a set of quality goals for different configurations must be met by *one and the same* product revision, and the CPU resources necessary for such variant build will make you weep.

I would also prefer "junction" (to use @HRMPW 's term), or "gate" (as in "quality gate") over "milestone" (which is a well defined term in project management). @michaelneale 's comment on BPM noteworthy, too. @michaelneale is right, when he states that the "COME FROM" in jglick 's remark has a serious application: an inversion of the flow control would enable the automation of "picking" a run based on arbitrary metrics (i.e. decisions based on a set of QM KPIs).

In the case of "feature branches" (which are to be merged at some point): here, we clearly are at an early point in an iteration (as a part of the "application life-cycle"), before a delivery is even desirable. Therefore, build definitions for feature branches shouldn't contain a "delivery" (-> milestone, junction, or quality gate), but they may well compete with other configurations ("branches") for "build resources" like CPU/executors, Selenium farms, or even HIL test rigs (-> stages, locks).

To sum it up: while two product configurations will make use of the same "machinery" for building and testing, the QM processes may only interfere with each other on a level controlled by configuration management, and build automation must support this separation.

amuniz:
@jtnord's concerns are only _fixable_ by adding explicit ordinals as parameter, and @michaelneale is bugging that option, so.... we have crossed bugs here: please guys, you have to reach an agreement, this PR will be blocked forever otherwise.

FYI: current code status is **no** ordinal parameter (the code generates and assumes consecutive ordinals per build). So I guess @michaelneale's bug is just preventive, right?

I kind of agree with both of you, but my inclination is to add ordinal as parameter so we can safely manage a bigger user-cases space.

amuniz:
@TG9541 Thanks for your thoughts.

> milestones in different "configuration branches" must not interfere with each other

How are you managing this currently? I can not think in any use of `stage` with `concurrency` to manage concurrency **only** in the build scope.

> I would also prefer "junction" (to use @HRMPW 's term), or "gate" (as in "quality gate") over "milestone" 

I'm open to suggestions. Changing the name is trivial **before** releasing, but unless there is a consensus to change it to something else, I'll keep `milestone`.

TG9541:
@amuniz 

> How are you managing this currently? I can not think in any use of stage with concurrency to manage concurrency only in the build scope.

That's very simple: one build definition per configuration. We apply some automation to creating build definitions, but I had hoped that the build management overhead could be reduced by introducing something like multibranch , or literate.

> I'm open to suggestions. Changing the name is trivial before releasing, but unless there is a consensus to change it to something else, I'll keep milestone.

That's OK, terminology isn't that important. What's important is the semantics.

michaelneale:
My :bug: is specifically on *mandatory* ordinals in all cases

jglick:
> My bug is specifically on mandatory ordinals in all cases

So would you be satisfied by my proposed compromise, to allow an ordinal to be defined

```groovy
milestone 3
```

but if omitted

```groovy
milestone()
```

to be inferred by counting?

> I would also prefer "junction" […] or "gate" […] over "milestone"

No strong opinion.

> this doesn't apply to the case where branches represent configurations

Sure, there could be (semi-)permanent branches using this step too. In such cases the milestones are compared per job, which is to say, per branch. My comment was in response to some earlier discussion about changing the `master` Pipeline definition in a temporary topic branch; in such a case you would probably have milestones disabled on these non-delivery branches (for example, by checking `env.BRANCH_NAME`), where only plain old resource locking matters, so the change to milestone semantics would only be of interest once the branch were merged.

Actually there _is_ a potential use for milestones even for non-delivery branches: if you are building pull requests or the like, and do not want to waste computer time building anything but the latest version of the code (i.e., you do not really care whether some earlier commits were bad, you will just wait for the latest result), you could start every script with

```groovy
milestone 1
```

to immediately abort any older build. I believe the Gerrit Trigger plugin does something like this automatically, in the case of a new build coming from the same “change” (updated patch).

jtnord:
> Actually there is a potential use for milestones even for non-delivery branches ...  I believe the Gerrit Trigger plugin does something like this automatically, in the case of a new build coming from the same “change” (updated patch).

Yes, that option exist and is very useful :)

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java
```diff
@@ -227,7 +227,7 @@ private AsynchronousExecution sleep() {
                     return;
                 }
                 try {
-                    execution.interrupt(Result.ABORTED);
+                    execution.interrupt(getExecutor().abortResult());
```
jglick: Extracted to https://github.com/jenkinsci/workflow-job-plugin/pull/1.

jglick:
Should be closed and refiled as a “creation PR” against an empty repository.

amuniz:
Closing in favor of https://github.com/jenkinsci/pipeline-milestone-step-plugin/pull/1

PR #360, opened 2016-03-10
Bump durable task plugin, specifically because of JENKINS-27152

svanoort:
This will fix some counterintuitive hangs due to the control files being inadvertantly removed (also downstream, it will be important for the docker pipeline plugin to take up after the next pipeline release).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,7 +76,7 @@
             <dependency>
                 <groupId>org.jenkins-ci.plugins</groupId>
                 <artifactId>durable-task</artifactId>
-                <version>1.5</version>
+                <version>1.8</version>
             </dependency>
             <dependency>
                 <groupId>org.jenkins-ci.plugins</groupId>


svanoort:
@reviewbybees

jglick:
Obviously I published this release and the changes in it because I consider them useful, but what is the motivation for bumping up the dependency here? Just to get integration testing?

jglick:
In other words, what is gained by forcing users to accept the `durable-task` update, when they would probably get it anyway and might have had some reason to be more conservative?

> it will be important for the docker pipeline plugin to take up after the next pipeline release

`docker-workflow` does not _require_ this update. It was modified to _accept_ it. In other words, the tricky aspect of JENKINS-27152 was that if it were changed solely in `durable-task`, `docker-workflow` would be broken. But I published `docker-workflow` 1.4 to mount the new temporary files directory; it continues to mount the workspace, where older versions of `durable-task` put temporary files.

svanoort:
>  what is gained by forcing users to accept the durable-task update

They get the fix out-of-box, which makes it easier to avoid hangs or issues around the control files.  For me, it was 'git clean -fdx' that triggered it, for others it might be something else.  Triggered by wondering if docker-workflow out-of-box includes the fix, since I'm advocating for it to be part of the setup wizard and this is one thing that can trip up a new user.

I suppose I should ask, since it sounds like you have something in mind: is there a specific reason you're nervous about bumping the version?

jglick:
> is there a specific reason you're nervous about bumping the version?

Well, it forces people to update whether they want to or not. I have unconfirmed reports of problems like [JENKINS-32000](https://issues.jenkins-ci.org/browse/JENKINS-32000).

> I'm advocating for it to be part of the setup wizard and this is one thing that can trip up a new user.

Anyone using the 2.0 setup wizard will get the latest available release of every plugin anyway.

PR #362, opened 2016-03-12
Using structs plugin

kohsuke:
`DescribableHelper` functionality has moved to structs plugin, in preparation for the death of `$class`

### DIFF
--- a/step-api/pom.xml
+++ b/step-api/pom.xml
@@ -42,7 +42,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.0-20160312.085007-3</version>
+            <version>1.0</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/step-api/pom.xml
+++ b/step-api/pom.xml
@@ -42,7 +42,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.0</version>
+            <version>1.1</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/step-api/pom.xml
+++ b/step-api/pom.xml
@@ -42,7 +42,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.0-SNAPSHOT</version>
+            <version>1.0-20160312.085007-3</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/step-api/pom.xml
+++ b/step-api/pom.xml
@@ -41,6 +41,11 @@
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>structs</artifactId>
+            <version>1.0-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>git</artifactId>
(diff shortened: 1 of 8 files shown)


kohsuke:
I talked to Jesse offline about this, but otherwise @reviewbybees

step-api/pom.xml
```diff
@@ -41,6 +41,11 @@
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>structs</artifactId>
+            <version>1.0-20160312.085007-3</version>
```
abayer: I assume there'll be an actual release before Pipeline's next release, yes? =)

kohsuke: Before this merges, yes. This is considered a part of the exit criteria for the structs plugin.

abayer: Then I am :bee:

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
```diff
@@ -73,6 +73,9 @@
 /**
  * Utility for converting between {@link Describable}s (and some other objects) and map-like representations.
  * Ultimately should live in Jenkins core (or Stapler).
+ *
+ * @deprecated
+ *      Moved to the structs plugin
```
amuniz: It seems it is not really moved but moved-renamed (it is now `DescribableModel `). Maybe a link to the new class in the javadoc makes sense (and developers will be thankful!)

jglick:
On my list to review…

step-api/src/main/java/org/jenkinsci/plugins/workflow/structs/DescribableHelper.java
jglick: Yes ought to be using `{@link DescribableModel}`.

step-api/src/main/java/org/jenkinsci/plugins/workflow/steps/StepDescriptor.java
```diff
@@ -110,7 +111,7 @@ public Step newInstance(Map<String,Object> arguments) throws Exception {
      * @throws UnsupportedOperationException if this descriptor lacks the ability to do such a calculation
      */
     public Map<String,Object> defineArguments(Step step) throws UnsupportedOperationException {
-        return DescribableHelper.uninstantiate(step);
+        return DescribableModel.uninstantiate_(step);
```
jglick: :ant: an oddly-named method. I would just delete it. Here it is not cumbersome to write

```java
return new DescribableModel(step.getClass()).uninstantiate(step);
```

and for the usage in `InputStepTest` it is easy enough to hardcode `InputStep.class`.

jglick: (Seems your initial code drop in `symbol-plugin` did not go through the PR process so there is no easy way to comment on that code directly. Generally I would suggest starting with an empty repository and then filing PR 1 to add initial proposed content.)

jglick:
:bee:

kohsuke:
Feedbacks to incorporate:

- [ ] remove `uninstantiate_`
- [ ] javadoc fix

kohsuke:
On 2nd thought I'm keeping `uninstantiate_` method because the instance method version is fundamentally redundant by taking two arguments. The underscore in the end and not in the beginning is to assist auto-completion discoverability.

It's up to calling code to decide what method to call, so if it bothers you, feel free to use the non-static version.

jglick:
@kohsuke you seem to think you released 1.0, yet you did not actually do so. There is no such release in the repository, and https://github.com/jenkinsci/structs-plugin/commits/master shows no release commit, though there is a https://github.com/jenkinsci/structs-plugin/tree/structs-parent-1.0 tag.

jglick:
:bee:

PR #368, opened 2016-03-18
Switch baseline to 1.642.3

jglick:
Pick up yesterday’s LTS release, so that we no longer suffer from the bug fixed in https://github.com/jenkinsci/jenkins/pull/1989, which gravely affects Pipeline build resumption.

Still waiting for `jenkins:1.642.3` to appear in Docker Hub.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,7 @@
       <tag>HEAD</tag>
   </scm>
     <properties>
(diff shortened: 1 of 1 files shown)


andresrc:
:bee: we faced in the ATH.

amuniz:
:bee:

PR #369, opened 2016-03-18
Reorganization

jglick:
As this plugin suite has grown, maintaining all of it in one big tree and doing massive synchronized releases has grown tedious. Would like to rearrange sources so that
* `workflow-aggregator` is hosted in its own repository, with its own lifecycle, to include the CD demo, and can depend on any thematically related plugins we like
* functional tests live in the closest possible plugin
* it is possible to move other plugins into their own repositories with their own lifecycles (especially the more “downstream” ones like `workflow-cps-global-lib` or `workflow-multibranch`)

- [X] break `SNAPSHOT` dependencies
- [X] move demo
- [x] switch to 2.0 versions
- [x] move `ExecutorStep` and `WorkspaceStep` into `workflow-durable-task-step`
- [x] move `StashStep` and `UnstashStep` into `workflow-basic-steps`
- [x] move `BuildTriggerStep` into its own plugin
- [x] move `InputStep` into its own plugin
- [x] move `StageStep` into its own plugin
- [x] move functional tests
- [ ] ~~kill `SingleJobTestBase`~~ (can be finished up later)
- [x] clean out root `pom.xml`
- [x] restore snapshot functionality to demo

@reviewbybees esp. @kohsuke


jtnord:
> move StageStep into its own plugin

Should that not be in the pipeline-stage-view-plugin (as that will be it's only reason for living when milestone/locks are available)?

jglick:
> Should that not be in the pipeline-stage-view-plugin (as that will be it's only reason

Well TBD if it is even needed for labeling (cf. #140).

aggregator/demo/snapshot-only-plugins.txt
```diff
@@ -0,0 +1,3 @@
+pipeline-build-step
+pipeline-input-step
+pipeline-stage-step
```
amuniz: Could any plugin be added here? I mean, plugins outside this repository?

api/src/main/java/org/jenkinsci/plugins/workflow/FilePathUtils.java
```diff
@@ -39,64 +38,82 @@
...
-    private final String path;
-
-    private FilePathPickle(FilePath v) {
-        // TODO JENKINS-26096 switch to FilePath.toComputer once it uses something like Listener here and so is actually reliable
-        slave = Listener.getChannelName(v.getChannel());
-        if (slave == null) {
-            throw new IllegalStateException("no known slave for " + v);
-        }
-        path = v.getRemote();
+public class FilePathUtils {
```
amuniz: Are changes in this class needed for the refactor? If not, I would propose to exclude it from this PR.

aggregator/demo/snapshot-only-plugins.txt
jglick: Yes! That is a key aspect of this change—the source location of plugins no longer matters for purposes of the aggregator or demo, so they could be moved into their own repositories, and there is no longer reduced support for plugins not in the original `org.jenkins-ci.plugins.workflow` group.

api/src/main/java/org/jenkinsci/plugins/workflow/FilePathUtils.java
jglick: The changes are needed so that code can be shared across several plugins. Formerly it was mostly in `workflow-support` (with a little copied into `workflow-api`); now there is one class defined in `workflow-api` and used in `workflow-api`, `workflow-support`, and `workflow-durable-task-step`.

kohsuke:
:bee:
Where is the new location of aggregator & demo?

jglick:
> Where is the new location of aggregator

For now, still in the same place. But I would like to move it into its own repository.

> & demo

Simply moved under `aggregator`, with the expectation that this would wind up in the same repository too. I think it makes sense to associate them.

PR #370, opened 2016-03-23
Remove DescribableHelperTest

jglick:
Cleanup from #362.

@reviewbybees esp. @kohsuke


kzantow:
... seems it would be worthwhile to keep this until a class is completely removed to me. Is this test particularly flaky or slow?

jglick:
> worthwhile to keep this until a class is completely removed

It is deprecated and no longer used in this repository. I do not plan on modifying it except at some point to delete it.

> Is this test particularly flaky or slow?

No, just want to clean up so there is not unnecessary duplicated code floating around; the test was already copied (with some refactoring) into the `structs` plugin.

kwhetstone:
:bee: All for cleaning up unused code. Should delivering this be conditional on [jenkinsci/structs-plugin/pull/1](https://github.com/jenkinsci/structs-plugin/pull/1) ?

kzantow:
:bee: sure, duplicated code is bad, mmkay?

jglick:
> Should delivering this be conditional on jenkinsci/structs-plugin/pull/1 ?

I do not think so—this is just a test of dead code. We _could_ have made #362 conditional on some resolution to that problem, I just did not notice at the time that the copied code was an old snapshot.

PR #371, opened 2016-03-29
[JENKINS-32925] Handle recursion

kohsuke:
Use stack to keep track of traversal.

Counter proposal to https://github.com/jenkinsci/structs-plugin/pull/1

@reviewbybees in particular @kwhetstone


jglick:
:bug: forgot `dsld` and `gdsl` views

kohsuke:
I didn't forget dsld and gdsl. Those do not have the recursion so they are not affected by this problem.

jglick:
OK :bee: in that case.

kwhetstone:
I'm fine with moving the stack management to the classes that use the DescribableModel, but there should be some doc there so anyone who uses this knows that recursion could cause a problem.

kwhetstone:
:bee:

PR #372, opened 2016-03-31
[JENKINS-26481] [JENKINS-27421] Fix ArrayList$Itr, and integration test for Object.each(Closure)

jglick:
[JENKINS-26481](https://issues.jenkins-ci.org/browse/JENKINS-26481)

More realistic test for https://github.com/cloudbees/groovy-cps/pull/23. Requires https://github.com/jenkinsci/script-security-plugin/pull/50 also. Additionally fixes the longstanding irritation with Java 5-style `for` loops, [JENKINS-27421](https://issues.jenkins-ci.org/browse/JENKINS-27421).

@reviewbybees esp. @kohsuke


jglick:
Confirmed:

* with sandbox off, `CpsDefaultGroovyMethods` is used, but the unserializable iterator breaks the build
* with sandbox on, `DefaultGroovyMethods` is used

kohsuke:
Thanks for the help. `Iterator` being unserializable is expected and that's fine. We use JBoss marshaller so that we can serialize those unserializable objects anyway.

The second one is unexpected so it needs to be looked at further.

jglick:
> We use JBoss marshaller so that we can serialize those unserializable objects anyway.

Already tried that. Does not work. #206 has details.

kohsuke:
So the problem at hand is that `SandboxInterceptor.onMethodCall` is hard-coded to look for a matching method on `DefaultGroovyMethods`. I think it needs to use `MetaClass` to look up the target of dispatch.

jglick:
Not following why that would matter—the lookup is only to see if the method is whitelisted, but if it is, then the actual delegation proceeds according to the `super` call which should behave as you want.

kohsuke:
Argument #1:
  - `[1,2,3].each { ... }` should be checked against `Object.each(Closure)` which is the 'declared' method signature.
    Whether this is defined as the native Java method or by a pseudo method like `DGM` is Groovy's
    implementation detail, and users shouldn't care the difference.
  - Then the bug is that `SandboxInterceptor.onMethodCall(...)` explicitly looks for GDK methods like the following.
    It should be just entirely commented out:
````
            // look for GDK methods
            Object[] selfArgs = new Object[args.length + 1];
            selfArgs[0] = receiver;
            System.arraycopy(args, 0, selfArgs, 1, args.length);
            if (GroovyCallSiteSelector.staticMethod(DefaultGroovyMethods.class, method, selfArgs) != null) {
                return onStaticCall(invoker, DefaultGroovyMethods.class, method, selfArgs);
            }
````

Argument #2:
  - `[1,2,3].each { ... }` should check against `DGM.each(Object,Closure)` because that's where the definition is,
    at least conceptually, but we should then let Groovy dispatch the call as
    `Object.each(Closure)`, like it does usual, to ensure maximum compatibility with Groovy method dispatching behaviour.
  - Today, after the whitelist check against `DGM.each(Object,Closure)` is performed, `SandboxInterceptor` dispatches
    the code as `DGM.each([1,2,3],{...})` which obviously is not going to work with `CpsDGM`.
  - CON: how do we know if Groovy actually dispatches a call to the target method? Isn't this too prone to vulnerabilities?

Argument #3:
  - `SandboxInterceptor` shouldn't be duplicating `MetaClass.invokeMethod` method selection logic because we'll never
    do it right, as this bug has illustrated. Let's let Groovy pick a method like it does during invocation,
    which is:
    - pick `MetaMethod`
    - call `MetaMethod.invoke(...)`
  - Thus the fix is to have `Checker` (in groovy-sandbox) do the following to pick the meta method, which handles
    DGM and all that, then introspect the obtained `MetaMethod` to determine where the dispatch is going:
  
            MetaMethod m = getMetaClass(_receiver).getMetaMethod(_method,_args);

  - CON: `MetaMethod` to DGM is a synthetic class like `dgm$170` and it's not obvious how to look at that instance
    and figure out that it's dispatching to `DefaultGroovyMethods.each()`

Argument #4:
  - `[1,2,3].each { ... }` should check against `CpsDGM.each(Object,Closure)` because that's where the actual definition is.
  - The fix would be then to open up `SandboxInterceptor.DGM_CLASSES` so that we can insert `CpsDGM` in there.
  
Consideration: backward compatibility with people's existing custom whitelist
  - #2 is backward compatible. #1 isn't.
  - #3 is kind of backward compatible but this ends up making people whitelist `CpsDGM` sometimes and `DGM` for others.


I think #2 is the minimum change to make it work, but I think #1 is a better option long term. WDYT?

**Comment by jglick** [comment]
The first option is no option at all, if I understand it correctly, since `Whitelist` operates on real `AccessibleObject`s.

The second option is how `script-security` works normally. So I think the problem is only that `SandboxInterceptor` winds up calling `super.onStaticCall` with `DefaultGroovyMethods`, bypassing your patching code. Perhaps rather than delegating to `onStaticCall`, it should just do the whitelist check but then proceed with the original `super.onMethodCall`. That seems pretty easy.

The third option sounds like a much bigger change if I read it right, though probably in the long term it is what we want.

The fourth option does not seem right.

**Comment by kohsuke** [comment]
:bee:

File: cps/src/main/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHack.java [hunk 1]
```diff
@@ -0,0 +1,114 @@
...
+            for (Map.Entry<Class,Map<String,Object>> entry : fields.entrySet()) {
+                Class definedIn = entry.getKey();
+                for (Map.Entry<String,Object> entry2 : entry.getValue().entrySet()) {
+                    String fieldName = entry2.getKey();
+                    Object value = entry2.getValue();
+                    rp.writeField(o, fieldName, value, definedIn);
+                    if (fieldName.equals("this$0")) {
+                        try {
+                            Field f = AbstractList.class.getDeclaredField("modCount"); // TODO if not handling only ArrayList.Itr, search in superclasses
+                            f.setAccessible(true);
```
kzantow: Any concern about security restrictions in different environments, this being an internal JVM class? Could you not just transform the list and/or iterator to an alternate serializable version?

jglick: > Any concern about security restrictions

You mean, on `setAccessible`? No, Jenkins routinely assumes it has full permissions.

> Could you not just transform the list and/or iterator to an alternate serializable version?

You could transform the iterator into an alternate version, but that would be relying even more on the details of the original’s state. Transforming the list would not work, I think, because we are not given the opportunity to replace it with something else—we can only replace which objects we _save_, and the problem here is the unserializability of the result of `iterator()`.

cps/src/main/java/org/jenkinsci/plugins/workflow/cps/persistence/IteratorHack.java
```diff
@@ -0,0 +1,114 @@
...
+    private static class Replacement extends Pickle {
+
+        private final Class type;
+        private final Map<Class,Map<String,Object>> fields;
+
+        Replacement(Object o) {
+            type = o.getClass();
+            fields = new HashMap<>();
+            ReflectionProvider rp = Jenkins.XSTREAM2.getReflectionProvider();
+            rp.visitSerializableFields(o, new ReflectionProvider.Visitor() {
```
kzantow: :ant: could cache this info per type, especially since this is for exactly one type

kzantow:
:bee: lots of people will be happy for this, @jglick

PR #374, opened 2016-04-01
[JENKINS-33971] Integration test for ReverseBuildTrigger.upstream2Trigger bug

jglick:
Downstream of https://github.com/jenkinsci/jenkins/pull/2207.

@reviewbybees esp. @fbelzunc

### DIFF
--- a/job/src/test/java/org/jenkinsci/plugins/workflow/ReverseBuildTriggerTest.java
+++ b/job/src/test/java/org/jenkinsci/plugins/workflow/ReverseBuildTriggerTest.java
@@ -1,0 +1,95 @@
+/*
+ * The MIT License
+ *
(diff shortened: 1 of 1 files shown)


amuniz:
:bee: although I think an acceptance test (in `acceptance-test-harness`) would be more useful. ATH is run and checked on every Jenkins core release (where the code responsible of this behavior lives), so if some day it regresses we are not going to notice it a the test lives here (in `workflow-plugin`). Unless someone is running and monitoring PCT on every core release, that I'm not sure.

jglick:
> an acceptance test […] would be more useful

Substantially harder to write in this case, as well as slower.

> Unless someone is running and monitoring PCT on every core release

Well if no one is not yet, then we should start.

PR #375, opened 2016-04-04
[JENKINS-31162] Support for Item categorization in Jenkins 2

recena:
This PR belongs to this [JENKINS-31162](https://issues.jenkins-ci.org/browse/JENKINS-31162) issue and allows to categorize the Pipeline jobs in a right way.

@reviewbybees especially @jglick


job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
```diff
@@ -597,6 +597,39 @@ public static void alias() {
             return new WorkflowJob(parent, name);
         }
 
+        /**
+         * Needed if it wants Pipeline jobs are categorized in Jenkins 2.x.
+         *
+         * TODO: Override when the baseline is upgraded to 2.x
```
amuniz: I would add: _and replace `standaloneprojects` string with `StandaloneProjectsCategory.ID`_

amuniz:
:bee:

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
recena: @amuniz But `StandaloneProjectsCategory.ID` is only available in Jenkins 2.0.

recena:
@amuniz Thanks. I forgot this include this https://github.com/jenkinsci/workflow-plugin/pull/375/commits/6bea194f47bd19bca555b87e9bbd4152c2f402f5 This was a @jglick recommendation [here](https://github.com/jenkinsci/external-monitor-job-plugin/pull/9#discussion_r57181394).

job/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowJob.java
amuniz: Yes, I know, I meant to add it in the Javadoc.

recena: Ah, good idea.

recena: @amuniz Done here https://github.com/jenkinsci/workflow-plugin/pull/375/commits/98a1afc7748398d08df1bdf068e038baee353934

amuniz:
re-:bee:

jglick:
The icon implies that there is a fix of [JENKINS-30363](https://issues.jenkins-ci.org/browse/JENKINS-30363). @gusreiber is this the icon you approve for that purpose?

jglick:
BTW

> To prevent XSS

is misleading—using `j:out` _disables_ XSS defense in this case, because we know there is no attack from this source.

Anyway, :bee:

recena:
@jglick Thanks for this note. Then I have to view `<j:out />` and `<?jelly escape-by-default='true'?>` because I misunderstood how work.

jglick:
```xml
<?jelly escape-by-default='true'?>
```

means that all output from `${expression}` is escaped (`<` → `&lt;` etc.), preventing you from accidentally introducing a wide range of XSS vulnerabilities.  (Otherwise you would need to use `st:out` to do the escaping explicitly, and you would probably forget.) `j:out` simply emits its argument raw, bypassing the escaping, for cases where the programmer knows that it is safe (and perhaps necessary) to include HTML tags.

PR #376, opened 2016-04-04
[JENKINS-31162] Support for Item categorization for MultiBranchProjects

recena:
This PR belongs to this [JENKINS-31162](https://issues.jenkins-ci.org/browse/JENKINS-31162) issue and allows to categorize the Multibranch projects in a right way.

downstream of https://github.com/jenkinsci/cloudbees-folder-plugin/pull/56

@reviewbybees especially @jglick


multibranch/pom.xml
```diff
@@ -88,7 +88,7 @@ THE SOFTWARE.
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>cloudbees-folder</artifactId>
-            <version>5.2.1</version>
+            <version>5.8-SNAPSHOT</version>
```
jglick: I suppose this is just so that `MultiBranchProjectDescriptor` and thus our own `DescriptorImpl` will implement `getCategoryId()`? If so, would it not be simpler to just implement `getCategoryId()` here directly, at least until the ID is defined by an API in `cloudbees-folder`?

amuniz: I think it's better with `5.8-SNAPSHOT` (if it does not introduce any issue, of course). In this way we are saving a TODO here (we already have a good amount to take care once plugins go to 2.0 baseline).

amuniz:
:bee:

multibranch/pom.xml
recena: I need `5.8-SNAPSHOT` because:

`WorkflowMultiBranchProject.DescriptorImpl` extends of `MultiBranchProjectDescriptor`, and this one extends of `AbstractFolderDescriptor`. For this relation, a priori, `WorkflowMultiBranchProject` should inherit the type of `ItemCategory` provides by `getItemCategoryId()`.

The downstream PR in this PR, [provides](https://github.com/jenkinsci/cloudbees-folder-plugin/pull/56/files#diff-b4544a5a8a71e3317534defa00e9cd95L43) this definition.

multibranch/src/main/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProject.java
```diff
@@ -71,7 +71,15 @@ public WorkflowMultiBranchProject(ItemGroup parent, String name) {
...
-            return "Multibranch Pipeline";
+            return Messages.WorkflowMultiBranchProject_DisplayName();
+        }
+
+        public String getDescription() {
+            return Messages.WorkflowMultiBranchProject_Description();
+        }
+
+        public String getIconFilePathPattern() {
+            return "plugin/workflow-multibranch/images/:size/pipelinemultibranchproject.png";
```
recena: `getCategoryId()` is being provided by `AbstractFolderDescriptor` due to `DescriptorImpl` extends of `AbstractFolderDescriptor`.

multibranch/pom.xml
jglick: I understand why you are bumping up the dependency, I am just pointing out that it is not _necessary_ because you could trivially implement `getCategoryId()` here, just leaving a one-line TODO comment to delete the override once the `cloudbees-folder` dependency is updated to 5.8+. The advantage would be that this PR could be merged now, rather than sometime later.

jglick:
:bee: but not currently mergeable.

recena:
@jglick Why not? I can see the green button :smile: or do you means for the downstream PR?

jglick:
Because it has an unreleased dependency, so merging this would block release of `workflow-multibranch`.

recena:
This PR has been moved to https://github.com/jenkinsci/workflow-multibranch-plugin/pull/1

PR #378, opened 2016-04-05
[JENKINS-34008] Add a hook listener for the global library

abayer:
[JENKINS-34008](https://issues.jenkins-ci.org/browse/JENKINS-34008)

Existing rebuild test validates this change.

cc @reviewbybees

### DIFF
--- a/cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/GlobalRepoHookListener.java
+++ b/cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/GlobalRepoHookListener.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jenkinsci.plugins.workflow.cps.global;
+
+import hudson.Extension;
+import hudson.ExtensionPoint;
+import javax.inject.Inject;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+public abstract class GlobalRepoHookListener implements ExtensionPoint {
+
+    public abstract void hook();
+
+    @Extension
+    @Restricted({NoExternalUse.class})
+    public static class UserDefinedGlobalVariableRepoListener extends GlobalRepoHookListener {
+        @Inject
+        UserDefinedGlobalVariableList globalVariableList;
+
+        @Override
+        public void hook() {
+            globalVariableList.rebuild();
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/GlobalRepoHookListener.java
```diff
@@ -0,0 +1,40 @@
...
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+public abstract class GlobalRepoHookListener implements ExtensionPoint {
+
+    public abstract void hook();
+
+    @Extension
+    @Restricted({NoExternalUse.class})
+    public static class UserDefinedGlobalVariableRepoListener extends GlobalRepoHookListener {
```
jglick: :ant: put this inside `UserDefinedGlobalVariableList` where it logically belongs.

cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/GlobalRepoHookListener.java
```diff
@@ -0,0 +1,40 @@
...
+
+import hudson.Extension;
+import hudson.ExtensionPoint;
+import javax.inject.Inject;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+public abstract class GlobalRepoHookListener implements ExtensionPoint {
+
+    public abstract void hook();
```
jglick: :ant: listener methods should have a meaningful name, and some Javadoc.

cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/GlobalRepoHookListener.java
abayer: Okiedokie.

cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/GlobalRepoHookListener.java
abayer: Yup, should have mentioned that this was a preliminary PR looking for feedback on the implementation in general. What would you prefer as a name?

jglick:
:bug: in order for such a hook to be useful, you would need API-supported access to the repository _without_ change notifications, which means you also need to define a way to get those files.

cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/GlobalRepoHookListener.java
jglick: Maybe `repositoryUpdated`?

BTW the naming of the class is inconsistent with `WorkflowLibRepository`. I might suggest just `WorkflowLibRepositoryListener`.

abayer:
I'm not sure what you mean.

jglick:
> I'm not sure what you mean.

Well it is pointless to be informed that files in some repository *changed* unless you can access them *once*, right? And there is no API for that currently.

abayer:
That's easy - see https://github.com/abayer/pipeline-action-plugin/blob/master/src/main/java/io/jenkins/plugins/pipelineaction/sources/GlobalRepoPipelineActionSet.java e.g.

cps-global-lib/src/main/java/org/jenkinsci/plugins/workflow/cps/global/GlobalRepoHookListener.java
abayer: I went with this name because that name is at this point fairly crap and way too long. =)

jglick:
Of course I know it is _possible_, but it is not exposed as a documented API.

abayer:
So injecting the repository, as is done in `UserDefinedGlobalVariableList`, isn't the right way to do it? Why not?

jglick:
Again, it works, but it is not a documented API.

abayer:
So...what needs to be done for it to be a documented API?

jglick:
https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/1

PR #380, opened 2016-04-11
Added info about compatibility of Cucumber Reports Plugin

damianszczepanik:


damianszczepanik:
Added in jenkinsci/cucumber-reports-plugin/pull/35 please if you can

PR #381, opened 2016-04-20
Suggestion for COMPATIBILITY.md

carlosrodlop:
Including within the Build steps and post-build actions the Backup plugin

JIRA OSS: https://issues.jenkins-ci.org/browse/JENKINS-34346

@reviewbybees

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -63,7 +63,7 @@ Entries list the class name serving as the entry point to the relevant functiona
 - [X] `PerfSigRecorder` and 5 more (`performance-signature-dynatrace`): supported as of 2.0
 - [ ] `StashNotifier` (`stashNotifier`): [issue 92](https://github.com/jenkinsci/stashnotifier-plugin/issues/92)
 - [X] `LambdaUploadBuildStep`, `LambdaInvokeBuildStep`, `LambdaEventSourceBuildStep` (`aws-lambda`): supported as of 0.5.0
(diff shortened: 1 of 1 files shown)


COMPATIBILITY.md
```diff
@@ -63,6 +63,7 @@ Entries list the class name serving as the entry point to the relevant functiona
 - [X] `PerfSigRecorder` and 5 more (`performance-signature-dynatrace`): supported as of 2.0
 - [ ] `StashNotifier` (`stashNotifier`): [issue 92](https://github.com/jenkinsci/stashnotifier-plugin/issues/92)
 - [X] `LambdaUploadBuildStep`, `LambdaInvokeBuildStep`, `LambdaEventSourceBuildStep` (`aws-lambda`): supported as of 0.5.0
+- [ ] `Backup`: [JENKINS-34346](https://issues.jenkins-ci.org/browse/JENKINS-34346)
```
jglick: :bug: There is no such build step that I can find.

oleg-nenashev:
I see no problem if somebody implements it.
By the way, Pipeline is not about Jenkins self-administration right now.
For such purpose I'd recommend Scriptler.

COMPATIBILITY.md
carlosrodlop: My fault, I didn't check the scope of the OSS Backup plugin, it doesn't work at job level so it doesn't make any sense to make it compatible within pipeline. No connection with the AbstractBuild, so it couldn't take part into a Build.

carlosrodlop:
According to my previous comments, I close this pull request.

PR #382, opened 2016-05-05
Adding Powershell to compat list

abayer:
cc @reviewbybees


recena:
👍

oleg-nenashev:
🐝

jglick:
Note that the actual `Builder` mentioned cannot be made Pipeline-compatible.

jglick:
(or could, but it is a bad idea)

PR #384, opened 2016-05-18
Add gatling plugin to compatibility list

cprice404:


COMPATIBILITY.md
```diff
@@ -65,6 +65,7 @@ Entries list the class name serving as the entry point to the relevant functiona
 - [X] `LambdaUploadBuildStep`, `LambdaInvokeBuildStep`, `LambdaEventSourceBuildStep` (`aws-lambda`): supported as of 0.5.0
 - [X] `CucumberReportPublisher` (`cucumber-reports`): supported as of 2.1.0
 - [ ] `Powershell` (`powershell`): [JENKINS-34581](https://issues.jenkins-ci.org/browse/JENKINS-34581)
+- [X] `Gatling` (`gatling`): supported as of 1.2.0
```
jglick: The `SimpleBuildStep` implementation in this plugin is called `GatlingPublisher` AFAICT.

cprice404: You are correct, sorry, didn't notice that it was class names rather than plugin names that were being used in this list; will fix.

cprice404:
updated.

COMPATIBILITY.md
jglick: Probably this should either be converted to a table, or the format changed entirely. TBD.

PR #387, opened 2016-05-19
[JENKINS-34943] [JENKINS-34943] Update compatibility with nexus and cppcheck

DJViking:


COMPATIBILITY.md
```diff
@@ -65,6 +65,8 @@ Entries list the class name serving as the entry point to the relevant functiona
 - [X] `LambdaUploadBuildStep`, `LambdaInvokeBuildStep`, `LambdaEventSourceBuildStep` (`aws-lambda`): supported as of 0.5.0
 - [X] `CucumberReportPublisher` (`cucumber-reports`): supported as of 2.1.0
 - [ ] `Powershell` (`powershell`): [JENKINS-34581](https://issues.jenkins-ci.org/browse/JENKINS-34581)
+- [X] `Nexus Artifact Uploader` (`nexus`): supported as of 2.2
```
jglick: No, this is confused.

First of all, the first string should be the class name of the `SimpleBuildStep`: `NexusArtifactUploader` I think.

The second string should be the plugin short name: `nexus-artifact-uploader`.

And whoever did this integration seems to have done it incorrectly. There is a dedicated `NexusArtifactUploaderStep` producing a `nexusArtifactUploader` step; *and* `NexusArtifactUploader` was made to implement `SimpleBuildStep`, meaning it will appear as an option to the `step` metastep. You only want to do one or the other, not both.

There are other problems in the integration, especially that `nexusPassword` is passed as a `String`, which is insecure (must use Credentials instead), but that does not matter here for purposes of recording *what* the integration consisted of.

COMPATIBILITY.md
```diff
@@ -65,6 +65,8 @@ Entries list the class name serving as the entry point to the relevant functiona
...
+- [ ] `CppcheckPublisher` (`cppcheck`): [JENKINS-34943](https://issues.jenkins-ci.org/browse/JENKINS-35096)
```
jglick: This part looks fine.

DJViking:
Fixed the naming of nexus-artifact-uploader in COMPATIBILITY.md

PR #388, opened 2016-05-20
Add `checkout scm` line to "Creating a Simple Pipeline"

chachi:
This `checkout scm` line is only mentioned in more complex pipelines, but seems to be required for even simple pipelines. Took me a long time to figure out I needed to add this to my `node {...}` declarations.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -97,6 +97,7 @@ Now, click on your Pipeline and **Configure** it to edit its script.
 
 ```groovy
 node {
+  checkout scm
   git url: 'https://github.com/jglick/simple-maven-project-with-tests.git'
   def mvnHome = tool 'M3'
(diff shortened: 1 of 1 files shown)


jglick:
Well it _does_ need to get mentioned more prominently, but in the suggested location it is wrong: there is a `git` step there already! `checkout scm` is used in

* multibranch Pipelines, when you wish to check out repository sources matching `Jenkinsfile`
* single Pipelines loading the script from SCM, when you wish to check out that same repository

chachi:
@jglick You're totally right; thanks for taking a look!

It's the latter case that threw me for a loop and I think could use its own section in the tutorial. I'm not sure it's explicitly mentioned anywhere that if the script was loaded from SCM, this does not mean the repository is.

PR #404, opened 2016-08-23
Update COMPATIBILITY.md

tarioch:
Claim plugin does not work with pipeline


jglick:
Would rather the original issue purporting to add Pipeline support be looked up and JENKINS-33969 made to block that, with some investigation of whether this is a genuine regression, a user error, etc. The compatibility list is only intended as a general guide for which plugins *purport* to have been updated, not a replacement for a live bug tracker.

PR #414, opened 2016-10-22
Fix reference to "unarchive"

david-resnick:
Actually, I'm not sure that 

`stash name: 'sources', includes: 'pom.xml,src/'` is shorthand for `step([$class: 'StashStep', includes: 'pom.xml, src/'])`.

But the current version mentions unarchive which is less correct.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -606,22 +606,22 @@ The rest of the Pipeline then runs these in parallel — so if you look at **tre
 If you only have the one agent configured with its two executors, this won't save as much time, but you may have multiple agents on different hardware matching the same label expression.
 
 This script is more complex than the previous ones so it bears some examination.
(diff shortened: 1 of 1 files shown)


TUTORIAL.md
```diff
@@ -606,22 +606,22 @@ The rest of the Pipeline then runs these in parallel — so if you look at **tre
...
 ```groovy
-archive 'pom.xml, src/'
+stash name: 'sources', includes: 'pom.xml,src/'
 ```
 
 is shorthand for the more general:
 
 ```groovy
-step([$class: 'ArtifactArchiver', artifacts: 'pom.xml, src/'])
+step([$class: 'StashStep', includes: 'pom.xml, src/'])
```
jglick: No actually it is not shorthand for anything; this is obsolete, left over from when the sample actually did use `archive`.

jglick:
I think fixed in #417. Sorry for taking so long to review PRs…

PR #423, opened 2016-12-15
Adding rocketchat-notifier plugin

hypery2k:


COMPATIBILITY.md
```diff
@@ -47,6 +47,7 @@ Newly filed issues should bear the label `pipeline` for ease of tracking.
 - [X] `HipChatNotifier` (`hipchat`): supported as of 1.0.0
 - [ ] `IronMQNotifier` (`ironmq-notifier`): [JENKINS-35505](https://issues.jenkins-ci.org/browse/JENKINS-35505)
 - [X] `LogParserPublisher` (`log-parser`): supported as of 2.0
+- [X] `RocketChatNotifier` (`rocketchat-notifier`): supported as of 2.0
```
jglick: The plugin name is `rocketchatnotifier` AFAICT.

hypery2k:
I've corrected the name. The workflow support can be found here: https://github.com/jenkinsci/rocketchatnotifier-plugin/blob/master/src/main/java/jenkins/plugins/rocketchatnotifier/workflow/RocketSendStep.java

COMPATIBILITY.md
```diff
@@ -47,6 +47,7 @@ Newly filed issues should bear the label `pipeline` for ease of tracking.
 - [X] `HipChatNotifier` (`hipchat`): supported as of 1.0.0
 - [ ] `IronMQNotifier` (`ironmq-notifier`): [JENKINS-35505](https://issues.jenkins-ci.org/browse/JENKINS-35505)
 - [X] `LogParserPublisher` (`log-parser`): supported as of 2.0
+- [X] `RocketChatNotifier` (`rocketchatnotifier`): supported as of 2.0
```
jglick: What is this 2.0? The wiki says the latest is 0.5.0. And the step exists from at least 0.2.1.

And how is this `RocketChatNotifier`? This is a custom step, which has its own section.

hypery2k:
@jglick Thanks for the hint. I've updated the version number

COMPATIBILITY.md
```diff
@@ -47,6 +47,7 @@ Newly filed issues should bear the label `pipeline` for ease of tracking.
 - [X] `HipChatNotifier` (`hipchat`): supported as of 1.0.0
 - [ ] `IronMQNotifier` (`ironmq-notifier`): [JENKINS-35505](https://issues.jenkins-ci.org/browse/JENKINS-35505)
 - [X] `LogParserPublisher` (`log-parser`): supported as of 2.0
+- [X] `RocketChatNotifier` (`rocketchatnotifier`): supported as of 0.5.0
```
jglick: Still incorrect as previously noted. This is a custom step (`rocketSend`). The original `RocketChatNotifier` post-build action is not compatible with Pipeline AFAICT.

PR #425, opened 2016-12-22
Build parameters not groovy variables anymore

GabLeRoux:
I had an issue when working with Build parameters, https://github.com/jenkinsci/pipeline-plugin/blob/master/TUTORIAL.md#build-parameters is confusing. I
found the solution in [JENKINS-40235](https://issues.jenkins-ci.org/browse/JENKINS-40235) and added details here: http://stackoverflow.com/a/41276956/1092815

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -259,8 +259,6 @@ See  Help in the **Snippet Generator** for the `withEnv` step for more details o
 
 If you have configured your pipeline to accept parameters when it is built — **Build with Parameters** — they are accessible as Groovy variables inside `params`. They are also accessible as environment variables.
 
-Example:
-
 **Example**: Using `isFoo` parameter defined as a boolean parameter (checkbox in the UI):
 
 ```groovy
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -257,12 +257,16 @@ See  Help in the **Snippet Generator** for the `withEnv` step for more details o
 
 ## Build Parameters
 
-If you have configured your pipeline to accept parameters when it is built — **Build with Parameters** — they are accessible as Groovy variables inside `params`. Example:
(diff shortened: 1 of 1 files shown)


GabLeRoux:
@jglick bump
Let me know if you think there's a way I can improve this pr :+1:

TUTORIAL.md
```diff
@@ -257,7 +257,16 @@ See  Help in the **Snippet Generator** for the `withEnv` step for more details o
 
 ## Build Parameters
 
-If you have configured your pipeline to accept parameters when it is built — **Build with Parameters** — they are accessible as Groovy variables of the same name.
+If you have configured your pipeline to accept parameters when it is built — **Build with Parameters** — they are accessible as Groovy variables inside `params`. Example:
```
jglick: Well, they are _also_ accessible as environment variables (and thus looking like Groovy variables) of the same name, though as plain text and not considering default values. `params` is better for most purposes so I am fine with just mentioning it.

TUTORIAL.md
```diff
@@ -257,7 +257,16 @@ See  Help in the **Snippet Generator** for the `withEnv` step for more details o
...
+
+```groovy
+node {
+  sh "isFoo is ${params.isFoo}"
+  sh "isFoo is " + params.isFoo
```
jglick: Better style to use single quotes `'` when not interpolating.

TUTORIAL.md
```diff
@@ -257,7 +257,16 @@ See  Help in the **Snippet Generator** for the `withEnv` step for more details o
...
+  if (params.isFoo) {
```
jglick: Might want to note that this assumes the `isFoo` parameter was defined a boolean parameter (checkbox in the UI). Well, maybe Groovy will treat nonempty string parameters as truish, not sure offhand.

GabLeRoux:
Thanks for your comments Jesse, really helpful :+1:

GabLeRoux:
I updated this pr based on your review :)

PR #429, opened 2017-01-20
mark scoring-load-balancer job-level preferences as unsupported

mrooney:
I wasn't sure if this belongs in "Misc" or the "Build wrappers" section, and I also guessed a little bit on BuildPreferenceJobProperty; that's the class that was persisted in the config.xml of the job so it seemed right.

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -149,7 +149,7 @@ Newly filed issues should bear the label `pipeline` for ease of tracking.
 - [X] `jobconfighistory`: supported as of 2.14
 - [X] `next-build-number`: supported as of 1.4
 - [ ] `tracking-svn` [JENKINS-38060](https://issues.jenkins-ci.org/browse/JENKINS-38060)
-- [ ] `BuildPreferenceJobProperty` (`scoring-load-balancer`): [JENKINS-41267](https://issues.jenkins-ci.org/browse/JENKINS-41267)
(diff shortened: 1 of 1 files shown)


COMPATIBILITY.md
```diff
@@ -149,6 +149,7 @@ Newly filed issues should bear the label `pipeline` for ease of tracking.
 - [X] `jobconfighistory`: supported as of 2.14
 - [X] `next-build-number`: supported as of 1.4
 - [ ] `tracking-svn` [JENKINS-38060](https://issues.jenkins-ci.org/browse/JENKINS-38060)
+- [ ] `BuildPreferenceJobProperty` (`scoring-load-balancer`): [JENKINS-41267](https://issues.jenkins-ci.org/browse/JENKINS-41267)
```
jglick: Better leave out the class name for now, as with other entries in this section.

mrooney: Okay, I've changed it, but there seems to be a precedent for providing it (PerfSigRecorder, LambdaUploadBuildStep, ConfigFileBuildWrapper, ListSubversionTagsParameterValue, etc) and it would have saved someone the work of finding the right class and give a little head start should they decide to implement. Anyway, is it mergeable at this point? I've struggled to find the status of Pipeline support for various plugins only to find out later there was an unmerged documentation PR for it that would have saved me a lot of time. I'd love to be able to help others with this.

PR #434, opened 2017-03-25
Update COMPATIBILITY.md to include R plug-in

kinow:
[Previously submitted](https://github.com/jenkinsci/pipeline-plugin/pull/427/) in the incorrect category, but also from a branch in the jenkinsci org. Re-submitting from the fork under the suggested category.

Cheers
Bruno

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -167,6 +167,7 @@ For cases when a first-class Pipeline step (rather than an adaptation of functio
 - [X] `marathon`: `marathon` step since 1.2.1
 - [ ] `Release Plugin` (`release`): [JENKINS-40765](https://issues.jenkins-ci.org/browse/JENKINS-40765)
 - [ ] `M2 Release Plugin` (`m2release`): [JENKINS-40766](https://issues.jenkins-ci.org/browse/JENKINS-40766)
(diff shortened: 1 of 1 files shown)


COMPATIBILITY.md
```diff
@@ -167,6 +167,7 @@ For cases when a first-class Pipeline step (rather than an adaptation of functio
 - [X] `marathon`: `marathon` step since 1.2.1
 - [ ] `Release Plugin` (`release`): [JENKINS-40765](https://issues.jenkins-ci.org/browse/JENKINS-40765)
 - [ ] `M2 Release Plugin` (`m2release`): [JENKINS-40766](https://issues.jenkins-ci.org/browse/JENKINS-40766)
+- [X] `r`: (`r-plugin`): `r` step as of 0.4 via [pull request 4](https://github.com/jenkinsci/r-plugin/pull/4)
```
daniel-beck: None of the other implemented steps mention where it's implemented. This makes sense as that's more the plugin's documentation, than here. Also, the plugin is just `r`, not `r-plugin`. So I'd expect something like:

- [X] `r`: `r` step as of 0.4

jglick: Right.

PR #435, opened 2017-04-08
JENKINS-40765 compatible since version 2.7

estarter:


oleg-nenashev:
Sorry, missed it. Please update the text to mention the version where it is available

daniel-beck:
Right, format should be adapted to the usual one for implemented steps.

PR #440, opened 2017-08-25
Updated support status for Coverity Plugin

jbriggs22:
As of Coverity plugin version 1.10.0 there is support for pipeline. We created a new publisher step instead of working pipeline support into the existing `CoverityPublisher`. Also added a `SimpleBuildWrapper` to use a `CoverityToolInstallation` on the path.


shamiwillms:
@jglick how do we get this merged?  is there a process we are not aware of? Thanks for your help.

jglick:
Needs a repo owner to merge things.

